PR #1, opened 2015-02-17
Code quality fixes

n-west:
Address most warnings from gcc -Wall and -Wextra as well as a couple of missing string.h includes from puppets. The last warning not related to ORC is in the convolution decoder


douggeiger:
The change in the last commit to make void(*_kernel_ptr)() static gives me compilation errors.

n-west:
Good catch. I thought there was a reason, but after removing it I don't see an error.

douggeiger:
Looks good to me

PR #4, opened 2015-03-05
Speed up static building in newer cmake versions by using object library...

n-west:
Copied over from https://github.com/n-west/volk/pull/1 via @hades-

Pending copyright assignment.


n-west:
applied

PR #5, opened 2015-03-19
Dickens osx fixes

n-west:
relevant patches from @michaelld re:https://github.com/gnuradio/gnuradio/pull/405#issuecomment-83123800

### DIFF
--- a/include/volk/volk_prefs.h
+++ b/include/volk/volk_prefs.h
@@ -14,7 +14,8 @@ typedef struct volk_arch_pref
 } volk_arch_pref_t;
 
 ////////////////////////////////////////////////////////////////////////
-// get path to volk_config profiling info
+// get path to volk_config profiling info;
+// returns \0 in the argument on failure.
 ////////////////////////////////////////////////////////////////////////
 VOLK_API void volk_get_config_path(char *);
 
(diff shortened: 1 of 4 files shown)


michaelld:
n-west: Great!

Did you see my patch "volk: undefine LOG_POLY_DEGREE at the end of each header using it, so-as to not conflict with other headers."? If not, I'll do it in a separate PR; if so, can you comment on why not those changes?

n-west:
Yea, I didn't include that because I had something similar to fix this in 079b33893ac61c28944622d139511cec853ffbe2. If you've got experience on a 'best practice' being to undefine the symbol I don't have an issue with doing that. I can see that potentially being the case since all of the code gets copied into a volk_machine template.

michaelld:
Whether to undefine or not depends on the scope of the variable. In this case, multiple files #define the same variable, but do not #undef them. So, the interpretation of which value is used is then up to the compiler -- I think most overwrite one value with the next, but I don't think there's any guarantee of this being the case. Hence, if the scope of the variable is just that one file (which I believe is the case here, since that variable is used in just those 2 headers, and nowhere else), then the best practice is to #undef it at the end.

PR #11, opened 2015-04-03
Fix add test

michaelld:
and some other tweaks. Sorry for combining all together; maybe cherry pick the others. We need to test VOLK_ADD_TEST more to make sure it works on various OSs. I think it'll work for most "bash" like shells. More testing needed on that commit.


n-west:
Looks like everything is still working for me and I don't see anything real risky here, so I'm inclined to merge it in.

Tomorrow I'm going to verify that I got all of your branches, run through my 3-arch test plan, merge and release.

PR #12, opened 2015-04-07
volk: accurate exp kernel.

trondeau:
A more accurate exp VOLK kernel than volk_32f_expfast_32f.Taken from
code licensed with zlib.

Moving this from gnuradio to here. We still need to figure out the zlib license issue first.


n-west:
🤔 still haven't figured out zlib license, but we have other LGPL code. I'm going to hold on to this for just a tad longer

marcusmueller:
@n-west um, any news on zlib compliance? Personally, I'd give my green light: GPL fully permits linking against zlib, and these kernels don't use GPL code themselves.

michaelld:
I like this addition to Volk. @marcusmueller notes astutely about zlib and GPL, so I'd say we make this 'go' once (2) below is done ... unless there's pushback from (1) in a timely manner.

For folks on this PR: (1) Any update on the license? (2) Can we get this PR rebased? Can't merge as is even if we wanted to!

n-west:
I don't really know how to deal with zlib / mixed licenses. The original source is http://gruntthepeon.free.fr/ssemath/ and I would guess there's enough divergence that you can claim tom/OG author of this created a new work even though there are similarities

michaelld:
@n-west @bhilburn has approved another PR using zlib license, so I'm going with doing it here is OK too.

jdemel:
The code looks good to me. Is the license statement correct? Do we need to add the GPL part too?

michaelld:
> The code looks good to me. Is the license statement correct? Do we need to add the GPL part too?

That's a good question!

@bhilburn Please advise on whether we need to add the GPL license at the top/start of of this PR's added file.

michaelld:
Based on #275 I believe the GPL header should be added. Also: This PR need to be rebased.

jdemel:
In order to some this up:

1. The GPL header needs to be added.
2. The PR needs to be rebased because `kernel_tests.h` received quite a few more updates since this PR was posted.

Are those the only two items that stop this PR from being merged? @trondeau Could you do those fixes? Then we can finally merge this PR before we hit the 5 year milestone.

michaelld:
@jdemel I strongly doubt @trondeau will be able to do anything here. Since this is not his code -- just the "glue" portion between Volk and the original code, which anybody can write -- our best option is to do the work ourselves & give him some credit. Should be pretty straight forward.

jdemel:
I just rebased this PR and resubmitted it in #353 . I assume we can continue our discussion there.

michaelld:
https://github.com/gnuradio/volk/pull/353

PR #14, opened 2015-04-09
lib: add back gcc_x86_cpuid.h

michaelld:
Since it is not part of the QA system removed in 270c961e, but rather part of the build system for older GCC compilers that do not provide cpuid.h

### DIFF
--- a/lib/gcc_x86_cpuid.h
+++ b/lib/gcc_x86_cpuid.h
@@ -1,0 +1,188 @@
+/*
+ * Copyright (C) 2007, 2008, 2009, 2010 Free Software Foundation, Inc.
+ *
+ * This file is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License as published by the
+ * Free Software Foundation; either version 3, or (at your option) any
+ * later version.
+ *
+ * This file is distributed in the hope that it will be useful, but
+ * WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * General Public License for more details.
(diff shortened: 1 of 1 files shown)


n-west:
Do you know what versions of GCC do not provide cpuid.h?

michaelld:
Not sure yet. It exists in 4.5, and am installing 4.3 and 4.4 to test those. It looks like it was not part of 4.2. Of course, all of these are pretty old. This header is included from "build/lib/volk_cpu.c", which was created from "tmpl/volk_cpu.tmpl.c", which contains the #include "gcc_x86_cpuid.h" header if <cpuid.h> is not found. So, this is certainly not part of the QA code. OTOH, if you were to ban any GCC < 4.5, then this would not be an issue, at least for GCC. I need to check with Clang, but I think all versions of it provide cpuid.h.

michaelld:
Looks like GCC 4.3 provides cpuid.h. I don't actually have a stock install of GCC 4.2, just Apple's -- so, it might be Apple issue of not providing cpuid.h in their version of GCC 4.2? Anyway, adding in this file takes care of the issue with that compiler. If you don't want to add this file back in, I can always block the installer from using that compiler -- it will then choose another, newer one, which will work. Your choice. If you do go with not adding in this file, then I'd recommend removing the #include for it.

n-west:
I removed the include in 4c967ef1fe49e13f85a466d3ff0f5d20c6782d78 since this is only needed for very old GCC's

michaelld:
NP. Worth a try. There's probably another way around having to use this header, if I had time to figure it out (which I don't; so, I'll just leave a patch in place in MacPorts to support older compilers, e.g., on 10.5 and 10.6).

PR #15, opened 2015-04-13
fixed the out-of-bounds access by using boost to split string

marcusmueller:


n-west:
Fixed (I think) in 314bab26ae972d62af248ae0d61e002544bd2188

PR #17, opened 2015-04-15
minifix: int counter=0; for(counter=0;... is redundant

marcusmueller:
kind of a matter of style, so please feel free to throw away as unnecessary.


n-west:
I'm going to leave this open for now. At some point I'd like to go through and add it in with some other style changes. I've got this in n-west/code-style.

PR #19, opened 2015-05-12
Coverity

n-west:
Address a few outstanding coverity issues.

@douggeiger, do you have any style/best practice suggestions on catching C++ exceptions? Is it best to catch a specific exception or just std::exception and print exception.what()?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,9 @@ enable_language(CXX)
 enable_language(C)
 enable_testing()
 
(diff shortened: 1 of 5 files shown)


douggeiger:
To address the question - first I would say that generally speaking you want to be as specific as possible when catching exceptions: i.e. if you know that a specific function can throw a specific exception, then catch that, and if possible deal with it (and if not, either error out, or re-throw, or throw std::runtime_error() - as may be appropriate for the location of the exception). Secondly, I have a patch to send you on top of this branch that could be rolled into this PR.

Issue #18, opened 2015-04-16
volk_profile and make test segfault on volk_32f_s32f_32f_fm_detect_32f kernel.


drmpeg:
Function parameters 2 (const float bound) and 3 (float\* saveValue) in volk_32f_s32f_32f_fm_detect_32f are reversed. This causes the saveValue pointer to be invalid and the reason for the segfault.

Since the saveValue pointer gets a floating point value, it can often be a reasonable address and this bug can be masked on many systems.

n-west:
I've never observed a segfault, but I guess I can see it happening. I think the easiest thing is to make a puppet for this block. I'll get around to it at some point if you don't beat me to it.

wkazubski:
This probably makes all my attempts of creating i586 (32-bit) binaries to fail. Build crashes at make test step. If I omit the test, an attempt to run gnuradio-companion fails with "Memory protection error" message.

n-west:
Gnuradio-companion is entirely unrelated to VOLK. The qa failure and volk profiler are the only affected applications.

wkazubski:
Thanks for advice. Gnuradio-companion crashes too soon to be an error somewhere in volk library.
BTW: If the patch is available, I can test it.

n-west:
Here's a proposed fix: https://github.com/n-west/volk/commit/a1902852b5163a03c34b559af2afb2a392196656

If you test it let me know if it fixes your problem and I'll merge it in.

wkazubski:
make test still fails on 32-bit intel, but with another error. Gdb says:
gdb test_all
[...](gdb) run
[...]/BUILD/volk/build/lib/test_all: symbol lookup error: /home/wk/rpmbuild/BUILD/volk/build/lib/test_all: undefined symbol: volk_32f_x2_fm_detectpuppet_32f_get_func_desc
[Inferior 1 (process 9667) exited with code 0177](gdb)

n-west:
@wkazubski It looks like you didn't run cmake. Part of the cmake step in VOLK is creating that function in some custome templating code. Run cmake, make, make test should make it work out.

wkazubski:
The problem happend while building a binary RPM package using rpmbuild command and a .spec file. Such build is from scratch by default, so it is impossible to omit cmake by mistake.
It also fails when build on openSUSE Build Service:
https://build.opensuse.org/package/show/home:wkazubski:test/volk
64-bit build is OK.

n-west:
[...]

PR #22, opened 2015-05-16
Small fixes for windows. 

### LINKED ISSUE #21: test_all fails on volk_32fc_x2_dot_prod_32fc_a_sse4_1 on windows.
Root cause is this initializer. 

https://github.com/gnuradio/volk/blob/master/kernels/volk/volk_32fc_x2_dot_prod_32fc.h#L314

Due to the way __m128i is defined on windows (by both the msvc and intel headers) the constant gets truncated and the sign flip never happens.

see this thread for details  https://gcc.gnu.org/ml/gcc-help/2009-04/msg00363.html

replacing neg with

```
__m128i neg = _mm_set_epi32(0,0,0,0x80000000);
```

fixes the issue.

LazyDodo:
- Depended on the fix for #18  being applied for X86 otherwise tests will crash. 
- Fixes #21 
- Tested with visual studio 2013

### DIFF
--- a/gen/archs.xml
+++ b/gen/archs.xml
@@ -73,6 +73,7 @@
   <flag compiler="gnu">-mpopcnt</flag>
   <flag compiler="clang">-mpopcnt</flag>
   <flag compiler="msvc">/arch:AVX</flag>
+  <flag compiler="msvc64">/arch:AVX</flag>
 </arch>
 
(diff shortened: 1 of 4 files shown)


LazyDodo:
You won't take it without copyright assignment.

PR #28, opened 2015-07-10
Properly initialize the all zero vectors for the AVX2 flavors.

evilynux:
In the AVX2 implementations, the zero_val vector is never initialized.


n-west:
Small nitpick (from http://libvolk.org/contributing.html): I prefer commit messages to have a 'component: message' format because it makes it much easier for me to see what's changed when looking at git log and shortlog.

For kernel changes I usually use: 'kernels: ' and like to include a reference to the kernel name. Ex: 'kernels: initialize zero vectors in 32f_binary_slicer_8i AVX2 implementations'. I can do the amend and keep the commit in your name later today, or if you push an update (I don't mind a force on this) I'll use that.

evilynux:
I'm not too familiar with this, but did that work?

n-west:
Yea, that looks good. Thanks!

PR #37, opened 2015-08-28
Make builds reproducible by getting rid of the builddate.

evilynux:
No one uses the build date anyway and this fix will make the Debian people happy.


n-west:
I tweaked some commit messages, but these should all be merged

PR #48, opened 2015-09-26
logging: log to stderr

pinkavaj:
This allows applications to use stdout for data output.

### DIFF
--- a/include/volk/logging.h
+++ b/include/volk/logging.h
@@ -1,6 +1,5 @@
-/* -*- c++ -*- */
 /*
- * Copyright 2014 Free Software Foundation, Inc.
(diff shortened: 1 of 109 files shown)


vsergeev:
:+1:, but I don't believe the Doxygen examples should use `VOLK_LOG()`.

n-west:
Always printing to stderr is not the right thing, especially since most of these messages are not errors. I'd rather see a lightweight thing similar to this that has something like a VOLK_LOG(LOG_LEVEL, varargs), where the log_level determines if something should be printed and can be either compile time definition or read from a preferences file. a compile time decision is probably least effort

n-west:
I accepted a PR to remove these statements until we can get a better minimalist logging setup.

PR #62, opened 2016-02-01
Improved processor/feature detection when building with MSVC

gnieboer:
Two changes to allow building in MSVC 2015. 
First is simply because windows cmake.exe appears to use uppercase processor names.
Second is because MSVC doesn't supports SSE/SSE2 arch flags when compiling 64-bit (because all 64-bit CPUs support SSE/SSE2) which will cause volk to not build the correct profiles.


n-west:
I changed the commit message and fixed white space, but this is merged now.

PR #64, opened 2016-02-15
Fix MSVC build

pcercuei:
Those few commits fix the Visual Studio build.

### DIFF
--- a/tmpl/volk_cpu.tmpl.c
+++ b/tmpl/volk_cpu.tmpl.c
@@ -73,7 +73,7 @@ struct VOLK_CPU volk_cpu;
 
 static inline unsigned int cpuid_count_x86_bit(unsigned int level, unsigned int count, unsigned int reg, unsigned int bit) {
 #if defined(VOLK_CPU_x86)
-    unsigned int regs[4] = {0};
+    int regs[4] = {0};
     cpuid_x86_count(level, count, regs);
     return regs[reg] >> bit & 0x01;
 #else
--- a/cmake/Modules/VolkAddTest.cmake
+++ b/cmake/Modules/VolkAddTest.cmake
@@ -166,7 +166,7 @@ function(VOLK_ADD_TEST test_name)
     #functionality will be removed in the future.  Leave it here for
     #now until someone can figure out how to do this in Windows.
     foreach(target ${test_name} ${VOLK_TEST_TARGET_DEPS})
-      get_target_property(location ${target} LOCATION)
+      get_target_property(location "${target}" LOCATION)
       if(location)
         get_filename_component(path ${location} PATH)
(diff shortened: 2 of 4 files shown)


n-west:
I think this undoes some changes that @gnieboer recently made to build on MSVC 15. I don't ever work with MSVC, and I can't test this stuff. Can you guys sort out what is actually required here? There's a lot of churn on the same bits of code here.

tmpl/volk_cpu.tmpl.c
```diff
@@ -73,7 +73,7 @@ struct VOLK_CPU volk_cpu;
 
 static inline unsigned int cpuid_count_x86_bit(unsigned int level, unsigned int count, unsigned int reg, unsigned int bit) {
 #if defined(VOLK_CPU_x86)
-    unsigned int regs[4] = {0};
+    int regs[4] = {0};
```
n-west, line 76: the rest of the code wants this as an unsigned int, except for the signature of the windows call. I'd much rather see a cast inside the MSVC specific code rather than changing types to satisfy MSVC warnings

guruofquality, line 76: Looks like the `(int *)` cast on the macro got lost, at least on my patch. So My MSVC specific macro looked like this `#define cpuidex_x86(op, cnt, r) __cpuidex(((int*)r), op, cnt)`, then you dont need to change the regs type. But it ended up in the code like `#define cpuid_x86_count(op, count, regs) __cpuidex(regs, op, count)`

gnieboer:
At first glance Josh's approach looks cleaner than mine.  I'll patch that into mine and post back any issues.

n-west:
merged. @pcercuei you might have to do a git reset --hard since I replaced the third commit with a cast to int*.

PS -- nice on the appveyor thing. I've heard some other people interested in CI before as well. We currently have it set up on gnuradio's jenkins. Do these other CI apps provide much advantage over jenkins?

guruofquality:
> PS -- nice on the appveyor thing. I've heard some other people interested in CI before as well. We currently have it set up on gnuradio's jenkins. Do these other CI apps provide much advantage over jenkins?

I've been using travis CI for linux and appveyor for windows. Its just as easy as a .yml file (for each service) in the root of the repo and its pretty well integrated with github. I think some of best features are telling you if a pull request causes any build failures and emailing everyone involved if a commit breaks something. People have even fixed pull requests without my even thinking about it because of the CI alone.

Volk is a reasonably small project so its quick to build and easy to get dependencies for. I don't know how touchy the units tests are, but even if the CI doesnt run the unit tests, knowing that the build compiled on several machines is pretty nice. I recommend it. Also, as an example if it helps: https://github.com/pothosware/SoapySDR/ (.travis.yml and appveyor.yml)

PR #65, opened 2016-02-24
The fix for compilation with cmake-3.5

yarda:
It seems gnuradio (volk) fails to build with cmake-3.5. The problem is probably that the project uses CMAKE_SOURCE_DIR/CMAKE_BINARY_DIR which seems to be never supported by cmake and now appears to be no longer allowed. The fix is probably to use PROJECT_SOURCE_DIR/PROJECT_BINARY_DIR instead. Downstream Fedora bugzilla:

https://bugzilla.redhat.com/show_bug.cgi?id=1311358

### DIFF
--- a/apps/CMakeLists.txt
+++ b/apps/CMakeLists.txt
@@ -19,23 +19,23 @@
 # Setup profiler
 ########################################################################
 if(MSVC)
-    include_directories(${CMAKE_SOURCE_DIR}/cmake/msvc)
+    include_directories(${PROJECT_SOURCE_DIR}/cmake/msvc)
 endif(MSVC)
 
 include_directories(
     ${CMAKE_CURRENT_SOURCE_DIR}
     ${CMAKE_CURRENT_BINARY_DIR}
-    ${CMAKE_SOURCE_DIR}/include
-    ${CMAKE_BINARY_DIR}/include
-    ${CMAKE_SOURCE_DIR}/lib
-    ${CMAKE_BINARY_DIR}/lib
+    ${PROJECT_SOURCE_DIR}/include
+    ${PROJECT_BINARY_DIR}/include
+    ${PROJECT_SOURCE_DIR}/lib
+    ${PROJECT_BINARY_DIR}/lib
     ${Boost_INCLUDE_DIRS}
 )
 
 # MAKE volk_profile
 add_executable(volk_profile
     ${CMAKE_CURRENT_SOURCE_DIR}/volk_profile.cc
-    ${CMAKE_SOURCE_DIR}/lib/qa_utils.cc
+    ${PROJECT_SOURCE_DIR}/lib/qa_utils.cc
 )
 
 
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -90,7 +90,7 @@ endif()
 ########################################################################
 execute_process(
     COMMAND ${PYTHON_EXECUTABLE} ${PYTHON_DASH_B}
-    ${CMAKE_SOURCE_DIR}/gen/volk_compile_utils.py
+    ${PROJECT_SOURCE_DIR}/gen/volk_compile_utils.py
     --mode "arch_flags" --compiler "${COMPILER_NAME}"
     OUTPUT_VARIABLE arch_flag_lines OUTPUT_STRIP_TRAILING_WHITESPACE
 )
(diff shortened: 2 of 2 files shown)


jmcorgan:
@n-west can you test this for both the stand alone and the submodule builds?

yarda:
Ok, I will try to rebuilt the volk standalone. I already successfully built the whole gnuradio downstream in Fedora Rawhide:
http://koji.fedoraproject.org/koji/taskinfo?taskID=13114026

yarda:
I successfully rebuilt volk standalone with the patch applied. I rebuilt it using cmake-3.4.1, but it shouldn't matter.

n-west:
@jmcorgan yes. Is this only an issue for VOLK, or GNU Radio too? I think GNU Radio uses CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR all over the place.

michaelld:
I don't know what version of CMake the OP is using, but CMAKE_SOURCE_DIR and CMAKE_BINARY_DIR have been part of CMake since at least 2.6. Their meaning has not changed between versions. See also:
https://cmake.org/cmake/help/cmake2.6docs.html#variable:CMAKE_SOURCE_DIR
https://cmake.org/cmake/help/v2.8.12/cmake.html#variable:CMAKE_SOURCE_DIR
https://cmake.org/cmake/help/v3.5/variable/CMAKE_SOURCE_DIR.html

yarda:
I forwarded FTBFS fix with comments from downstream Fedora cmake maintainer. Adding him to the discussion (through the original ticket).

opoplawski:
It's not that you cannot use CMAKE_SOURCE_DIR/CMAKE_BINARY_DIR - it's that you cannot _set_ them.  Which is what volk/CMakeLists.txt did, and then used CMAKE_SOURCE_DIR/CMAKE_BINARY_DIR in the rest of that project assuming that they would have the changed values.  cmake 3.5 prevents you from changing them now (and which you shouldn't have ever done), so instead volk needed to use PROJECT_SOURCE_DIR/PROJECT_BINARY_DIR.  The CMAKE_*_DIR usage in the rest of gnuradio is fine.  HTH.

michaelld:
Ah; OK; that makes sense now. Thanks for the clarification. Let me poke around some.

michaelld:
PROJECT_{SOURCE,BINARY}_DIR has been part of CMake since at least 2.6: https://cmake.org/cmake/help/cmake2.6docs.html#variable:PROJECT_SOURCE_DIR
It is still part of CMake 3.5, with the same meaning: https://cmake.org/cmake/help/v3.5/variable/PROJECT_SOURCE_DIR.html
For stand-alone projects, using CMAKE_{SOURCE,BINARY}_DIR is fine. Using PROJECT\* is required for embedded projects, such as when Volk is a submodule of GNU Radio. So, we don't need to worry about GR's use of these variables, just Volk's. I'll test out the commit & report back.

michaelld:
Volk builds fine as a stand-alone package with this commit. Next testing as an embedded project.

michaelld:
I don't think this patch cover every instance of those variables. Anybody know if we have to change them all, or just some?

yarda:
I don't know, but without this patch it didn't compile, with this patch it compiles OK.

rakuco:
> I don't think this patch cover every instance of those variables. Anybody know if we have to change them all, or just some?

For the record, the commit that caused this behavior change in CMake is kitware/cmake@360e4e1db0cc89ea4f092eeb1ba1e6df8ee782b7.

After this commit, setting `CMAKE_{BINARY,SOURCE}_DIR` like volk does works in the `CMakeLists.txt` where this is done as well as the other files it includes, but not in files added via `add_subdirectory`.

This pull request changes the `CMakeLists.txt` files in the directories added via `add_subdirectory()` from volk's top-level `CMakeLists.txt`, and I believe they should be enough to get things to work both in a standalone build as well as in a build part of GNU Radio. The other references to `CMAKE_{BINARY,SOURCE}_DIR` are either in the file that changes those two variables (the top-level `CMakeLists.txt`) or files referenced from it (`Doxyfile.in` which is processed in the top-level `CMakeLists.txt` and `cmake/Modules/Volk{Python,Version}.cmake`, both of which are added via `include()` calls in the top-level `CMakeLists.txt`).

I guess if you really want to be safe and consistent you could change all occurrences and stop changing `CMAKE_{BINARY,SOURCE}_DIR`, but that is not strictly required at the moment.

PR #68, opened 2016-03-08
Additional MSVC compatibility fixes

gnieboer:
The four most recent commits (bottom ones) are the actual new changes, ignore the prior ones.  Recommend cherry picking only those.  All are atomic commits.

The previous pull request would allow volk to compile on MSVC, but would only recognize the generic and sse2 arches.  These four commits should allow detection of the appropriate capabilities on MSVC for both 32-bit and 64-bit.


n-west:
I cherry-picked the last 4 and did some minor editing.

Issue #73, opened 2016-04-08
32fc_max_index_16u


n-west:
The QA seems to report 1-off for SSE3 impl and for some reason (probably historical) QA has a tolerance of 3. Fix the 1-off and set tolerance to 0.

n-west:
@trondeau FYI since you might be using these.

The complex versions are only able to return index_max within the width of the SSE3 vector because of this line
https://github.com/gnuradio/volk/blob/master/kernels/volk/volk_32fc_index_max_16u.h#L134

The real versions make clever use of SSE 4.1 blendv instructions with https://github.com/gnuradio/volk/blob/master/kernels/volk/volk_32f_index_max_16u.h#L85 to get the actual index_max. The complex versions should remove the sse3 and replace it with an sse3 + sse 4.1 kernel (SSE3 is needed for the hadd to get magnitude without doing weird shuffling).

Since protokernel names aren't exported we can make the change whenever is convenient. I won't hold up your PR for this unless you say you'll fix it :-)

n-west:
This is more critical now that @gnuradio/packet3 work is being merged in

n-west:
fixed

PR #74, opened 2016-04-09
volk_32fc_index_max_16u

trondeau:
changed output variable from unsigned int to unsigned short to match the function name.


n-west:
Seems unlikely, but does this fix https://github.com/gnuradio/volk/issues/73 ?

trondeau:
Probably unrelated to #73.

I force updated the PR with the same fix to volk_32f_index_max_16u. However, we might want to be careful because we're iterating over an int but the target is now a short. Might need some protection against overflowing?

n-west:
Yes, seems like this should either cast num_points to a short/uint16_t or loop on a condition like `number < MIN(num_points, USHRT_MAX)` (where the min is obviously only taken once).

Also, while we're making type consistency changes I think uint16_t is preferred over short. Grepping through existing code it doesn't seem consistent one way or the other on using stdint types, but since we explicitly list bit size in kernel names I'd like new code to use stdints.

trondeau:
Ok, changed the data types to stdint types, fixed some issues, limiting num_points in 16u version, and added a 32u version.

Let me know if you want any of this squashed.

PR #77, opened 2016-04-23
Ported volk gnsssdr kernels

carlesfernandez:
New kernels:
- volk_32fc_convert_16ic: generic, SSE2 and NEON implementations
- volk_16ic_convert_32fc: generic, SSE2, AVX and NEON implementations
- volk_16ic_x2_multiply_16ic: generic, SSE2, AVX2 and NEON implementations
- volk_16ic_x2_dot_prod_16ic: generic, SSE2, AVX and NEON implementations

This pull-request also contains some small changes in qa_utils.cc for handling 16ic kernels.


n-west:
FYI, this is held up at this point because of C99 issues. We currently don't set a C standard because the only way to do so with cmake appears to be non-portable. That means the default for some compilers that we care about supporting is C89 which doesn't support declaring variables inside the for loop

n-west:
Relevant GR PR: https://github.com/gnuradio/gnuradio/pull/825

carlesfernandez:
Commit 65ec4a3 seems to fix the issue

PR #82, opened 2016-05-23
An AVX and generic complex division proto-kernel

git-artes:
As agreed with Marcus Müller, I'm posting a pull request regarding the AVX and generic implementation of the complex division kernel. 
Regarding the implementation, the number of registers it uses may be much less, but I've seen no gain in performance in doing so, so I've prioritized readability.


n-west:
SSE3 needs to be put in a new PR

git-artes:
OK. I understand that @marcusmueller should do such PR, since he authored the SSE3 kernel.

PR #83, opened 2016-05-25
BOOST_ROOT handling

hades-:
Don't override the BOOST_ROOT environment variable. This change allows BOOST_ROOT to be used as either an environment variable or a cmake argument.


hades-:
Close this request... It has changes from the maint branch as well as my change. I'll redo it cleanly.

PR #87, opened 2016-05-31
Don't override the BOOST_ROOT environment variable.

hades-:
No reason to override the BOOST_ROOT environment variable when building.


n-west:
merged

PR #94, opened 2016-07-14
Port templates from python-cheetah to python-mako

stwunsch:


stwunsch:
Note: mako version needs to be set on the minimum possible version

n-west:
merged in to master-mako. This will be the start of volk 2.0!

PR #100, opened 2016-10-06
Volk snaps

guruofquality:
I needed to make some minor changes to package volk into a snap package.
- The cmake-based build module sets a blank CMAKE_INSTALL_PREFIX (it does destdir installs). The fix was to quote the prefix escape code so it always sees the same number of arguments. My personal preference would have been to use `file(TO_CMAKE_PATH prefix "${prefix}")` to keep the paths unix-style inside the code, anyway you would still need the quotes in this case.
- Snaps are relocatable, we we need a dynamic install prefix. I added support for a "VOLK_PREFIX" environment variable. I think the use of this prefix is limited at the moment, but at least it will be correct and future-proofed. This is also helpful for the PothosSDR windows installer which also has relocatable installs based on the user's selection.
- Const correctness: The constants functions should really be retuning `const char*`, and also because we would have to cast the result of getenv() to `char *` to avoid the compiler warning/error.

### DIFF
--- a/include/volk/constants.h
+++ b/include/volk/constants.h
@@ -27,11 +27,11 @@
 
 __VOLK_DECL_BEGIN
 
-VOLK_API char* volk_prefix();
-VOLK_API char* volk_version();
-VOLK_API char* volk_c_compiler();
-VOLK_API char* volk_compiler_flags();
-VOLK_API char* volk_available_machines();
+VOLK_API const char* volk_prefix();
+VOLK_API const char* volk_version();
+VOLK_API const char* volk_c_compiler();
+VOLK_API const char* volk_compiler_flags();
+VOLK_API const char* volk_available_machines();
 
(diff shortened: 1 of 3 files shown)


n-west:
This is pretty cool. What's your experience with snaps been so far?

n-west (approved):

guruofquality:
Thanks for the merge. Snaps have bit of a learning curve, but its certainly been simpler than say debian packaging. I think its a reliable way to deliver an end-use case application like a gui or a server with all of the dependencies in order. I'm not so sure about deploying development libraries this way. That said, there is a pretty nice permissions system for isolating snaps from the surrounding system, but it needs some more work for supporting devices (example libusb). Not trying to shamelessly self-promote, but since you asked, here is a blog I put together on the subject: https://myriadrf.org/blog/snap-packages-limesdr/ :-)

PR #102, opened 2016-12-20
Build on python3 system

djanderson:
This pull request adds support for systems whose default python is python3.
It retains backwards compatibility with Python 2.7.

In a few places, I used "six" to maintain compatibility. This is the same tool that will be required in gnuradio's 3.8 branch to add python3 support.

### DIFF
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -41,7 +41,7 @@ else(PYTHON_EXECUTABLE)
 
     #and if that fails use the find program routine
     if(NOT PYTHONINTERP_FOUND)
-        find_program(PYTHON_EXECUTABLE NAMES python python2 python2.7 python2.6 python3)
+        find_program(PYTHON_EXECUTABLE NAMES python python2 python2.7 python3)
         if(PYTHON_EXECUTABLE)
             set(PYTHONINTERP_FOUND TRUE)
         endif(PYTHON_EXECUTABLE)
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -36,11 +36,12 @@ if(PYTHON_EXECUTABLE)
 else(PYTHON_EXECUTABLE)
 
     #use the built-in find script
+    set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6)
     find_package(PythonInterp 2)
 
     #and if that fails use the find program routine
     if(NOT PYTHONINTERP_FOUND)
-        find_program(PYTHON_EXECUTABLE NAMES python python2 python2.7 python2.6 python2.5)
+        find_program(PYTHON_EXECUTABLE NAMES python python2 python2.7 python2.6 python3)
         if(PYTHON_EXECUTABLE)
             set(PYTHONINTERP_FOUND TRUE)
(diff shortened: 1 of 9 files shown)


cmake/Modules/VolkPython.cmake
```diff
@@ -36,11 +36,12 @@ if(PYTHON_EXECUTABLE)
 else(PYTHON_EXECUTABLE)
 
     #use the built-in find script
+    set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6)
     find_package(PythonInterp 2)
 
     #and if that fails use the find program routine
     if(NOT PYTHONINTERP_FOUND)
-        find_program(PYTHON_EXECUTABLE NAMES python python2 python2.7 python2.6 python2.5)
+        find_program(PYTHON_EXECUTABLE NAMES python python2 python2.7 python2.6 python3)
```
jmcorgan: Since our 3.8 branch will drop support for Python 2.6, this could be removed here.

jmcorgan (commented):
It's not clear whether the changes in the cmake file(s) are backwards compatible with python2; unless those code fragments are included in something that imports print_function, etc.

cmake/Modules/VolkPython.cmake
djanderson: In these cases, the parens are syntactically irrelevant in py2. I added the `print_function` import in the python files not because they were particularly necessary, but because it will now throw an error if someone tries to use the older style print statement. We could add the `print_function` import into the cmake file snippet but it seems a bit much for a 1-liner.

Let me know and I'll update my pull request tomorrow. Thanks for the review.

djanderson:
In these cases, the parens are syntactically irrelevant in py2. I added the print_function import in the python files not because they were particularly necessary, but because it will now throw an error if someone tries to use the older style print statement. We could add the print_function import into the cmake file snippet but it seems a bit much for a 1-liner.

Let me know and I'll update my pull request tomorrow. Thanks for the review.

jmcorgan:
What I mean, is, will those code fragments in the cmake files get executed when cmake is run on a python2 only system? And if so, will they cause errors?

djanderson:
The 3 changes I made in the .cmake file--putting parens around the print, passing the byte string literal to hashlib, and the floor division operator--are all valid in python 2.7+, but necessary in python 3. The statements will now produce the expected output regardless of python version (within the supported range of 2.7+).

```bash
$ python -c "from distutils import sysconfig; print(sysconfig.get_python_lib(plat_specific=True, prefix=''))"
lib/python2.7/dist-packages
$ python3 -c "from distutils import sysconfig; print(sysconfig.get_python_lib(plat_specific=True, prefix=''))"
lib/python3/dist-packages
$ python -c "import hashlib; print(hashlib.md5(b'asdf').hexdigest()[:5])"
912ec
$ python3 -c "import hashlib; print(hashlib.md5(b'asdf').hexdigest()[:5])"
912ec
$ python -c "print(5//2)"
2
$ python3 -c "print(5//2)"
2
```

jmcorgan:
Got it, thanks.

n-west:
I went ahead and merged in to master. This does mean that gnuradio won't be able to use the next master release until next, but that's OK. I'll blow away the master-mako branch.

PR #103, opened 2017-01-15
python3: mako template doesn't seem to understand map generator

djanderson:
My previous PR contained a regression when run with py3 that I just caught:

At the line: https://github.com/gnuradio/volk/blob/master/tmpl/volk_machine_xxx.tmpl.c#L46

it appears that if `kernels` is any iterator it should work, but for some reason the map iterator returned by python3's `map` function is not iterated by mako.

This commit forces both py2 and py3 to evaluate to a list.

### DIFF
--- a/gen/volk_kernel_defs.py
+++ b/gen/volk_kernel_defs.py
@@ -205,7 +205,7 @@ class kernel_class:
 __file__ = os.path.abspath(__file__)
 srcdir = os.path.dirname(os.path.dirname(__file__))
 kernel_files = glob.glob(os.path.join(srcdir, "kernels", "volk", "*.h"))
-kernels = map(kernel_class, kernel_files)
+kernels = list(map(kernel_class, kernel_files))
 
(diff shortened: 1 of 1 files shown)


n-west:
This doesn't pass QA on my machine. I get a bunch of incompatible pointer types for the function pointers and then segfaults during QA.

Also, when you fix it can you squash changes? It looks like there's 2 commits to change 1 line here.

carlesfernandez:
Leaving ```list``` instead of ```set``` is working for me in python27 and python3. So commit ada0f6301d42ff5b2dd158fb1be33450b51a6985 seems the right one.

djanderson:
@carlesfernandez, thanks for testing that.

I did list->set to keep a consistent coding style w/ the other calls to `map` in the file, and it looked like the iteration for `kernels` in the c template I linked wasn't order dependent, but I must have missed something.

Anyway, I nuked that change and tests are passing again.

PR #112, opened 2017-04-01
Use 'powf' instead of 'pow'

michaelld:
... to match variables and avoid implicit type conversion. Makes some older compilers happy, allowing 'make test' to pass.


michaelld:
Anyone want to comment on this PR? @n-west

PR #132, opened 2017-12-04
Add pip install and boost to env

jsonzilla:


n-west:
Did you intend to close and delete branch so quickly? At first glance this looks like it will be needed eventually. I think @guruofquality was using appveyor... is that still kicking?

jsonzilla:
I made an incorrect adjustment, I'll read the treads of the releases in order to understand how to work properly with the appveyor. Can not get it to find the correct boost version for the build

PR #133, opened 2017-12-05
Update the appveyor config file

jsonzilla:
Use appveyor boost lib
Use pip to install other dependices

Now it is possible to perform the build, so much that it is possible to observe a linkage problem. [https://ci.appveyor.com/project/0unit/volk](https://ci.appveyor.com/project/0unit/volk)


jsonzilla:
I have a mistake in pass the path for linking

PR #134, opened 2017-12-06
Disable SSE when compile vs2013 or higher

jsonzilla:
So far it has been the way I found to be able to compile using Visual Studio, without disabling SSE to 64 bit, it apprehends a linkage error.

I still do not know the general structure of the project, so I did it in my view was palliative, just to allow me to use in my project.

I also adjusted the file used for the appveyor build to make it easier to see the behavior I mentioned above. https://ci.appveyor.com/project/0unit/volk

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,21 +2,15 @@ clone_depth: 1
 
 os: Visual Studio 2013
 
-install:
-    - echo "Installing Boost libraries..."
-    - nuget install boost_system-vc120
-    - nuget install boost_filesystem-vc120
-    - nuget install boost_chrono-vc120
-    - nuget install boost_program_options-vc120
-    - nuget install boost_unit_test_framework-vc120
+environment:
+    BOOST_ROOT: C:\Libraries\boost_1_59_0
+    BOOST_LIBRARYDIR_WIN64: C:\Libraries\boost_1_59_0\lib64-msvc-12.0
 
+install:
     - echo "Installing Cheetah templates..."
-    - appveyor DownloadFile https://pypi.python.org/packages/source/C/Cheetah/Cheetah-2.4.4.tar.gz
-    - 7z x Cheetah-2.4.4.tar.gz
-    - 7z x -y Cheetah-2.4.4.tar
-    - cd Cheetah-2.4.4
-    - c:\Python27\python.exe setup.py build
-    - c:\Python27\python.exe setup.py install
+    - pip install Cheetah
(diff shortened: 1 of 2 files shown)


n-west:
@guruofquality do you have any opinions on this? I like referring to you regarding windows and msvc matters :-)

guruofquality:
Not entirely sure about the change to lib/CMakeLists.txt. The little over-rule mmx section was due to mmx intrinsics just not compiling on 64-bit, so we kill the arch entirely. @0unit did sse intrinsics functions fail to build on msvc2013 64-bit compile? If so, thats sort of how I read that change.

FWIW, I have been building volk maint branch on msvc 2015 64-bit, and I think that we have basically avoided building sse and sse2 arches because the flags check should have failed and prevented those arches from being enabled. So just generic and avx arches would have been compiled, unless we did something special with the flag checks for 64-bit.

(but forgive my recollection if I got any of that wrong)

So here is the run-down in terms of flag differences between 64 and 32 bit builds, so maybe we are doing this approach wrong. Here are the flags:

* 64-bit builds arch flags: https://msdn.microsoft.com/en-us/library/jj620901(v=vs.140).aspx `/arch:[AVX|AVX2]  `
* 32-bit builds arch flags: https://msdn.microsoft.com/en-us/library/7t5yh4fd.aspx `/arch:[IA32|SSE|SSE2|AVX|AVX2]  `

Basically 32-bit builds have your normal sse and sse2 flags, and in addition avx and avx2. But 64-bit on the other hand only has the avx/avx2. It looks as if sse2 is basically implied, so we should be able to build sse and sse2 arches, we just dont need the flag check because its always enabled, in-fact the flag check would be disabling a useful arch in this case (I believe).

In other words, imagine the arch xml file looked more like this to account for the flag differences between build types 32 vs 64:

```
<arch name="sse">
  #<flag compiler="msvc64"></flag> //disabled
  <flag compiler="msvc32">/arch:SSE</flag>

<arch name="sse2">
  #<flag compiler="msvc64"></flag> //disabled
  <flag compiler="msvc32">/arch:SSE2</flag>

<arch name="avx">
  <flag compiler="msvc64">/arch:AVX</flag>
  <flag compiler="msvc32">/arch:AVX</flag>
```

noc0lour:
Hm, so the change in `lib/CMakeLists.txt` is unecessary you say? @guruofquality

noc0lour:
Merged in #239

PR #140, opened 2018-01-28
Replaced math.h double functions by their float equivalent

marcusmueller:
Where applicable. In hopes it speeds up things.

Note that I intentionally did not update copyright years, as I consider this
changeset to be too small for that.

### DIFF
--- a/include/volk/volk_complex.h
+++ b/include/volk/volk_complex.h
@@ -47,6 +47,11 @@ template <typename T> inline T lv_conj(const T &x){
 
 #else /* __cplusplus */
 
+#if __STDC_VERSION__ >= 199901L /* C99 check */
+/* this allows us to conj in lv_conj without the double detour for single-precision floats */
+#include <tgmath.h>
+#endif /* C99 check */
+
 #include <complex.h>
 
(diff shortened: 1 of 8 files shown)


guruofquality:
Something I actually learned about just recently:
* tgmath http://en.cppreference.com/w/c/numeric/tgmath
* and generic: http://en.cppreference.com/w/c/language/generic

So the original `cos` function may actually be a macro calling the correct `cosf` function based on the data type.

marcusmueller:
@guruofquality never ceases to amaze :D

I must admit I didn't check whether we're using tgmath or math.h ; will do that later.

marcusmueller:
realization: tgmath is C99, not sure we can use that :(

marcusmueller:
So, what I did is the following: Check for C99 in the place where lv_conj etc are defined; if we've got C99, use tgmath to avoid using the double versions. Still, in all places where we can know that we're dealing with `float` (instead of `double`), I've replaced the function with its `f` variant – so that pre-C99 compilers do benefit from the same optimizations (and also, C++ programs including VOLK headers).

PR #142, opened 2018-01-29
test CI

noc0lour:
Will it build?

### DIFF
--- a/README
+++ b/README
@@ -1,3 +1,4 @@
+test CI
 #
 # Copyright 2015 Free Software Foundation, Inc.
 #


n-west:
yep :-) https://github.com/n-west/volk/blob/master/.travis.yml

noc0lour:
Was just wondering about the F26 test failures in GR, but it was the nested volk qa failing :/, ubuntu 14.04 dosen't even build with volk master, which is a non issue, I think. 

https://ci.gnuradio.org/buildbot/#/builders/24/builds/6

P.S. was just wondering how hard it is to integrate volk building into the existing CI

n-west:
https://www.youtube.com/watch?v=bVURsxuwy_E

I do intend to commit that travis-ci stuff. I guess I lost track that we had a build bot type thing.

n-west:
The current status is that maint is waaaaaaaaay overdue on a release to the point that I psuedo-released it on my personal github, but never pushed it to github/volk. My master branch I intend to release a new major revision when I abandon libbboost as a dependency so that it's just C and some minimal python that's required to build.

noc0lour:
Travis is fine if it works for volk. For GNU Radio travis is killing off the build because of build times.

noc0lour:
So, the buildbot will pick up changes in volk PRs now and build & test for Ubuntu 16.04 & 14.04 & F26 which have their default compilers and tools.

noc0lour:
It will not be displayed for now.

noc0lour:
Adding travis to that shouldn't be an issue though.

Issue #146, opened 2018-02-09
volk_profile should quit after the help message when the flag --help is passed


carlesfernandez:
Using the master branch, when I execute
```
$ volk_profile --help
```
it displays the help message and then executes the profiler. This option should only display the message and exit the program.

n-west:
I merged your stuff and also added some features to the new option parser to do the check for us-- hopefully that works as expected. Thanks!

PR #147, opened 2018-02-09
Exit volk_profile after showing the help message

carlesfernandez:
Simple fix for #146

### DIFF
--- a/apps/volk_profile.cc
+++ b/apps/volk_profile.cc
@@ -67,6 +67,13 @@ int main(int argc, char *argv[]) {
     profile_options.add((option_t("json", "j", "Write results to JSON file named as argument value", set_json)));
     profile_options.add((option_t("path", "p", "Specify the volk_config path", set_volk_config)));
     profile_options.parse(argc, argv);
+    
+    for (int arg_number = 0; arg_number < argc; ++arg_number) {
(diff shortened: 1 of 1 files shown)


apps/volk_profile.cc
```diff
@@ -67,6 +67,13 @@ int main(int argc, char *argv[]) {
     profile_options.add((option_t("json", "j", "Write results to JSON file named as argument value", set_json)));
     profile_options.add((option_t("path", "p", "Specify the volk_config path", set_volk_config)));
     profile_options.parse(argc, argv);
+    
+    for (int arg_number = 0; arg_number < argc; ++arg_number) {
+        if (std::string("--help") == std::string(argv[arg_number]) ||
+            std::string("-h") == std::string(argv[arg_number])) {
```
marcusmueller, line 73: Albeit a simple method of checking for an argument: 

Since `option_list` handles `-h`||`--help` specially internally, I'd argue that adding a member `helping` to the class that's set to `true` iff the `parse` hits that argument in https://github.com/gnuradio/volk/blob/9955b620641244c62dfed65c6de1d0b89b38b922/apps/volk_option_helpers.cc#L111-L114 wouldn't be less beautiful than this; it would reduce your patch to this file to `if(profile_options.helping) { return 0; }`.

PR #150, opened 2018-02-13
Includes for GCC 7.2 compatibility

Ka-zam:
Just tried VOLK on a fresh Ubuntu 18.04 install and wouldn't compile with GCC 7.2. These includes fixed it.

### DIFF
--- a/apps/volk_option_helpers.h
+++ b/apps/volk_option_helpers.h
@@ -6,6 +6,8 @@
 #define VOLK_VOLK_OPTION_HELPERS_H
 
 #include <string>
+#include <cstring>
+#include <limits.h>
 #include <vector>
 #include <map>
 


marcusmueller:
That's relatively surprising, as we usually do test on GCC 7.2; whilst these amends look pretty reasonable, I'd really love to see the error that happens without them, whilst I'm still trying to figure out how to make a Ubuntu 18.04 build container¹.

¹<sub>why, o why, must Ubuntu once again ship multiple versions of Boost? Why is libgsl23 in truth GSL 2.4? What do Canonical maintainers actually do professionally?`</rant>`</sub>

apps/volk_option_helpers.h
```diff
@@ -6,6 +6,8 @@
 #define VOLK_VOLK_OPTION_HELPERS_H
 
 #include <string>
+#include <cstring>
```
marcusmueller, line 9: I'm personally a big fan of ICWY (Include what you use), where you use it, so in my world, this would be included in `volk_option_helpers.cc` itself, not in its symbol- and type-defining header.

noc0lour:
Btw. this is probably not a `GCC 7.2` problem, Fedora & my own machines build volk just fine with the current state. More like a `Ubuntu 18.04` special snowflake problem.

marcusmueller:
Well, cstring should be included when you use `strncmp`, so I wonder why we're not seeing fails there. So, in that case, Ubuntu *is* a snowflake, but a probably more correct one, in this case.

noc0lour:
Ah, include-what-you-use I remember a great tool: https://gist.github.com/noc0lour/6de146ccab8447fa6e148b294f01cdb4 , though your technically right about the `<cstring>` which is drive-by included by the compiler nowadays (or is it -std=gnu ?). Also `<cstdlib>` is missing for using `atoi` and `atof`.

Edit: after I rerun compilation with headers fixed by `iwyu` tool and missing `cstring` and `cstdlib` and `limits.h` it fails .... wow, shuffling headers does a lot (: I'll submit a comprehensive PR I think

Ka-zam:
These are the error given when limits.h and cstring.h are not included:


`volk_option_helpers.cc:66:27: error: ‘INT_MIN’ was not declared in this scope
             int int_val = INT_MIN;
                           ^~~~~~~

/src/volk/apps/volk_option_helpers.cc:66:27: note: suggested alternative: ‘INT8_MIN’
             int int_val = INT_MIN;
                           ^~~~~~~
                           INT8_MIN

/src/volk/apps/volk_option_helpers.cc:103:38: error: ‘strncmp’ was not declared in this scope
                                 if ((strncmp(next_arg, "-", 1) == 0) || (strncmp(next_arg, "--", 2) == 0)) {
                                      ^~~~~~~

/home/magnus/src/volk/apps/volk_option_helpers.cc:103:38: note: suggested alternative: ‘wcsncmp’
                                 if ((strncmp(next_arg, "-", 1) == 0) || (strncmp(next_arg, "--", 2) == 0)) {
                                      ^~~~~~~
                                      wcsncmp

`

Seems reasonable to include cstring when using strncmp as previously noted.

PR #156, opened 2018-02-14
Added 32bit reversal kernel in multiple generic variants

marcusmueller:
These variants get heavily optimized and vectorized by my compilers, so
I actually don't think implementing intrinsic-based specialization
kernels pays at this very point in time.

The different variants still lend themselves greatly to such an
operation, so I'll reserve that for later. Should be a piece of cake!

The interesting part is that
https://github.com/gnuradio/gnuradio/issues/1610 is right, the method
currently used in gr-pager for bit-reversal is among the worst (on my
laptop, actually the worst with some distance to the second worst)
implementation.

Will integrate into GR soon.

generic unaligned implementations are:

* a simple 32bit-bitfield that gets manually reversed
* four simple 8bit bitfields, getting manually reversed
* a 256-element uint8_t[] LUT
* the 1972 "great trick" that gr-pager uses, including a %1023 op
* the 2001 "improved great trick" from Bit twiddling hacks
* an idea derived from binary trees: first swap the 16 bit halves,
  then swap 8bit-tuples, then swap 4bit-tuples, then swap 2bit-tuples,
  then consecutive bits.
* the same idea, but first swap the consecutive bits, then 2bit-tuples,…

The interesting aspect is that these implementations actually behave
pretty differently on different machines, as it seems, but not very
differently on the same machine between Clang4 and GCC7.

Hence, I'm keeping the large number of generic implementations (seems
silly at first sight); I haven't even benchmarked this on 32bit or ARM.


n-west:
That's a lot of bits. Any reason for chosing revert over reverse?

marcusmueller:
I'm a non-native speaker and was confused? Should I convert to reverse?

n-west:
Ah, ok. Wasn't sure if revert was more meaningful than reverse. Perhaps umkehren, although I don't actually know if that makes sense because I'm a non-speaker.

marcusmueller:
yep, "umkehren"; I'm pretty sure I meant "reverse", not "revert".

Issue #157, opened 2018-02-16
Broken MSVC compile due to "not" in place of "!"


marcusmueller:
as @ptrkrysik reports in #143, the build on windows is currently broken.

ptrkrysik:
The commit 6e4d2a1d62f5f5016c7dec4fe78be2595b0b8653  by @n-west fixed the issue.

PR #158, opened 2018-02-16
Replacing 'not' with '!' to make the volk_profile.cc compile with MSVC

marcusmueller:
Fix for #157


n-west:
fixed on master. just did it myself because there were nasty conflitcs

PR #164, opened 2018-03-05
tests: split testcases in ctest for granualar failure reports

noc0lour:
Previously one would have to parse the output of ctest -V or ctest
--output-on-failure to understand which tests failed.

Leveraging integration into ctest we can use one executable and use
commandline arguments to call various subtestcases to integrate better
with ctest

### DIFF
--- a/cmake/Modules/VolkAddTest.cmake
+++ b/cmake/Modules/VolkAddTest.cmake
@@ -23,23 +23,39 @@ endif()
 set(__INCLUDED_VOLK_ADD_TEST TRUE)
 
 ########################################################################
+# Generate a test executable which can be used in ADD_TEST to call
+# various subtests.
+#
+# SOURCES        - sources for the test
+# TARGET_DEPS    - build target dependencies (e.g., libraries)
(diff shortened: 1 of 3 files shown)


noc0lour:
This patch assumes: 
 - Tests are named after `kernels/volk/*.h` since there is no `${volk_kernels}` variable in cmake
 - There is only one test case per kernel

With the second patch it will print to `cerr` if no test was run for a kernel, to avoid confusion.

hcab14:
Hi Andrej,

this looks good. I was thinking about something like this myself.

marcusmueller (approved):

hcab14:
@noc0lour:
I am not sure if this is now the case or if in future this may become relevant but can wildcards be easily added in case there is more than one test per kernel?

noc0lour:
If we need more than one test configuration per kernel, things have to be shuffled around to actually configure tests in CMakeLists.txt
But really every component can and should define one or more testcases in their test scope and then we run one test per kernel (top-level) which then can execute more than one test.

PR #176, opened 2018-04-03
AVX support for volk_32fc_x2_s32f_square_dist_scalar_mult_a

lexknuther:


n-west:
I admit the normal way kernels handle num_points or work_size is a tad awkward, but it looks like you are treating lv_32fc_t and float as the same size with this approach

kernels/volk/volk_32fc_x2_s32f_square_dist_scalar_mult_32f.h
```diff
@@ -84,6 +84,62 @@
...
+    upper_magnitudes_squared_top = _mm_permute_ps(upper_magnitudes_squared_top, 0x4E);
+    upper_magnitudes_squared = _mm256_insertf128_ps(upper_magnitudes_squared, upper_magnitudes_squared_top, 0);
+
+    __m256 ordered_magnitudes_squared = _mm256_blend_ps(lower_magnitudes_squared, upper_magnitudes_squared, 0xCC);
+    __m256 scalars = _mm256_set1_ps(scalar);
+    __m256 output = _mm256_mul_ps(ordered_magnitudes_squared, scalars);
+
+    _mm256_store_ps(target, output);
+    target += work_size;
+    points += work_size;
```
n-west, line 132: line in question. points +8 will skip 8 lv_32fc_t's

n-west:
Nevermind. I see you're loading up 2 registers worth of complexes... I'm sure you'll figure out why qa fails :-)

lexknuther:
I fixed the kernel.  The generic loop at the bottom to take care of the points that aren't a multiple of the work size had a buffer over-read and over-write.

lexknuther:
Yeah, gotta double that throughput from 4 at time to 8 at a time :-).

n-west:
That would do it!

PR #189, opened 2018-06-25
travis: add py3 job, fix job numbering

noc0lour:
This PR covers building volk with py3 on travis


noc0lour:
maybe we want to add `dist:xenial` to the list as well.

marcusmueller (approved):

noc0lour:
Need to remove the Xenial commit due to it not being available on travis.

PR #192, opened 2018-07-24
Fix memory error and add ASAN build target 

dlrobertson:
## Add ASAN Build Target

Allows building an address sanitized build with `-DCMAKE_BUILD_TYPE=ASAN`. See [The LLVM docs](https://clang.llvm.org/docs/AddressSanitizer.html) for details.

## Fix read out of bounds

After compiling an address sanitized build I ran the unit tests and hit a heap overflow, which looked like an off by one error. I'm sure there is a more elegant solution, and I'm more than happy to work on it. Any feedback is welcome :smile:

It looks like if `frame_size` is `66536`, `num_branches` is `4096`. On the last iteration of the loop storing chunks in the `frame_ptr`, `temp_ptr ==  temp_ptr + 16 + (4095 * 16) == temp + 66536` and the use of `_mm_loadu_si128` will read out of bounds.


n-west:
awesome-- thanks!

Issue #188, opened 2018-06-09
`32f_s32f_convert_8i`: off-by-one errors


marcusmueller:
Paul Boven writes:

> Hi everyone,
>
> I'm trying to perform 2 bit sampling of an RF signal. In one approach, 
> I'm using a float->char block, and noticed that around zero, a number of 
> float inputs become quantized in a bin that's one off from the correct 
> value. The ones that are wrong are always off by one, with their 
> quantization error always in the direction of zero.
> 
> The problem can be demonstrated with a really simple flowchart, using 
> the following blocks:
> 
> * Noise Source (Noise Type: Gaussian, Amplitude: 1, Seed: 0, Output 
> type: Float)
> * Throttle
> The throttle is then connected to two blocks:
> * A file-sink (Type Float) and a
> * 'Float to Char' block.
> * The float to char block is again connected to a File Sink, now of type 
> Char.
> 
> As an example, I've plotted all the samples that quantized as zero. 
> These should fall in the range [-0.5:0.5], but occasionally we also get 
> hits that lie within [-1:1]. These mishaps are rare (about one in 2000). 
> It also shows that they only occur at multiples of 8192 samples, and 
> zooming in reveals that they always happen shortly before the next 
> multiple of 8192, not after.
> 
> For other values than 0, the same applies, but the misquantizations are 
> only in one direction, ending up in a lower bin if the input is 
> positive, or in a higher bin if the input is negative. Again, the 
> misquantizations only occur in half the bin: For a value of 1, the float 
> value should be in [0.5:1.5], but occasionally a value in [1.5:2] also 
> ends up being quantized as 1.
> 
> This seems to me a bug that is somehow related to internal block 
> boundaries, but I'm not familiar enough with the internals of GnuRadio 
> to figure out just what's going wrong.
> 
> The problem does NOT occur when converting to Short or Int.
> 
> This is using GnuRadio 3.7.11 (as packaged with Ubuntu 18.04).
> 
> Regards, Paul Boven.

![float_to_char](https://user-images.githubusercontent.com/958972/41193483-cd08d664-6c0c-11e8-9e11-158e4f18ff7f.png)

marcusmueller:
Can reproduce: when setting the tolerance for the convert kernel to 0 in `kernel_tests.h`, we get multiple failures of ONLY the SSE2 and AVX2 kernel, both aligned and unaligned.

[...]

PR #198, opened 2018-09-20
convert: truncate rounding on avx,sse2 for float to integer to match the generic impl

noc0lour:
This actually fixes off-by-one errors allowed by previous test settings


noc0lour:
Apparently this was discussed in detail in #188 but nothing was done about that.

noc0lour:
Also adding documentation what is done (e.g. truncation for this PR) will help people to find out what's up.

noc0lour:
Let's do proper rounding, to avoid discontinuities around `0`.

PR #199, opened 2018-09-21
convert: Use proper rounding for float to int conversion

noc0lour:
Previously SIMD/generic implementations used different types 
of conversion (rounding/truncating) even in the same kernel. 

For DSP applications rounding is desirable to avoid the discontinuity around `0`. 
Since truncating will map (-1,1) to 0.


noc0lour:
Currently only the `s32f_32f_convert_i8` is converted to rounding mode. 
@n-west let me know if a change like that would be acceptable to resolve #188, especially since I've added a static function to abstract the per-sample conversion.

PR #206, opened 2018-10-08
added volk_8*_convert_*f

xloem:
based on volk_8u_s32f_convert_32f

These kernels perform direct conversion from 8-bit values to floating point.

Discussed briefly on IRC.


xloem:
(fails on aarch64, a platform I lack.  will set up travis on my fork, fix, and reopen)

Issue #210, opened 2018-10-23
Make test failed


bachnguyen93:
I ran `make test` for the source code version 1.4 from the repository in Xubuntu and received the following error
> The following tests FAILED:
> 3 - qa_volk_32f_log2_32f (Failed)
> 15 - qa_volk_16ic_magnitude_16i (Failed)
> 95 - qa_volk_32f_s32f_convert_32i (Failed)
> 109 - qa_volk_32f_x2_s32f_interleave_16ic (Failed)
> 120 - qa_volk_32fc_s32f_magnitude_16i (Failed)
> Errors while running CTest
> Makefile:105: recipe for target 'test' failed
> make: *** [test] Error 8

Then when I check the version of volk after a `sudo make install`, it gave me version 1.2.1 that was installed together with gnuradio, not the 1.4 that I installed manually from the source code. I also got the error below when trying to run `volk-config-info` or `volk_profile`

> error while loading shared libraries: libvolk.so.1.4.1git: cannot open shared object file: No such file or directory 

Did I miss something during the installation?

michaelld:
The `make test` results aren't a huge surprise. You can run `ctest -V -R qa_volk_32f_log2_32f` to get more specifics as to what's going on; ditto for the other tests. I see some of these tests fail too, due to the precision tolerance being a little too tight (e.g., to 7 decimal places instead of 4).

michaelld:
That's odd about the `volk` version being off. Maybe you have another install somewhere? Like, did you have one from the system package manager as well as the one you just installed, in different PREFIXes? That'd be my first guess.

Did you run `sudo ldconfig` after `sudo make install` ... that is generally required to pick up the new libraries.

noc0lour:
Since @n-west merged a PR of mine removing the `safeguards` from some tests to actually have some meaning `make test` fails.
Edit: It does so correctly. These are legitimate bugs.

michaelld:
Ah OK that makes some sense. I've seen more tests fail recently than in prior years. Which PR is it? I'm curious to look through the changes to see what they do & maybe how to fix things for some tests.

noc0lour:
https://github.com/gnuradio/volk/pull/199
for the conversion kernels (float -> int) it's quite straightforward to implement fixes. The log2f thing: not so much.

michaelld:
OK yeah those are good changes. I see how they might make testing fail. Not sure a good solution off the top of my head. Hopefully someone with more time than me will work things out.

[...]

Issue #205, opened 2018-10-08
NEON broken on raspbian


xloem:
It appears that after e98e9277409cd658f1a5b65ffaa0caab842bebce raspbian builds no longer include neon support.

It seems the cmake neon test is run without passing `-mfpu=neon` to the compiler, so it fails:
```
Performing C SOURCE FILE Test neon_compile_result failed with the following output:
Change Dir: /home/pi/src/volk/build/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_edc61/fast"
/usr/bin/make -f CMakeFiles/cmTC_edc61.dir/build.make CMakeFiles/cmTC_edc61.dir/build
make[1]: Entering directory '/home/pi/src/volk/build/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_edc61.dir/src.c.o
/usr/bin/cc    -Wall -Dneon_compile_result   -o CMakeFiles/cmTC_edc61.dir/src.c.o   -c /home/pi/src/volk/build/CMakeFiles/CMakeTmp/src.c
/home/pi/src/volk/build/CMakeFiles/CMakeTmp/src.c: In function ‘main’:
/home/pi/src/volk/build/CMakeFiles/CMakeTmp/src.c:2:43: warning: ‘dest’ is used uninitialized in this function [-Wuninitialized]
 int main(){ uint8_t *dest; uint8x8_t res; vst1_u8(dest, res); }
                                           ^~~~~~~~~~~~~~~~~~
/home/pi/src/volk/build/CMakeFiles/CMakeTmp/src.c:2:22: note: ‘dest’ was declared here
 int main(){ uint8_t *dest; uint8x8_t res; vst1_u8(dest, res); }
                      ^~~~
/home/pi/src/volk/build/CMakeFiles/CMakeTmp/src.c:2:43: warning: ‘res’ is used uninitialized in this function [-Wuninitialized]
 int main(){ uint8_t *dest; uint8x8_t res; vst1_u8(dest, res); }
                                           ^~~~~~~~~~~~~~~~~~
/home/pi/src/volk/build/CMakeFiles/CMakeTmp/src.c:2:38: note: ‘res’ was declared here
 int main(){ uint8_t *dest; uint8x8_t res; vst1_u8(dest, res); }
                                      ^~~
In file included from /home/pi/src/volk/build/CMakeFiles/CMakeTmp/src.c:1:0:
[...]

PR #215, opened 2018-11-23
ARMv7 compilation / tests with absolute tolerance

hcab14:
- make volk again compile on ARMv7 https://github.com/gnuradio/volk/issues/205
- `_mm*_cvtps_epi32` -> `_mm*_cvttps_epi32` in  `volk_32f_s32f_convert_32i` `volk_32f_x2_s32f_interleave_16ic` (_truncated_ float to int conversion) https://github.com/gnuradio/volk/issues/210
- added a test with an absolute tolerance for the `volk_32f_log2_32f` kernel https://github.com/gnuradio/volk/issues/210


noc0lour (approved):
Need review from someone else on the ARMv7 part. 
For the interleave kernel we might want to add to the documentation that this kernel is in truncating mode instead of rounding.

The test changes look good and reaasonable.

balister (commented):
Setting the mfpu flag in CMakelists isn't a great idea. This flag should be set in a toolchain file that knows for sure if the machine you are compiling for supports neon. I need to hunt up where it is set orginally and see what is going on

balister (changes requested):
Please drop the have_mfpu_neon pieces. You should enable neon using flags in the toolchain file. toolchains are hardware specific and this is the way we should handle neon detection. There is code in the cmake that tests from neon with compile time tests, so the flags must be set correctly.

hcab14:
Hi @balister 

> Please drop the `have_mfpu_neon` pieces. You should enable neon using flags in the toolchain file. toolchains are hardware specific and this is the way we should handle neon detection.

Ok. It is done.

But note that before ARMv8 was added, VOLK, did compile natively on ARMv7 devices without using a custom toolchain and now it does not, so this should be added somewhere to the documentation.

balister:
Sorry for the trouble. Inserting the mfpu flag likely caused trouble for the original Pi board and armv5 machines which do not have neon at. Hopefully, we are more consistent in the future.

hcab14:
Don't worry. Anyway, testing for `have_mfpu_neon` in lib/CMakeLists.txt felt somewhat unnatural.

PR #227, opened 2018-12-29
Adds neon intrinsics implementation of volk_32i_s32f_convert_32f

ast:


ast:
Found some further optimizations. Will send another PR when done.

PR #228, opened 2018-12-29
Adds neon intrinsics implementation of volk_32f_s32f_convert_32i

ast:


ast:
Found some further optimizations. Will send another PR when done.

PR #230, opened 2019-01-07
Change more kernels to rounding instead of truncating

noc0lour:
This will have the effect that the behaviour across all arches is the same and sane for 
DSP.


noc0lour:
The rounding in the orc implementation of 32fc_s32f_magnitude_16i should be fine since the result is always positive (and thus +0.5 and truncating will results in correct rounding).

PR #233, opened 2019-01-07
kernels: fix more rounding issues

noc0lour:
in 32f_s32f_convert_32i the AVX kernel needs to use the non-truncating
intrinsic and in the 32f_x2_s32f_interleave_16ic kernel the SSE2 kernel
needs to use the non-truncating intrinsic.


noc0lour:
@n-west quick, merge, as long as the tests pass ;)

PR #237, opened 2019-02-20
Fixes for complex conj-dot-prod kernel for both aligned and unaligned

mfogle:
Fixes for complex conj-dot-prod kernel for both aligned and unaligned
    avx2 architectures.  The incrementing of input and taps pointers
    performed by the for-loop to perform the conj-dot-prod operation on
    the residual number of quad points (i.e. when num_points is not a multiple
    of 4) was not correct.


noc0lour:
These bugs are already fixed with cf5d05e96297daf2614a6f5f62c1b44bb3819987. 
Thanks for submitting the pull request anyway!

PR #244, opened 2019-05-15
orc: remove broken 32fc_s32f_magnitude_16i kernel

noc0lour:


michaelld (approved):
Looks like this Orc kernel is a 1-off for some specific entries. I haven't looked further into it. At least on my OS testing, the only OSs where this kernel would be faster than the others is old Intel, before any real SIMD of any sort -- those are quite rare these days! Even on old OSX/PPC, the generic kernel is faster! On modern OSX/Intel, our SIMD kernels are faster. Hence +1 from me here!

PR #245, opened 2019-05-15
cmake: set cmake_c_standard and cmake_cxx_standard

noc0lour:


michaelld:
I've been using C++11 (not C11) for Volk release and devel for a long time in MacPorts ... but, my note about why is that it's anticipated that Volk will require C++11 in the future & it doesn't yet. I think this is still true: the Volk code does not technically require C11 or C++11. Is this correct? If so, why this change?

noc0lour:
It requires C11/C++11, at least GCC 4.8/4.9/5 was complaining about that in the CI.

michaelld:
OK; good to know. I see no real harm in requiring C/C++11. Those GCC versions are pretty old by now, so making sure there's a compatible compiler shouldn't be difficult on just about any OS from up to 6-8 years ago. I'd be mostly worried about Linux & Windows (if such a build exists). Anyone using GR on a Mac will be Brew or Fink or MacPorts, which will install a compatible compiler if/as necessary.

michaelld (approved):
LGTM

PR #246, opened 2019-05-20
Search for system-wide volk_config file

Zlika:
Hi!
The volk_config file generated by volk_profile has no reason to be different from one user to another on the same computer. So I think it would be interesting that volk also searches for the volk_config file in a system-wide location, i.e. /etc/volk/volk_config.
Regards,
Thomas

### DIFF
--- a/lib/volk_prefs.c
+++ b/lib/volk_prefs.c
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <unistd.h>
 #include <volk/volk_prefs.h>
 
 void volk_get_config_path(char *path)
@@ -16,6 +17,13 @@ void volk_get_config_path(char *path)
     if(home!=NULL){
         strncpy(path,home,512);
         strcat(path,suffix2);
+        return;
+    }
+
+    //check for system-wide config file
+    if (access("/etc/volk/volk_config", F_OK) != -1){
+        strncpy(path, "/etc", 512);
+        strcat(path, suffix2);
         return;
     }
 
--- a/lib/volk_prefs.c
+++ b/lib/volk_prefs.c
@@ -1,7 +1,13 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#if defined(_MSC_VER)
+#include <io.h>
+#define access _access
+#define F_OK 0
+#else
 #include <unistd.h>
+#endif
 #include <volk/volk_prefs.h>
 
 void volk_get_config_path(char *path)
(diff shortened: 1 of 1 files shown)


hcab14:
There is an environment variable called `VOLK_CONFIGPATH`, and you can set, _e.g._, `VOLK_CONFIGPATH=/etc/volk'. So in my opinion there is no need to change the code.

balister:
I disagree, I'd love to be able to install a default volk config on a
device and not depend on users remembering to run volk_profile.

For embedded boards, I install the config file for root, and I think it
would be better to install it for system wide use.

Zlika:
I share @balister 's point of view. I think VOLK_CONFIGPATH is just a workaround to deal with a non-standard behavior.

noc0lour:
Good idea overall. I'll review later.

michaelld:
I like this idea, too

balister (commented):
The system config file should only be used if a local config file isn't found. We do want users to be able to easily over ride the system wide file (and report back better system wide options)

michaelld:
> The system config file should only be used if a local config file isn't found. We do want users to be able to easily over ride the system wide file (and report back better system wide options)

Agreed! ... and that is what this change does: it checks for 2 other options before the system-wide one.

hcab14:
If I read the code correctly, the order is

1. `${VOLK_CONFIGPATH}/volk/volk_config`
2. `/etc/volk/volk_config`
3. `${HOME}/.volk/volk_config`
4. `${APPDATA}/.volk/volk_config`

michaelld:
OK yup I was mistaken there! What @hcab14 writes is correct. That's not desirable, in my opinion: the system-wide preference should be the final item. Do others have a preference?

Zlika:
I agree that system-wide preference should be after user preference. I don't know about APPDATA.

Zlika:
I reworked the PR to search for a system-wide config file only in last resort, and I also made a small change that should fix the Windows build.

michaelld (approved):
LGTM; thx!

noc0lour:
@Zlika do you already have a CLA with the FSF/GNU Radio? If not, please send an email to bhilburn@gnuradio.org and thanks for the effort. I should have asked you earlier

Zlika:
Hi @noc0lour. I don't have a CLA yet (I read that a CLA may not be necessary for such small contributions), I'll ask @bhilburn to sign it.

noc0lour:
Yep, your change is still small, but I don't know if it is small enough. @bhilburn  ?

bhilburn:
Unfortunately this changeset does cross the threshold of needing a CLA. @Zlika did already start the CLA process with us / the FSF yesterday, so it's underway, though =)

Zlika:
CLA signed!

Zlika:
Hi @bhilburn. I sent you the signed CLA 2 weeks ago. It is ok to merge this PR now?

bhilburn:
@Zlika - Yup! You're all set, here. 

@noc0lour @marcusmueller - clear to proceed as soon as you're ready. There are a couple of unit test failures, but they don't look specific to this PR?

noc0lour:
I refactored your changes into one fix & one system-config commit and pushed to master. Thanks!

PR #248, opened 2019-05-22
cmake: add back pre-"modern CMake targets" find variables

michaelld:
These variables were removed in beeb942ec7bd838ecf3e0e7f47688298d6057e8a:
+ VOLK_FOUND
+ VOLK_LIBRARIES
+ VOLK_INCLUDE_DIRS
and are used by the CMake script trying to find Volk. Without them, the finding script has to manually determine these variables, including whether Volk was even found .. and this should not be the case! Although GNU Radio was fixed to work without these variables, there are a variety of other projects that still rely on these variables to find and use Volk. This change adds these variables back into the CMake Volk find script. NOTE that the CMake targets are left untouched, and VOLK_LIBRARIES is set to the target library.

### DIFF
--- a/cmake/Modules/VolkConfig.cmake.in
+++ b/cmake/Modules/VolkConfig.cmake.in
@@ -3,3 +3,32 @@ get_filename_component(VOLK_CMAKE_DIR "${CMAKE_CURRENT_LIST_FILE}" PATH)
 if(NOT TARGET Volk::volk)
   include("${VOLK_CMAKE_DIR}/VolkTargets.cmake")
 endif()
+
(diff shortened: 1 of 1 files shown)


michaelld:
I'll admit that I'd love to just entirely jump from old cmake finding to modern targets ... it would be lovely! But virtually none of the projects that require Volk use modern targets yet, and they instead still use the old cmake find variables. So, we need these variables around for the interim -- and probably long term since some of these projects are likely to never be updated. It also will simplify the GR cmake code to go back to using at least VOLK_FOUND as provided herein ... as it could come from either the current Volk release or this devel version if this change is included.

PR #251, opened 2019-05-23
Boost: build without Boost and C++17

jdemel:
This commit enables us to build VOLK without Boost as a dependency. If
you set the `CMAKE_CXX_STANDARD` flag to 17, the build system will not
search for Boost anymore. In this case the STL `<filesystem>` will be
used. This include does only work with the latest compilers e.g. GCC9
and later. It does not work with GCC7 or Clang6.
This commit might help to avoid build problems that arise with Boost
1.70.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,7 +77,6 @@ endif(MSVC)
 ########################################################################
 # Dependencies setup
 ########################################################################
-
 # Python
 include(VolkPython) #sets PYTHON_EXECUTABLE and PYTHON_DASH_B
 VOLK_PYTHON_CHECK_MODULE("python >= 2.7" sys "sys.version.split()[0] >= '2.7'" PYTHON_MIN_VER_FOUND)
@@ -111,10 +110,14 @@ if(MSVC)
     endif(BOOST_ALL_DYN_LINK)
 endif(MSVC)
 
-include(VolkBoost)
+if(CMAKE_CXX_STANDARD GREATER_EQUAL 17)
+    set(Boost_LIBRARIES "")
+else()
+    include(VolkBoost)
 
-if(NOT Boost_FOUND)
-    message(FATAL_ERROR "VOLK Requires boost to build")
+    if(NOT Boost_FOUND)
+        message(FATAL_ERROR "VOLK Requires boost to build")
+    endif()
 endif()
 
 # Orc
--- a/README.md
+++ b/README.md
@@ -45,3 +45,10 @@ Build with cmake:
     $ sudo make install
 
 That's it!
+
+## BoostSANS
+
+It is possible to build VOLK without Boost. Though, this depends on the latest C++ compilers. Be aware that this requires GCC9 or later. In `CMakeLists.txt` change the line `set(CMAKE_CXX_STANDARD 11)` to `set(CMAKE_CXX_STANDARD 17)`.
+Rerun the build instructions.
+
+Voila! That's it!
--- a/apps/volk_profile.cc
+++ b/apps/volk_profile.cc
@@ -20,9 +20,16 @@
  * Boston, MA 02110-1301, USA.
  */
 
+
+// For C++17 '__cplusplus == 201703L'
+#if __cplusplus > 201700L
+#include <filesystem>
+#else
 #include <boost/filesystem/operations.hpp>   // for create_directories, exists
 #include <boost/filesystem/path.hpp>         // for path, operator<<
 #include <boost/filesystem/path_traits.hpp>  // for filesystem
+#endif
+
 #include <stddef.h>                          // for size_t
 #include <sys/stat.h>                        // for stat
 #include <volk/volk_prefs.h>                 // for volk_get_config_path
@@ -38,8 +45,11 @@
 #include "volk_option_helpers.h"             // for option_list, option_t
 #include "volk_profile.h"
 
-
+#if __cplusplus > 201700L
+namespace fs = std::filesystem;
+#else
 namespace fs = boost::filesystem;
+#endif
 
 volk_test_params_t test_params(1e-6f, 327.f, 131071, 1987, false, "");
 


michaelld:
I had no issues with currently Volk and Boost 1.70.0, once CMake correctly finds Boost. The issue is that Boost 1.70.0's CMake find scripts don't yet work properly. The Boost folks have already committed a fix upstream for those scripts. If one uses the CMake Boost find script with Boost 1.70.0 installed, things work just fine for CMake configuration.

michaelld:
FYI I love the idea of not requiring Boost for Volk! Even if it means using C++17!

jdemel:
It's good to know, that the issue can be fixed with some updated CMake magic. Why did it break in the first place?

michaelld:
Well, as it turns out, "mostly fixed". To use C++17's filesystem, libc++fs has to be installed too. I can compile the code, but I can't link it because OSX is missing this library. No idea why they didn't include fs in libc++.

michaelld:
Not sure what you mean by "Why did it break in the first place?"? If you mean Boost's CMake find scripts, it's because the Boost folks released them too soon. Otherwise, ?

jdemel:
Actually, I updated the `README` too, to clarify that one needs the latest compilers to compile VOLK without Boost.
I tested with GCC9, and it works. Also, if you set the standard to C++11, it'll automatically uses Boost. This patch does not work with GCC7 or Clang6 (default versions in Ubuntu 18.04).
Prior to GCC9, you'd need to tell the linker to explicitly link against `libc++fs`. In the GCC9 release notes you can find a remark that this linker flag is not necessary anymore. This fact motivated this patch.

I thought about explicitly detecting the compiler version. Though, the C++17 standard defines `<filesystem>` as part of the STL. Thus, I'd argue that if you want to use C++17, you'd need to use a fully standard compliant compiler. I tested this patch in a Docker Container with `ubuntu:rolling` (points to 19.04) and I explicitly installed GCC9/G++9. Also I told CMake to use this compiler.

@michaelld, that's exactly what I meant and what I was asking for. Thanks.

michaelld:
@jdemel Thank you for that background info; it truly is very useful! Apparently even the latest Clang doesn't integrate `libc++fs` yet ... I'll try GCC9 & see what happens there. It might be that we need more qualifiers beyond just c++17 to get this working!

noc0lour:
Hm, in theory It should build with libc++ and clang > 7.0 but it fails linking with `libc++fs.a` my cmake command:
```cmake ../ -DCMAKE_INSTALL_PREFIX=~/.usrlocal -DCMAKE_BUILD_TYPE=Debug -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -G Ninja -DCMAKE_CXX_FLAGS="-std=c++17 -stdlib=libc++" -DCMAKE_EXE_LINKER_FLAGS="-lc++abi -lc++fs"```

Also set cmake_cxx_standard to 17 in CMakeLists.txt.

Error:
```
[94/94] Linking CXX executable apps/volk_profile
FAILED: apps/volk_profile 
: && /usr/lib/ccache/bin/clang++  -std=c++17 -stdlib=libc++ -Wall -g  -lc++abi -lc++fs   -rdynamic apps/CMakeFiles/volk_profile.dir/volk_profile.cc.o apps/CMakeFiles/volk_profile.dir/__/lib/qa_utils.cc.o apps/CMakeFiles/volk_profile.dir/volk_option_helpers.cc.o  -o apps/volk_profile  -Wl,-rpath,/home/andrej/git/gnuradio/volk/build/lib: lib/libvolk.so.1.4.1git -ldl -lorc-0.4 -lm && :
/usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: apps/CMakeFiles/volk_profile.dir/volk_profile.cc.o: in function `write_results(std::__1::vector<volk_test_results_t, std::__1::allocator<volk_test_results_t> > const*, bool, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >)':
/usr/include/c++/v1/filesystem:1096: undefined reference to `std::__1::__fs::filesystem::path::__parent_path() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: apps/CMakeFiles/volk_profile.dir/volk_profile.cc.o: in function `std::__1::__fs::filesystem::exists(std::__1::__fs::filesystem::path const&)':
/usr/include/c++/v1/filesystem:1656: undefined reference to `std::__1::__fs::filesystem::__status(std::__1::__fs::filesystem::path const&, std::__1::error_code*)'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: apps/CMakeFiles/volk_profile.dir/volk_profile.cc.o: in function `write_results(std::__1::vector<volk_test_results_t, std::__1::allocator<volk_test_results_t> > const*, bool, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >)':
/usr/include/c++/v1/filesystem:1096: undefined reference to `std::__1::__fs::filesystem::path::__parent_path() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: /usr/include/c++/v1/filesystem:1096: undefined reference to `std::__1::__fs::filesystem::path::__parent_path() const'
/usr/lib/gcc/x86_64-pc-linux-gnu/8.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: apps/CMakeFiles/volk_profile.dir/volk_profile.cc.o: in function `std::__1::__fs::filesystem::create_directories(std::__1::__fs::filesystem::path const&)':
/usr/include/c++/v1/filesystem:1587: undefined reference to `std::__1::__fs::filesystem::__create_directories(std::__1::__fs::filesystem::path const&, std::__1::error_code*)'
clang-7: error: linker command failed with exit code 1 (use -v to see invocation)
ninja: build stopped: subcommand failed.```

michaelld:
That's exactly what I see. I'm using `clang -v` of `Apple LLVM version 10.0.1 (clang-1001.0.46.4)` ... which should provide full c++17 compatibility. But `libc++fs` isn't available for some reason!

noc0lour:
You should have `libc++fs.a` available somewhere in `lib/`. `nm` shows that it contains definitions for the right symbols. But somehow the linker is not picking up this static lib.

jdemel:
I went through the [GCC9 change list](https://www.gnu.org/software/gcc/gcc-9/changes.html). And this is the line that let to this PR:

> Using the types and functions in `<filesystem>` does not require linking with -lstdc++fs now.

For Clang I found [this](https://libcxx.llvm.org/docs/UsingLibcxx.html).

> Starting with LLVM 9.0, support for `<filesystem>` is provided in the main library and nothing special is required to use `<filesystem>`.

That's the current state.

I thought about explicit checks for different compilers but `<filesystem>` is part of C++17. We  explicitly set the standard to C++11. And the pitfalls attached with going for C++17 are documented in the README. Thus, users who activate this feature are required to know what they do and how to handle this situation. I could imagine to add the links and description to the appropriate section in the README.

carlesfernandez:
Hi guys, 

inspired on this PR, I tried to automate things in #255. No way to make it work for macOS, though. I can't use std::filesystem even manually calling the compiler in a simple test with the supposed "right flags". It just seems not to be there.

n-west:
Try std::experimental::filesystem. Support for filesystem is wonky.
https://en.cppreference.com/w/cpp/experimental/fs/path

carlesfernandez:
I've tried that, both with AppleClang 10.0.1 and clang 8.0, without luck.

michaelld:
Clang just doesn't properly support libc++fs yet; not directly or via ::experimental:: ... it's a real PITA for those of us wanting to use that c++ feature!

n-west:
Agreed- I did this last year (https://github.com/n-west/volk/commit/bfd4f7858efd15c0f20abbeddc2f763f31caf7b9) and concluded that compiler support was a long enough way off that it wasn't worth pursuing. Kind of sad that it doesn't seem to be a whole lot better. The Apple provided clang was several versions old IIRC; however, clang in general does have support. I am not sure of newer releases of Apple clang.

https://libcxx.llvm.org/docs/UsingLibcxx.html#using-filesystem

michaelld:
How does this PR compare with https://github.com/gnuradio/volk/pull/555 ? Guessing this one can be closed ...

jdemel:
After #255 got merged, this PR is obsolete. Thus, I close it.

PR #255, opened 2019-06-11
Build Volk without Boost if C++17 is available

carlesfernandez:
A CMake module checks if std::filesystem is available. If so, it sets the C++ standard to 17 and Boost is not required anymore. No user intervention required.

This is an alternative solution to #251 but it does not require user's modification of the CMake script. It works for g++ and clang on Linux, not tested on Windows.

Note to macOS users: the std::filesystem detection successfully fails, since AppleClang seems to not be proving std::filesystem. The same for clang 8.0 installed with Macports.


carlesfernandez:
I've added std::experimental::filesystem to the search, which covers g++ versions older than 8.0 (e.g,  g++ 7.4 that comes with Ubuntu 18.04)

jdemel:
I like the fact that you check for `filesystem`. My thoughts for #251 basically were, that you'd use a recent compiler with C++17 support anyways if you want these features. Thus, we don't need to check for `filesystem`.
Clang8/GCC9 fully support `std::filesystem`. Thus, I'd assume we can rely on these compilers. What about earlier versions? Do we need to fear they might cause trouble? Actually in this case, I'd vote for not supporting these compiler versions.

Does this patch set the required linker arg for GCC7 etc.?

jdemel:
I had some time to play around with your PR. It works well on my machine. I just add some more comments.

CMakeLists.txt
```diff
@@ -23,10 +23,14 @@ set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "Choose build type: None D
 project(volk)
 
 enable_language(CXX)
-set(CMAKE_CXX_STANDARD 11)
-
 enable_language(C)
-set(CMAKE_C_STANDARD 11)
+
+if(NOT (CMAKE_VERSION VERSION_LESS "3.1"))
```
jdemel: We should require CMake 3.8 right at the beginning of this CMakeLists.txt and thus we don't need to check for older versions anymore. Currently, this is still at 2.8.12 but I issued a PR to bump the minimum Cmake version to 3.8 which makes sense to align VOLK and GR requirements.

CMakeLists.txt
```diff
@@ -111,10 +115,29 @@ if(MSVC)
     endif(BOOST_ALL_DYN_LINK)
 endif(MSVC)
 
-include(VolkBoost)
+# Check if we have std::filesystem
+if(POLICY CMP0057)
+    cmake_policy(SET CMP0057 NEW)
+endif()
+
+if(CMAKE_VERSION VERSION_LESS 3.8)
```
jdemel: One more good reason to require CMake 3.8.

CMakeLists.txt
```diff
@@ -111,10 +115,29 @@ if(MSVC)
...
+    set(FILESYSTEM_FOUND FALSE)
+else()
+    find_package(FILESYSTEM COMPONENTS Final Experimental)
+endif()
 
-if(NOT Boost_FOUND)
-    message(FATAL_ERROR "VOLK Requires boost to build")
+#### Set C++ standard to 17 if std::filesystem is available
+if(${FILESYSTEM_FOUND})
+    set(CMAKE_CXX_STANDARD 17)
```
jdemel, line 119: You also set this value in `FindFILESYSTEM.cmake`. Is this required in both places?

cmake/Modules/FindFILESYSTEM.cmake
```diff
@@ -0,0 +1,240 @@
+#[=======================================================================[.rst:
```
jdemel, line 24: Did you copy this file from somewhere? In this case, do we need a special license header for this file?

jdemel (commented):
Thanks for your work. I prefer it over my solution. I guess this can be merged if we resolve the questions that arose.

CMakeLists.txt
carlesfernandez: I'm not sure, since the CMake module does:
```
include(CMakePushCheckState)
cmake_push_check_state()
...
cmake_pop_check_state()
```
so I'm not sure if the right compiler flags survive. This is why I included that line again.

cmake/Modules/FindFILESYSTEM.cmake
carlesfernandez: I took the file from https://github.com/vector-of-bool/CMakeCM/blob/master/modules/FindFilesystem.cmake , and then I did some modifications. The original license is BSD 3-Clause.  Should I include that in the header?

jdemel: Yeah. I think you should do so. My takeaway from other license header discussion is that we should put the GPL header first because some changes were made and in this case followed by `BSD 3-Clause`.

CMakeLists.txt
jdemel: In this case it should be in.

jdemel:
#280 just got merged. @carlesfernandez could you rebase this PR on current master? Then remove those CMake lines that check for CMake < 3.8? 
Together with this license header thing, this would make this PR ready to be merged. 

@michaelld what's your take on this PR?

CMakeLists.txt
michaelld: We've now set the min CMake version to 3.8 ... so please fix up this clause.

CMakeLists.txt
michaelld: ditto CMake version fixup

cmake/Modules/FindFILESYSTEM.cmake
michaelld: agreed with @jdemel !

michaelld (changes requested):
A few requested changes for CMake version fixes now that we're requiring at least 3.8.

I'll also need to do some testing on OSX, since IIRC the C++17 Filesystem feature is messed up on various Clang and AppleClang. Hopefully the .cmake script knows how to handle it!

michaelld:
I broadly like the idea of moving away from Boost, and as C++17 becomes more available and ubiquitous, I'm all for using it instead of Boost. That said, we do have to keep supporting Boost & thus make sure our code is both-ways compatible as we move forward. No Boost-specific nor C++17+-specific code here!

carlesfernandez:
I removed the code related to CMake < 3.8 and I've added the GPL header to the CMake module. I also included a clause to account for the AppleClang compiler, which provides std::filesystem starting from version 11.

cmake/Modules/FindFILESYSTEM.cmake
carlesfernandez: Please let me know if the header is now _how it should be_. In my understanding, it is ok to release under the GPL a derived work from a BSD-3 code, as long as you retain the original copyright notice. But I really would like to confirm with you guys that this is the right way of doing it, or if there is a more adequate wording for this case.

michaelld: I'd recommend including `See https://cmake.org/licensing for details.` as should have been in the original.

michaelld: The GPL header looks good to me; thanks!

carlesfernandez: Included `See https://cmake.org/licensing for details.` in f7e2d9bcbf84fd4151b8eb088c13f449abcd84d3

michaelld: @bhilburn In your understanding of licensing, is this header info (1) sufficient both for the GPL; (2) in keeping enough of the prior license as need be; (3) in allow Carles to note that he changed the original file? Thx!

michaelld: I'll give Ben through the end of this work-week to respond. If not, we'll go with this license as current & fix it up if/as need be later.

bhilburn: Yup, current version is good 👍

michaelld: great; thanks @bhilburn !

jdemel (approved):
Looks good to me now.

michaelld:
Initial testing with just C++11 works on Clang (meaning, it uses Boost). Moving up to C++14 and C++17 ...

michaelld:
How does this PR compare with https://github.com/gnuradio/volk/pull/251 ? Guessing this one is the one to concentrate on ...

jdemel:
This PR is an alternative to #251 . Basically it automates `std::filesystem` detection. I'll close #251 if this PR gets merged.

michaelld:
Ok sounds good. I stalled testing due to other work. I'll get back on testing hopefully tomorrow / Friday.

michaelld (approved):
The code reads well to me. I've tested this as best I can. I can't get std::filesystem to work on Clang.

michaelld:
CLA in place. CI passes. 2 approvals. Merging.

PR #263, opened 2019-07-30
Added versioning Document

marcusmueller:
Starting with version 2.0.0, VOLK adapts [Semantic Version](https://semver.org) as its versioning scheme.

To highlight a few core aspects as relevant to VOLK:

  * Adding a new kernel (e.g. `volk_32fc_prove_riemann_hypothesis`) is a backwards-compatible functionality addition. That requires increasing the MINOR version (clause 7).
  * Adding a new proto-kernel (e.g. `volk_32fc_prove_riemann_hypothesis_a_neon`) is _also_ a backwards-compatible functionality addition, and hence also increases MINOR version. It's appreciated to bundle these.
  * Fixing a bug in a proto-kernel that doesn't change the function signature nor intended behavior only requires MINOR to be increased.
  * Changing what a kernel intentionally does, or aspects the use of the dispatcher, addition of a new architecture or other fundamental changes increase the MAJOR version.

### DIFF
--- a/versioning.md
+++ b/versioning.md
@@ -1,0 +1,279 @@
+Versioning
+==========
+
+Starting with version 2.0.0, VOLK adapts [Semantic Version](https://semver.org) as its versioning scheme.
+You'll find the full text of Semver 2.0.0 below.
+
+To highlight a few core aspects as relevant to VOLK:
+
+  * Adding a new kernel (e.g. `volk_32fc_prove_riemann_hypothesis`) is a backwards-compatible functionality addition. That requires increasing the MINOR version (clause 7).
+  * Adding a new proto-kernel (e.g. `volk_32fc_prove_riemann_hypothesis_a_neon`) is _also_ a backwards-compatible functionality addition, and hence also increases MINOR version. It's appreciated to bundle these.
+  * Fixing a bug in a proto-kernel that doesn't change the function signature nor intended behavior only requires PATCH to be increased.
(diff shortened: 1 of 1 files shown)


versioning.md
```diff
@@ -0,0 +1,279 @@
...
+==========
+
+Starting with version 2.0.0, VOLK adapts [Semantic Version](https://semver.org) as its versioning scheme.
+You'll find the full text of Semver 2.0.0 below.
+
+To highlight a few core aspects as relevant to VOLK:
+
+  * Adding a new kernel (e.g. `volk_32fc_prove_riemann_hypothesis`) is a backwards-compatible functionality addition. That requires increasing the MINOR version (clause 7).
+  * Adding a new proto-kernel (e.g. `volk_32fc_prove_riemann_hypothesis_a_neon`) is _also_ a backwards-compatible functionality addition, and hence also increases MINOR version. It's appreciated to bundle these.
+  * Fixing a bug in a proto-kernel that doesn't change the function signature nor intended behavior only requires MINOR to be increased.
```
michaelld: Should this be `PATCH` and not `MINOR`? Seems to match the definition below of the former, not the latter.

michaelld (commented):
I like that we're adding specific VERSIONING here, the specific versioning chosen, and the example of what changes to the code would change which part of the version. Wondering if there are any other examples that would be relevant, and also my question about whether `MINOR` is better for the 3rd item.

versioning.md
marcusmueller: Indeed!

Issue #110, opened 2017-02-26
Make error: processor does not support ARM mode `sbfx r11,r1,#2,#1'


MiBatu:
Fail to install on RPi3 (Debian):

Building ASM object lib/CMakeFiles/volk_obj.dir/__/kernels/volk/asm/neon/volk_32f_x2_dot_prod_32f_neonasm_opts.s.o
/home/pi/gnuradio/volk/kernels/volk/asm/neon/volk_32f_x2_dot_prod_32f_neonasm_opts.s: Assembler messages:
/home/pi/gnuradio/volk/kernels/volk/asm/neon/volk_32f_x2_dot_prod_32f_neonasm_opts.s:46: Error: selected processor does not support ARM mode `sbfx r11,r1,#2,#1'
lib/CMakeFiles/volk_obj.dir/build.make:1462: recipe for target 'lib/CMakeFiles/volk_obj.dir/__/kernels/volk/asm/neon/volk_32f_x2_dot_prod_32f_neonasm_opts.s.o' failed
make[2]: *** [lib/CMakeFiles/volk_obj.dir/__/kernels/volk/asm/neon/volk_32f_x2_dot_prod_32f_neonasm_opts.s.o] Error 1
CMakeFiles/Makefile2:195: recipe for target 'lib/CMakeFiles/volk_obj.dir/all' failed
make[1]: *** [lib/CMakeFiles/volk_obj.dir/all] Error 2
Makefile:138: recipe for target 'all' failed
make: *** [all] Error 2

n-west:
See this blog post:

http://lukeberndt.com/2016/compiling-gnuradio-on-a-raspberry-pi/

The raspberry pi OS is built with all kinds of strange flags that make detection hard.

MiBatu:
Thank you, but this link doesn't work =(

"Error establishing a database connection"

MiBatu:
I've pushed 'make' to go forward by 'make -B -i' and there is some other warnings:

_[ 72%] Building C object lib/CMakeFiles/volk_obj.dir/volk_machine_neon_hardfp.c.o
In file included from /home/pi/volk/build/lib/volk_machine_neon_hardfp.c:113:0:
/home/pi/volk/kernels/volk/volk_32fc_s32fc_multiply_32fc.h: In function ‘volk_32fc_s32fc_multiply_32fc_neon’:
[...]

PR #267, opened 2019-08-01
Adds Raspberry Pi 4 support. Removed defunct neon asm kernel.

ast:
Adds toolchain file for Raspberry Pi 4 to be used like this
cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/arm_cortex_a72_hardfp_native.cmake ..

volk_32f_x2_dot_prod_32f_a_neonasm_opts.s contains an instruction 'sbfx' that seems not
to be supported on newer ARM CPUs and has given a lot of people headaches while trying to
build on modern boards. I suggest this file is removed (at least temporarily).


ast:
Added a commit with a neon version of volk_32fc_s32fc_x2_rotator_32fc. Provides x17 performance increase on Raspberry Pi 4.

balister:
A ci test failed, I fond the failing one and pasted here so you can find it faster than I did :)

85: Test command: /bin/sh "/home/travis/build/gnuradio/volk/build/lib/volk_32fc_s32f_magnitude_16i_test.sh" "/home/travis/build/gnuradio/volk/build/lib"
212485: Test timeout computed to be: 9.99988e+06
212585: RUN_VOLK_TESTS: volk_32fc_s32f_magnitude_16i(131071,1)
212685: a_avx2 completed in 0.122796 ms
212785: a_sse3 completed in 0.54835 ms
212885: a_sse completed in 0.510571 ms
212985: generic completed in 0.508554 ms
213085: u_avx2 completed in 0.126189 ms
213185: offset 82176 in1: 309 in2: 310 tolerance was: 0
213285: volk_32fc_s32f_magnitude_16i: fail on arch a_avx2
213385: offset 82176 in1: 309 in2: 310 tolerance was: 0
213485: volk_32fc_s32f_magnitude_16i: fail on arch a_sse3
213585: offset 82176 in1: 309 in2: 310 tolerance was: 0
213685: volk_32fc_s32f_magnitude_16i: fail on arch u_avx2
213785: Best aligned arch: a_avx2
213885: Best unaligned arch: u_avx2
2139 85/130 Test  #85: qa_volk_32fc_s32f_magnitude_16i ....................***Failed    0.02 sec
2140test 86
2141        Start  86: qa_volk_32fc_s32f_power_32fc

balister:
I'm confused why you need to remove a file that should only be built when you are building for an armv7 + neon machine when you are building for an aarch64 machine. The code you are removing should not be included the 64 bit build. Why is the test for the armv7 neon compile succeeding?

ast:
Not sure why qa_volk_32fc_s32f_magnitude_16i fails. It's completly unrelated to my work and my commit doesn't touch that or related files. It's the a_avx2 a_sse3 u_avx2 versions that fails.

Unfortunately Travis doesn't even test the ARM build.

armv8 can be used in armv7 and aarch64 mode. Default Raspbian for Raspberry Pi 3 and 4 currently runs in armv7 mode (32 bit). The kernel in this commit uses NEON compiler intrinsics and will compile to optimized code on both armv7 and aarch64.

I have not been able to compile `volk_32f_x2_dot_prod_32f_a_neonasm_opts.s` even for armv7. I suggest it's at least temporarily removed until a solution can be found as it breaks the neonv7 build. I would love to get more input on this if someone has been able to build it on a modern-ish arm cpu?

`Error: selected processor does not support ARM mode `sbfx r11,r1,#2,#1'`

https://github.com/gnuradio/volk/issues/110

balister:
I'd suggest splitting this pull request into 3 pieces:, the toolchain add, the new kernel, and the remove the kernel.

Regarding the kernel you want to remove, this has worked on everything else arm related, new and old. The problem is most likely the compiler flags you are using. Try adding -mthumb and see what happens.

balister:
And some thumb2 background. (-mthumb should be thumb2)

balister:
From a zynq:

RUN_VOLK_TESTS: volk_32f_x2_dot_prod_32f(131071,1987)
generic completed in 3532.58ms
a_generic completed in 3535.75ms
neonopts completed in 3362.08ms
neon completed in 3363.63ms
a_neonasm completed in 1976.36ms
a_neonasm_opts completed in 3362.3ms
Best aligned arch: a_neonasm
Best unaligned arch: neonopts

ast:
Ok well that settles it!

We just have to find a way so it doesn't break the build on Raspberry Pi 3 and 4!

Could you provide the gcc version and flags you are using?
I'm currently using these:
`-mtune=cortex-a72 -mfpu=neon-fp-armv8 -mfloat-abi=hard`

No luck with -mthumb(2) unfortunately.
 
```
gcc (Raspbian 8.3.0-6+rpi1) 8.3.0

sorry, unimplemented: Thumb-1 hard-float VFP ABI

gcc: error: unrecognized command line option ‘-thumb2’; did you mean ‘-mthumb’?
```

ast:
Philip, exactly which cpu is on your zynq? Trying to track this change down...

ast:
Ok found the problem! `sbfx` is indeed a thumb2 instruction like you said. But the cflags are not copied to the asm flags automatically.

https://github.com/gnuradio/volk/blob/d8f182144fd997d5a265eee97033ffba5d1ad912/lib/CMakeLists.txt#L457

This line:
`set(ARCH_ASM_FLAGS "-mfpu=neon -g")`

Should probably be:
`set(ARCH_ASM_FLAGS "-mthumb -mfpu=neon -g")`

Or maybe `CMAKE_ASM_FLAGS` should be set in the cmake toolchain file instead:
`set(CMAKE_ASM_FLAGS "-march=armv8-a -mthumb -mtune=cortex-a72 -mfpu=neon-fp-armv8 -mfloat-abi=hard -g" CACHE STRING "" FORCE) #same flags for asm sources
`

This is quite significant since the neon build has been broken on Raspberry Pi for at least three years... The `apt` version that ships with Raspbian is not optimized and about 50% slower than neon build.

balister:
Raspian is built to run on cpu's that may not have NEON. So I am not surprised there.

The flags for volk have evolved over the years, and there has been a fair bit of people commiting fixes for their specific situation without realizing the variety of possible cases in the ARM space.

The line that sets flags in volk came should go, since it is making assumptions that might be wrong (although less wrong since NEONV7 flag is only set if the compiler and provided flags can compile and/or assemble NEON. To add a thumb2 flag, we would want to add a compile test for this.

But, You caught on to the proper fix, set the flags properly in the toolchain file. OE is doing this, while trying to recall the history of flags, I remembered this:

https://github.com/openembedded/openembedded-core/commit/43ce4b804d433662fe77c6f5298060ba74a0e639

It does look like the collection of toolchains should be updated.

ast:
It's great that raspbian supports all the older hardware, but the performance penalty is huge in GNURadio...

I will put together a PR with a suggestion for the cmakefile, and toolchain files for rbpi3 and 4.

balister:
And send the kernel in a separate pull request!

ast:
Will do 👍🏼

marcusmueller:
Because I need to release VOLK now, I manually picked (and slightly re-ordered) your code additions in #268. I consider this code reviewed, so I'm merging it, EXCEPT the asm removal, which I'm not clear about whether it entails anything else. It's something we can still easily do for 2.0.1

marcusmueller:
(so, please check whether what you'll find on master shortly is what you wanted. You'll notice neon_intrinsics.h is really just slightly reordered. Please re-submit the asm kernel removal as separate PR!)

ast:
Looks good Marcus. Sorry about being slow, pesky day job taking too much time.

I have some small modifications to the tool chain after my discussion with Philip. We will move asm flags from the cmake file to the toolchain files. Will submit a PR for that at the end of this week.

I also have a lot more kernels that will come as individual PRs one at a time.

marcusmueller:
@ast don't apologize! I'm happy you did this, and I'm sorry that schedule didn't allow me to stall.

By the way, those are awesome prospects!

PR #270, opened 2019-08-06
CMAKE_ASM_FLAGS should be set in the toolchain file

ast:
As per discussion with @balister


michaelld (commented):
How about in the case of "neonv7" if "ARCH_ASM_FLAGS" is not set, then set it as before. That way if someone hasn't picked a specific cmake toolchain file this code might still do something useful.

michaelld:
This is a subset of #273 ... yes? If so, can we close this one & just do the other one?

ast:
Yes, remove this one and merge/review #273, it's more complete.

michaelld:
Closing per agreement.

PR #271, opened 2019-08-06
Adds toolchain file for Raspberry Pi 3 & updated Raspberry 4.

ast:
README.md should probably be updated with an example:

```
# Raspberry Pi 4
cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/arm_cortex_a72_hardfp_native.cmake ..

# Raspberry Pi 3
cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/arm_cortex_a53_hardfp_native.cmake ..
```

### DIFF
--- a/cmake/Toolchains/arm_cortex_a53_hardfp_native.cmake
+++ b/cmake/Toolchains/arm_cortex_a53_hardfp_native.cmake
@@ -4,6 +4,6 @@
 ########################################################################
 set(CMAKE_CXX_COMPILER g++)
 set(CMAKE_C_COMPILER  gcc)
-set(CMAKE_CXX_FLAGS "-march=armv8-a -mtune=cortex-a72 -mfpu=neon-fp-armv8 -mfloat-abi=hard" CACHE STRING "" FORCE)
+set(CMAKE_CXX_FLAGS "-march=armv8-a -mtune=cortex-a53 -mfpu=neon-fp-armv8 -mfloat-abi=hard" CACHE STRING "" FORCE)
 set(CMAKE_C_FLAGS ${CMAKE_CXX_FLAGS} CACHE STRING "" FORCE) #same flags for C sources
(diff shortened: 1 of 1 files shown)


michaelld:
This is great to have available for easy access and use! Is there any reasonable and reliable way to have CMake auto-determine that one of these should be selected, and then which one, such that the user can just do the usual "cmake .."? That would be ideal & awesome!

balister:
@michaelld if there was a way for cmake to read your mind to know what the target hardware was, we wouldn't have to go to all this trouble. The alternative is runtime detection if ISA features. Compile time requires knowledge of the target hardware.

ast:
I can't wait for the first build system that reads my mind 😂

I'd argue that if you are able to use
`cmake ..`
you are also able to use
`cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/arm_cortex_a72_hardfp_native.cmake ..`
it's a power user feature.

But I think it should be **clearly** documented in the README.md build examples.

michaelld:
Ah OK yeah that all makes sense. I was going for native-builds, not cross-builds. Yes: Let's make this a power-user's feature that documented in README.md .

ast:
@michaelld for x86 you can still just type `cmake ..` and be done with it but there are so many ARM cores the detection gets a bit tricky.

Code might build and even execute but it might not be the best version for the platform. The build system would have to build and benchmark. I guess it would be possible but it's just way more complicated.

michaelld:
@ast understood & agreed!

michaelld (approved):
LGTM

michaelld:
@jdemel what do you think about this PR?

jdemel:
Looks good to me too. I have no experience with ARM. @ast and @balister have this experience and approve this PR. So let's merge it.

jdemel (approved):

PR #273, opened 2019-08-07
Toolchain stuff

balister:
ast's work and a commit to fix CI


balister:
This pull request combines 270 and 272. It looks like the additional fix fixes the CI test using a toolchain with some arm thing. Lets merge this and work on any other fallout, like the OE sdk will break, but I am looking at that. I really should check the provided toolchain can assemble thumb2 instructions before enabling the neon asm code.

ast:
Ok Philip! You were faster than me! Depending on the arm target cpu we might have to add
`-mthumb` to `CMAKE_ASM_FLAGS` in  `arm-linux-gnueabihf.cmake toolchain`.

I won't submit a PR for this then but will wait and see what happens to this.

balister:
Regarding -mthumb, there is one v7 assembly kernel that has one thumb instruction. I looked at its performance and it isn't getting used. SO I have a pull request in the wings to remove that kernel. Moral of the story, if you are writing arm asm kernels, don't use thumb instructions.

michaelld (commented):
I think it's wise to move CPU-specific settings into the Toolchains. Is the ASM all that really needs to go there at the moment? If so, I'd say this change is good to go. Wanted to ask first if this is enough.

Issue #216, opened 2018-11-23
Using the Intel Software Development Emulator for CI tests


hcab14:
Continuing the discussion from https://github.com/gnuradio/volk/pull/190#issuecomment-425729635: would it be possible to put the Intel SDE file e.g. on the gnuradio webserver?

For a similar discussion see https://github.com/shibatch/sleef/issues/85

jdemel:
It seems like they could solve their license concerns. @jeffhammond helped them out.
Basically, they use a Travis secure environment variable to hide the URL where they download SDE from. We need to confirm that this approach is the way to go and that it is covered by the SDE license. Then we need to figure out where our URL variable points to. Finally, we need to integrate this into our CI tests. Maybe we use a distinct test for this configuration since it runs way longer than without SDE.

jeffhammond:
There is a better to do it now.  See https://github.com/mmcloughlin/meow/blob/871e50784daff1c9ec5ef66fe3d8fda6bbcd42cf/.travis.yml.  This is not my work.  BLIS uses this method now (https://github.com/flame/blis/blob/master/travis/do_sde.sh).

hcab14:
Is this way of having the Intel emulator acceptable? If yes I'll go ahead and make a pull request.

jeffhammond:
I am not a lawyer and cannot tell you what is acceptable.

michaelld:
Closed via #304

PR #190, opened 2018-07-17
Add avx2 and avx512 support

jessica-iwamoto:
Added AVX, AVX2, FMA, and AVX512-F support to kernels and added compiler checks for AVX512-F and AVX512-CD

PR #278, opened 2019-08-20
Fix AVX FMA rotator

### LINKED ISSUE #2748: Rotator is broken in 3.8
While porting one of my out-of-tree modules to GNU Radio 3.8, I noticed that the Rotator block is broken. The problem can be reproduced with the following flowgraph:

Screenshot from 2019-08-19 08-45-39

This produces the following output:

Screenshot from 2019-08-19 08-46-05

I haven't yet had time to investigate what's going on here.

argilo:
Fixes https://github.com/gnuradio/gnuradio/issues/2748.

This change makes the loops on lines 684-707 and 782-805 match the ones on lines 653-677 and 751-775, after which the rotator correctly rotates the final `(fourthPoints % ROTATOR_RELOAD) * 4`samples.

### DIFF
--- a/kernels/volk/volk_32fc_s32fc_x2_rotator_32fc.h
+++ b/kernels/volk/volk_32fc_s32fc_x2_rotator_32fc.h
@@ -690,7 +690,7 @@ static inline void volk_32fc_s32fc_x2_rotator_32fc_a_avx_fma(lv_32fc_t* outVecto
         yhp = _mm256_movehdup_ps(inc_Val);
 
         tmp1 = aVal;
-        tmp1p = ylp;
+        tmp1p = phase_Val;
 
         aVal = _mm256_shuffle_ps(aVal, aVal, 0xB1);
         phase_Val = _mm256_shuffle_ps(phase_Val, phase_Val, 0xB1);
@@ -788,7 +788,7 @@ static inline void volk_32fc_s32fc_x2_rotator_32fc_u_avx_fma(lv_32fc_t* outVecto
         yhp = _mm256_movehdup_ps(inc_Val);
 
         tmp1 = aVal;
-        tmp1p = ylp;
+        tmp1p = phase_Val;
 
         aVal = _mm256_shuffle_ps(aVal, aVal, 0xB1);
         phase_Val = _mm256_shuffle_ps(phase_Val, phase_Val, 0xB1);
--- a/lib/kernel_tests.h
+++ b/lib/kernel_tests.h
@@ -24,6 +24,13 @@ std::vector<volk_test_case_t> init_test_list(volk_test_params_t test_params)
     volk_test_params_t test_params_inacc = test_params.make_tol(1e-2);
     volk_test_params_t test_params_inacc_tenth = test_params.make_tol(1e-1);
 
+    volk_test_params_t test_params_power(test_params);
+    test_params_power.set_scalar(2.5);
+
+    volk_test_params_t test_params_rotator(test_params);
+    test_params_rotator.set_scalar(std::polar(1.0f, 0.1f));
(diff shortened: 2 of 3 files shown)


michaelld:
I'm in general agreement that this PR's code changes make sense ... but how to test? In Volk itself when I try to test for this kernel I get: ```91: Did not run a test for kernel: volk_32fc_s32fc_x2_rotator_32fc !```

hcab14:
The CI tests use emulation (qemu) to test ARM NEON kernels. For non-ARM kernels, only those kernels are run which use instructions available on the CI instance.

A while ago I have proposed to use also emulation for the non-ARM kernels
https://github.com/gnuradio/volk/pull/190#issuecomment-423807005
https://github.com/gnuradio/volk/issues/216

This might be a good time to restart the discussion.

In a nutshell: 
- qemu does not support newer intel CPUs, therefore one needs to use the [Intel SDE](https://software.intel.com/en-us/articles/pre-release-license-agreement-for-intel-software-development-emulator-accept-end-user-license-agreement-and-download) to run all tests for non-ARM kernels independently of which SIMD instructions are available on the CI instance
 - The Intel SDR executable needs to be stored in some central private place from which it can be downloaded to the CI instance.

argilo:
Running `volk_test_all` on my machine does execute the broken AVX FMA rotator kernels, so I was curious why the tests pass. The reason is that the rotator tests are executed with `phase_inc` set to 327.0. This value comes from here:

https://github.com/gnuradio/volk/blob/1299d72c396a88fd2679adfd7a919ac00d2cf678/lib/testqa.cc#L43

327.0 isn't appropriate for the rotator, since `phase_inc` is supposed to be a value on the complex unit circle. The rotated samples very quickly diverge to infinity, so the tests are effectively useless.

argilo:
Perhaps the tests should fail if any sample (even in the generic kernel) is a non-number.

argilo:
I checked for other tests that end up with non-numbers and `volk_32fc_s32f_power_32fc` also has this problem, presumably because raising something to the 327th power blows up if the input has a magnitude much greater than 1.

michaelld:
LOL yeah that's not a reasonable check.

argilo:
I've added commits to address the deficiencies in the tests. With those in place, the AVX FMA kernels can be verified to fail without my changes, and pass with them.

michaelld:
Excellent! Let us try out these changes to verify on our ends, but I'm glad for us to be fixing these sorts of issues ... so thank you!

argilo:
Is there anything I can do to help move this along? The rotator block in GNU Radio 3.8 is badly broken due to this bug, so it would be good to get the fix merged soon.

jdemel:
Just to make sure how to test this on older machines.

1. Download [SDE](https://software.intel.com/en-us/articles/intel-software-development-emulator)
2. Unpack to location of choice.
3. Run `[SDElocation]/sde -- [volk-build-dir]/lib/volk_test_all`

OR

4. Run `[SDElocation]/sde -- bash` and in your VOLK build dir `ctest`. This option increases test runtime by at least an several orders of magnitude on my machine.

I tried SDE for 5min now and I'm glad it seems to work out of the box.

Anyway, on my machine all tests pass. As far as I can tell all of you came to the same conclusion that the tests pass. I suggest to merge this PR. I'd argue it is beyond the scope of this PR to solve our CI issues.

We need to discuss how we can run our tests and make sure new archs are tested as well.
We have a license issue with SDE for our CI infrastructure. Thus, we can't just pull it in and test all kernels automatically. Though, we can do it manually on our machines.

michaelld (approved):
Works for me & @jdemel ; merging. Sorry this took so long!

PR #280, opened 2019-09-14
cmake: bump minimum version to 3.8

jdemel:
This commit bumps the minimum required CMake version to 3.8. This is in
line with GNU Radio 3.8 requirements. Ubuntu 18.04 provides CMake 3.10.
Other distros are probably ahead of Ubuntu LTS.
Alternatively, we could bump the minimum version to 3.5.1 because this
is the version Ubuntu 16.04 ships. Though, I'd argue that `master`
should require newer versions. This would also be in line with hopefully
moving to C++17 in the near future.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@
 ########################################################################
 # Project setup
 ########################################################################
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.8)
 set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "Choose build type: None Debug Release RelWithDebInfo MinSizeRel")
 project(volk)
 
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -208,14 +208,6 @@ if(CPU_IS_x86)
     # Disable SSE4a if Clang is less than version 3.2
     if("${CMAKE_C_COMPILER_ID}" MATCHES "Clang")
       # Figure out the version of Clang
-      if(CMAKE_VERSION VERSION_LESS "2.8.10")
-        # Extract the Clang version from the --version string.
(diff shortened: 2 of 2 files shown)


michaelld (commented):
I have no objection to moving to CMake 3.8 minimum. That said: Beyond "that's what GR is doing", is any of the CMake scripting in Volk requiring anything newer than 2.8.12? Also, wondering we can think of any CMake programming benefits that we could embrace by bumping the minimum to 3.8?

jdemel:
I think we don't really check for older CMake versions in our CI. Or at least at some point we drop these checks. support. I'd argue we should require minimum versions for all our tools that we also test. Otherwise issues pop up at some point that things break. Travis tells us that the CI tests for Ubuntu 14.04 fail. That's an example for a system we will probably drop soon because it is no longer supported.
If we bump the minimum version, new PRs can actually rely on this minimum version and use it. @noc0lour  bumped the minimum CMake version in GR because it enabled more modern CMake. That would be a benefit we could embrace with this PR.
A quick `grep -irn cmake_version .` in the VOLK root folder shows several places that check for a minimum CMake version. I hope we can move to more comprehensible CMake code with a newer CMake version. 
The results of `grep -irn cmake_version .` should be fixed though because the all check for CMake version < 2.8.12 anyways.

jdemel:
One good reason to bump the minimum version is PR #255 . At the moment it includes some If CMake Version < 3.8 stuff. This would be obsolete.

noc0lour:
Anything below 3 doesn't make sense anyway and >3 enables modern CMake features as well.

michaelld:
OK I'm on board this bump. @jdemel can you put together a 2nd commit that addresses the results of `grep -irn cmake_version .` ?

jdemel:
@michaelld Yes. I'll do so.

michaelld (approved):
LGTM; thx!

PR #284, opened 2019-09-18
VOLK C++ allocator and std::vector type alias added

hcab14:
This adds a C++ allocator for VOLK (`volk::alloc`) and a `std::vector` alias, called `volk::vector` that uses `volk_malloc` and `volk_free`. It might be useful for GNURadio and other projects using VOLK and C++.


jdemel:
Could you elaborate on the use case of this PR? e.g. a simple example how to use it?

hcab14:
Instead of doing this (see _e.g._ https://github.com/gnuradio/gnuradio/blob/master/gr-qtgui/lib/waterfall_sink_f_impl.cc)
```
gr_complex *z = (gr_complex*)volk_malloc(n * sizeof(gr_complex));
// initiallize
memset(z, 0, n*sizeof(gr_complex));
// resize
volk_free(z);
z = (gr_complex*)volk_malloc(m * sizeof(gr_complex));
// initiallize
memset(z, 0, m*sizeof(gr_complex));
// deallocate
volk_free(z);
```

one can write simply
```
volk::vector<gr_complex> z(n,0);
z.resize(m);
std::fill_n(z.begin(), n, 0);
```
_i.e._ the usual benefits of using objects as opposed to pointers (RAII) + the type information is manifest (no void* pointers). 

For gnuradio the command `grep -r volk_free . | grep .cc: | wc -l` finds 235 occurrences of `volk_free`.

jdemel:
This seems like a really nice feature for C++. And I fully agree that we should follow RAII.
So, this would be the way to go here?
```
int n = 5;
volk::vector<short> out(n, 0);
volk::vector<char> in(n, 20);
volk_8i_convert_16i(out.data(), in.data(), in.size());
```

hcab14:
Yes.

include/volk/volk_alloc.h
```diff
@@ -0,0 +1,80 @@
...
+      return p;
+
+    throw std::bad_alloc();
+  }
+
+  void deallocate(T* p, std::size_t) noexcept { volk_free(p); }
+
+} ;
+
+template <class T, class U>
```
michaelld: What do these operators actually do?

michaelld (commented):
I like this addition ... allows Volk usage in a more c++ manner. I'm wondering if we need to somehow designate this file as C++ code, via some obvious header or the like -- for example using `.hh` or `.hpp`. I believe that some of the headers are C-only, so I'd hope the naming would make it clear that this one is not C.

include/volk/volk_alloc.h
hcab14: This code is taken from https://en.cppreference.com/w/cpp/named_req/Allocator and adapted to use `volk_alloc`/`volk_free` instead of `malloc`, `free`. As for `operator==` and `operator!=`, see the explanation in https://stackoverflow.com/questions/47186641/why-c-custom-allocator-needs-comparison-operators

michaelld: About `operator==` ... wow. That's clear as mud. I guess it's required, and that's the syntax ... so that's just the way it is! Do you have a better explanation? Just curious now ...

hcab14: From  https://en.cppreference.com/w/cpp/named_req/Allocator:
Screenshot from 2019-10-28 09-19-01

So the way `operator==` is defined is saying that any object allocated with `volk_malloc` can be deallocated with `volk_free`.

michaelld: Ah OK now -that- makes sense! And so `operator!=` means if `volk_malloc` didn't do the allocation then `volk_free` can't deallocate it? Or something like that. Makes sense too.

hcab14: > Ah OK now -that- makes sense! And so `operator!=` means if `volk_malloc` didn't do the allocation then `volk_free` can't deallocate it? Or something like that. Makes sense too.

No, `operator!=` is just the logical negation of `operator==`.

Here (https://en.cppreference.com/w/cpp/memory/allocator/operator_cmp) a condition is given for when `operator==`, `operator!=` always return `true` and `false`: this is the case for **stateless** allocators, which I think is true for `volk_malloc`, `volk_free`, and for  `volk_get_alignment`.

hcab14:
> I like this addition ... allows Volk usage in a more c++ manner. I'm wondering if we need to somehow designate this file as C++ code, via some obvious header or the like -- for example using `.hh` or `.hpp`. I believe that some of the headers are C-only, so I'd hope the naming would make it clear that this one is not C.

I see your point. Let me know which extension you prefer.

michaelld:
I'm good with either. I use both in my projects, though I stick with the same within a given project.

@hcab14 @jdemel Do either of you have a preference?

jdemel:
I don't have a preference. Do we already have other files with that extension?

michaelld:
A quick review of the Volk files shows that all of the `.h` headers and all of the '.c` files are C. all of the `.cc` files are C++. So I'd propose using `.hh` to match the `.cc`. Simple and inline with the current naming IMHO.

michaelld:
or ... not quite: `volk_complex.h` is sort of a mix: if C++ then it provides C++ code, and if C then it provides C code.

jdemel:
How do we proceed here? My takeaway from this discussion is that we favor `.hh` because it is in line with `.cc`.
Since `volk_complex.h` is a mix, we should use it as a special case and ignore it in our discussion.

hcab14:
> How do we proceed here? My takeaway from this discussion is that we favor `.hh` because it is in line with `.cc`.
> Since `volk_complex.h` is a mix, we should use it as a special case and ignore it in our discussion.

I have renamed `volk_alloc.h` to `volk_alloc.hh`

michaelld (approved):
LGTM

michaelld:
How to proceed? I'm good with the current state of things, and am ready to merge. How about you @jdemel ?

jdemel:
Yeah. In this case let's merge this PR. Looks good to me.

PR #285, opened 2019-09-18
fix for volk_32fc_s32f_magnitude_16i (different results of SIMD kernels w.r.t. generic kernel)

hcab14:
This is an attempt to fix `volk_32fc_s32f_magnitude_16i` which is broken, as can be seen by running the following command
```
while [ 1 ]; do volk_profile -i 1 -n -v 13100071 -R volk_32fc_s32f_magnitude_16i ; done |grep fail
```

Unfortunately I do not completely understand why the proposed changes seem to work. Adding the `volatile` keywords seems to prevent the compiler from making optimizations which break the agreement between the generic implementation and the SIMD implementations.

### DIFF
--- a/kernels/volk/volk_32fc_s32f_magnitude_16i.h
+++ b/kernels/volk/volk_32fc_s32f_magnitude_16i.h
@@ -124,9 +124,11 @@ volk_32fc_s32f_magnitude_16i_a_avx2(int16_t* magnitudeVector, const lv_32fc_t* c
   number = eighthPoints * 8;
   magnitudeVectorPtr = &magnitudeVector[number];
   for(; number < num_points; number++){
-    float val1Real = *complexVectorPtr++;
-    float val1Imag = *complexVectorPtr++;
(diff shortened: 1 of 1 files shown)


jdemel:
This patch changes the code that handles the "end of vector" stub. This is just the generic kernel code. Instead of copypasta, would it be possible to just call the generic kernel on that stub? That would eliminate some copypasta and be more verbose. Also, it would hopefully prevent future regressions if one changes these lines again.

hcab14:
Are there any comments/suggestions?

jdemel (approved):
Looks good to me now.

michaelld (approved):
LGTM

hcab14:
I wonder why this problem is not seen by CI.

Unfortunately I cannot test this locally but you may try the code in PR https://github.com/gnuradio/volk/pull/297


> Dropping the `volatile` is fine.

Just dropping the volatile breaks the CI tests.

hcab14:
@gvanem: this seems to be a bug in the Intel C++ compiler
https://software.intel.com/en-us/forums/intel-c-compiler/topic/542377

PR #286, opened 2019-09-18
fix for flaky kernel volk_32f_log2_32f

hcab14:
This "fix" consists of a modification of the generic kernel
-  when the result of `log2f(x)` is `-inf` the return value is set fo `-127.0f`
-  the opposite situation, _i.e._, `log2f(x)` = `+inf` is not tested since floats in the range [-1.0f 1.0f] are generated for the tests.

I have put "fix" in quotation marks, since one could also require the SIMD kernels to return `+-inf` whenever appropriate but this would incur a cost on the speed.

Any comments/suggestions are more than welcome.

### DIFF
--- a/kernels/volk/volk_32f_log2_32f.h
+++ b/kernels/volk/volk_32f_log2_32f.h
@@ -27,6 +27,9 @@
  *
  * Computes base 2 log of input vector and stores results in output vector.
  *
+ * Note that this implementation is not conforming to the IEEE FP standard, i.e.,
+ * +-Inf outputs are mapped to +-127.0f and +-NaN input values are not supported.
+ *
  * This kernel was adapted from Jose Fonseca's Fast SSE2 log implementation
  * http://jrfonseca.blogspot.in/2008/09/fast-sse2-pow-tables-or-polynomials.htm
  *
@@ -96,6 +99,12 @@
 
 #define LOG_POLY_DEGREE 6
 
+// +-Inf -> +-127.0f in order to match the behaviour of the SIMD kernels
+static inline float log2f_non_ieee(float f) {
+  float const result = log2f(f);
+  return isinf(result) ? copysignf(127.0f, result) : result;
+}
+
 #ifdef LV_HAVE_GENERIC
 
 static inline void
@@ -105,11 +114,8 @@ volk_32f_log2_32f_generic(float* bVector, const float* aVector, unsigned int num
   const float* aPtr = aVector;
   unsigned int number = 0;
 
-  for(number = 0; number < num_points; number++) {
-    float const result = log2f(*aPtr++);
-    *bPtr++ = isinf(result) ? -127.0f : result;
-  }
-
+  for(number = 0; number < num_points; number++)
+    *bPtr++ = log2f_non_ieee(*aPtr++);
 }
 #endif /* LV_HAVE_GENERIC */
 
--- a/kernels/volk/volk_32f_log2_32f.h
+++ b/kernels/volk/volk_32f_log2_32f.h
@@ -150,7 +150,8 @@ volk_32f_log2_32f_a_avx2_fma(float* bVector, const float* aVector, unsigned int 
 
   number = eighthPoints * 8;
   for(;number < num_points; number++){
-    *bPtr++ = log2f(*aPtr++);
+    float const result = log2f(*aPtr++);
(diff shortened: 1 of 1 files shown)


hcab14:
.. the CI failures are all due to `volk_32fc_s32f_magnitude_16i`, which should be fixed in https://github.com/gnuradio/volk/pull/285 .

jdemel:
Just like my comment in #285, would it be possible to just call the generic kernel on that stub? These are a lot of copypasta changes.

I use the [log2f documentation](https://en.cppreference.com/w/c/numeric/math/log2).
Since it doesn't make sense to use this kernel in the range (-inf, 0], I'd argue the best fix would be to adjust the kernel test data to only use input values in the range (0, inf). 

This seems like we need to adjust our test code here. Specifically the code in in `qa_utils.*`. We need to set the min/max values in `load_random_data` according to the needs of the kernel. This does probably entail a change to `volk_type_t` and `volk_test_params_t`. 

Thanks for working on this.

hcab14:
> Just like my comment in #285, would it be possible to just call the generic kernel on that stub? These are a lot of copypasta changes.

Done.

> I use the [log2f documentation](https://en.cppreference.com/w/c/numeric/math/log2).
> Since it doesn't make sense to use this kernel in the range (-inf, 0], I'd argue the best fix would be to adjust the kernel test data to only use input values in the range (0, inf).
> 
> This seems like we need to adjust our test code here. Specifically the code in in `qa_utils.*`. We need to set the min/max values in `load_random_data` according to the needs of the kernel. This does probably entail a change to `volk_type_t` and `volk_test_params_t`.
> 

The question is if we want volk_32f_log2_32f to behave exactly as log2f in which case we have to implement the correct behaviour for negative inputs, -Inf, +Inf, -NaN, +NaN, and  IEEE FP arithmetic. This can easily make the SIMD kernels slower by a factor of two or more.

jdemel:
> > I use the [log2f documentation](https://en.cppreference.com/w/c/numeric/math/log2).
> > Since it doesn't make sense to use this kernel in the range (-inf, 0], I'd argue the best fix would be to adjust the kernel test data to only use input values in the range (0, inf).
> > This seems like we need to adjust our test code here. Specifically the code in in `qa_utils.*`. We need to set the min/max values in `load_random_data` according to the needs of the kernel. This does probably entail a change to `volk_type_t` and `volk_test_params_t`.
> 
> The question is if we want volk_32f_log2_32f to behave exactly as log2f in which case we have to implement the correct behaviour for negative inputs, -Inf, +Inf, -NaN, +NaN, and IEEE FP arithmetic. This can easily make the SIMD kernels slower by a factor of two or more.

I agree, we do not need to aim for full compliance here. I suggest to restrict the test input data to the range (0, inf).

kernels/volk/volk_32f_log2_32f.h
```diff
@@ -403,7 +397,8 @@ volk_32f_log2_32f_u_generic(float* bVector, const float* aVector, unsigned int n
   unsigned int number = 0;
 
   for(number = 0; number < num_points; number++){
-    *bPtr++ = log2f(*aPtr++);
+    float const result = log2f(*aPtr++);
```
michaelld, line 406: I'd advise to create a macro that does these 2 lines, then the macro can be called from both here and the `generic` function. That way any change to the macro affects everything; just have to make 1 change for all kernels to benefit!

michaelld (changes requested):
Whatever we decide with regard to the `inf` question, let's add it to the documentation in the header. With that & making a macro or inline function or the like from the core work code, I'm good with this change.

kernels/volk/volk_32f_log2_32f.h
hcab14: All SIMD kernels call the generic kernel, but anyway it is good to make it clear that this implementation is not IEEE FP conform.

hcab14:
.. the CI failure is not for this kernel

kernels/volk/volk_32f_log2_32f.h
michaelld: Looks good!

michaelld (approved):
LGTM; thx!

michaelld:
@jdemel I'm good with this one. What sayeth thou?

jdemel (approved):
Looks good to me too.

Issue #229, opened 2018-12-29
qa_volk_8u_x3_encodepolarpuppet_8u: heap buffer overflow detected by using ASAN


noc0lour:
```
129/360 Test #129: qa_volk_8u_x3_encodepolarpuppet_8u .................***Failed    0.08 sec
RUN_VOLK_TESTS: volk_8u_x3_encodepolarpuppet_8u(131071,1)
generic completed in 9.164ms
=================================================================
==10469==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x631000038800 at pc 0x7faa090c752f bp 0x7ffc1bcc6210 sp 0x7ffc1bcc6208
READ of size 16 at 0x631000038800 thread T0
    #0 0x7faa090c752e in volk_8u_x2_encodeframepolar_8u_u_ssse3 /home/andrej/git/gnuradio/gnuradio/volk/kernels/volk/volk_8u_x2_encodeframepolar_8u.h:178:15
    #1 0x7faa090c8af8 in volk_8u_x3_encodepolar_8u_x2_u_ssse3 /home/andrej/git/gnuradio/gnuradio/volk/kernels/volk/volk_8u_x3_encodepolar_8u_x2.h:116:3
    #2 0x7faa090c8c51 in volk_8u_x3_encodepolarpuppet_8u_u_ssse3 /home/andrej/git/gnuradio/gnuradio/volk/kernels/volk/volk_8u_x3_encodepolarpuppet_8u.h:81:3
    #3 0x7faa08c39474 in volk_8u_x3_encodepolarpuppet_8u_manual /home/andrej/git/gnuradio/gnuradio/build/volk/lib/volk.c:10875:5
    #4 0x445043 in run_cast_test4(void (*)(void*, void*, void*, void*, unsigned int, char const*), std::vector<void*, std::allocator<void*> >&, unsigned int, unsigned int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) /home/andrej/git/gnuradio/gnuradio/volk/lib/qa_utils.cc:222:19
[...]

PR #288, opened 2019-09-22
fix for spurious heap buffer overflow

hcab14:
in the last iteration of the loop an element outside the allocated region is loaded which is then never being used -> use `__VOLK_PREFETCH`

performance is about the same as before

https://github.com/gnuradio/volk/issues/229

### DIFF
--- a/kernels/volk/volk_8u_x2_encodeframepolar_8u.h
+++ b/kernels/volk/volk_8u_x2_encodeframepolar_8u.h
@@ -156,27 +156,27 @@ volk_8u_x2_encodeframepolar_8u_u_ssse3(unsigned char* frame, unsigned char* temp
   frame_ptr = frame;
   temp_ptr = temp;
 
-  // load first chunk.
-  // Tests show a 1-2% gain compared to loading a new chunk and using it right after.
-  r_temp0 = _mm_loadu_si128((__m128i*) temp_ptr);
-  temp_ptr += 16;
+  // prefetch first chunk
+  __VOLK_PREFETCH(temp_ptr);
 
   const __m128i shuffle_stage4 = _mm_setr_epi8(0, 8, 4, 12, 2, 10, 6, 14, 1, 9, 5, 13, 3, 11, 7, 15);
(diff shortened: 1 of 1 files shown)


michaelld (changes requested):
I see the issue & agree that what you propose is a good solution. Thx!

Requests:

(1) Please move the comment "// start loading next chunk." along with the other lines; means adding it back in where you've already moved the other lines of code.

(2) Please put an empty line after the `__VOLK_PREFETCH` line inside the loops, which follows the current code's pattern of blankspace.

(3)The implementations for `volk_8u_x2_encodeframepolar_8u_a_avx2` in this same file contains the same error. Please fix it too (same same fix) & then we can finalize testing & get this committed!

hcab14:
Thank you for your review

> (1) Please move the comment "// start loading next chunk." along with the other lines; means adding it back in where you've already moved the other lines of code.
> (2) Please put an empty line after the `__VOLK_PREFETCH` line inside the loops, which follows the current code's pattern of blankspace.

Please have a look and see if it is now ok

> (3)The implementations for `volk_8u_x2_encodeframepolar_8u_a_avx2` in this same file contains the same error. Please fix it too (same same fix) & then we can finalize testing & get this committed!

While looking at the code I realized that the SSE3 code can be shortened using the same technique: it contained an explicit last iteration at the end which now is inside the for loop.

michaelld (approved):
LGTM & CI passes. The failure is for a different kernel/s.

michaelld:
I like the way you rolled the final iteration back into the main loop, which as you note shortens the code very nicely!

michaelld:
@jdemel Do you wish to review here?

jdemel:
I like this PR too. After the changes @michaelld requested the code looks fine.

hcab14:
> I like this PR too. After the changes @michaelld requested the code looks fine.

Let me know if further changes are required or if I missed to implement any of the requested changes.

michaelld (approved):
LGTM; thanks!

PR #289, opened 2019-09-30
bug fixes for volk_32f_index_max_{16,32}u

### LINKED ISSUE #169: volk_32f_index_max_32u fails tests on Win64/MSVC 2015
RUN_VOLK_TESTS: volk_32f_index_max_32u(131071,1)
a_sse4_1 completed in 0ms
u_sse4_1 completed in 1ms
a_sse completed in 0ms
u_sse completed in 0ms
a_avx completed in 0ms
u_avx completed in 0ms
generic completed in 0ms
offset 0 in1: 38723 in2: 93108 tolerance was: 0
volk_32f_index_max_32u: fail on arch a_sse4_1
offset 0 in1: 38723 in2: 93108 tolerance was: 0
volk_32f_index_max_32u: fail on arch u_sse4_1
offset 0 in1: 38723 in2: 93108 tolerance was: 0
volk_32f_index_max_32u: fail on arch a_sse
offset 0 in1: 38723 in2: 93108 tolerance was: 0
volk_32f_index_max_32u: fail on arch u_sse
offset 0 in1: 38723 in2: 70001 tolerance was: 0
volk_32f_index_max_32u: fail on arch a_avx
offset 0 in1: 38723 in2: 70001 tolerance was: 0
volk_32f_index_max_32u: fail on arch u_avx
Best aligned arch: a_sse4_1
Best unaligned arch: u_sse

hcab14:
There were two problems with these kernels, both occurring only when input vector contains more than one identical number
1. the computation of the mask used in the SIMD kernels was wrong: it selected `<=` instead of `<`
2. when extracting the maximum and the maximum index in the SIMD kernels it was not guaranteed that the first maximum was selected

This should fix https://github.com/gnuradio/volk/issues/169 and CI test failures.

### DIFF
--- a/kernels/volk/volk_32f_index_max_16u.h
+++ b/kernels/volk/volk_32f_index_max_16u.h
@@ -109,7 +109,7 @@ volk_32f_index_max_16u_a_avx(uint16_t* target, const float* src0,
     currentValues  = _mm256_load_ps(inputPtr); inputPtr += 8;
     currentIndexes = _mm256_add_ps(currentIndexes, indexIncrementValues);
 
-    compareResults = _mm256_cmp_ps(maxValues, currentValues,14);
+    compareResults = _mm256_cmp_ps(currentValues, maxValues, _CMP_LE_OS);
 
     maxValuesIndex = _mm256_blendv_ps(currentIndexes, maxValuesIndex, compareResults);
     maxValues      = _mm256_blendv_ps(currentValues, maxValues, compareResults);
@@ -123,6 +123,9 @@ volk_32f_index_max_16u_a_avx(uint16_t* target, const float* src0,
     if(maxValuesBuffer[number] > max){
       index = maxIndexesBuffer[number];
       max = maxValuesBuffer[number];
+    } else if(maxValuesBuffer[number] == max){
+      if (index > maxIndexesBuffer[number])
+        index = maxIndexesBuffer[number];
     }
   }
 
@@ -170,7 +173,7 @@ volk_32f_index_max_16u_a_sse4_1(uint16_t* target, const float* src0,
     currentValues  = _mm_load_ps(inputPtr); inputPtr += 4;
     currentIndexes = _mm_add_ps(currentIndexes, indexIncrementValues);
 
-    compareResults = _mm_cmpgt_ps(maxValues, currentValues);
+    compareResults = _mm_cmple_ps(currentValues, maxValues);
 
     maxValuesIndex = _mm_blendv_ps(currentIndexes, maxValuesIndex, compareResults);
     maxValues      = _mm_blendv_ps(currentValues, maxValues, compareResults);
@@ -184,6 +187,9 @@ volk_32f_index_max_16u_a_sse4_1(uint16_t* target, const float* src0,
     if(maxValuesBuffer[number] > max){
       index = maxIndexesBuffer[number];
       max = maxValuesBuffer[number];
+    } else if(maxValuesBuffer[number] == max){
+      if (index > maxIndexesBuffer[number])
+        index = maxIndexesBuffer[number];
     }
   }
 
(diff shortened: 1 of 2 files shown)


kernels/volk/volk_32f_index_max_16u.h
```diff
@@ -109,7 +109,7 @@ volk_32f_index_max_16u_a_avx(uint16_t* target, const float* src0,
     currentValues  = _mm256_load_ps(inputPtr); inputPtr += 8;
     currentIndexes = _mm256_add_ps(currentIndexes, indexIncrementValues);
 
-    compareResults = _mm256_cmp_ps(maxValues, currentValues,14);
+    compareResults = _mm256_cmp_ps(currentValues, maxValues, _CMP_LE_OS);
```
michaelld: A couple queries here, noting that the `14` from the original code corresponds to `_CMP_GT_OS`:

(1) I believe that using `LE` (or `GE` if the first 2 arguments are swapped) just changes the index returned of the max value to being the -final- index versus the -first- index if using `LT` (or `GT`). Yes? We should pick which one we want to go with and document it in the header.

(2) Do you know the difference between the `_CMP_*_OS` (ordered, signaling) and `_CMP_*_OQ` (ordered, non-signaling)? I'm not sure it makes a difference here, but maybe it does?

michaelld (commented):
I'm not convinced that the original kernel code is buggy. Reading through it, the flow makes sense to me. Your version also makes sense, though it has to work a little harder to do so because (I think) using `LE` versus `GT` means the final index selections also has to be `LE`. I don't see this kernel breaking CI, nor does it fail for me in a local build.

I'm still open to arguments as to what this change truly does and fixes. See my queries in the other comment.

michaelld:
Minimally, I'd like to see 2 changes to the current kernel: (1) 14 -> `_CMP_GT_OS`; and (2) We should pick which index (first or last) we want to go with and document it in the header.

kernels/volk/volk_32f_index_max_16u.h
hcab14: > A couple queries here, noting that the `14` from the original code corresponds to `_CMP_GT_OS`:
> 
> (1) I believe that using `LE` (or `GE` if the first 2 arguments are swapped) just changes the index returned of the max value to being the -final- index versus the -first- index if using `LT` (or `GT`). Yes? 

No: `LE` gives the **first** index and `LT` the **final** index:
```
mask = _mm256_cmp_ps(currentValues, maxValues, _CMP_LE_OS); 
// mask = currentValues <= maxValues ? 0xFFFFFFFF : 0                                                                                
maxValuesIndex  = _mm256_blendv_ps(currentIndexes, maxValuesIndex , mask); 
// maxValuesIndex  = mask ? maxValuesIndex  : currentIndexes
```
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=_mm256_cmp_ps
https://software.intel.com/sites/landingpage/IntrinsicsGuide/#text=blendv_ps

Do you agree?

I have matched the SIMD versions to the generic version.

> We should pick which one we want to go with and document it in the header.
> 
I agree

> (2) Do you know the difference between the `_CMP_*_OS` (ordered, signaling) and `_CMP_*_OQ` (ordered, non-signaling)? I'm not sure it makes a difference here, but maybe it does?

According to https://stackoverflow.com/questions/16988199/how-to-choose-avx-compare-predicate-variants, non-signaling does not raise a FP exception is an operand contains NaN while signaling does.

hcab14:
> Minimally, I'd like to see 2 changes to the current kernel: (1) 14 -> `_CMP_GT_OS`; 

Done with corresponding changes which ensure that the first maximum index is found

> and (2) We should pick which index (first or last) we want to go with and document it in the header.

Let me know if the changes are fine.

jdemel (approved):
This PR looks good to me.

michaelld (approved):
Thanks for the changes. LGTM.

Issue #266, opened 2019-07-31
volk_8u_conv_k7_r2puppet_8u : flaky kernel on OSX 10.14


michaelld:
Here's the output from `volk_profile`
```
RUN_VOLK_TESTS: volk_8u_conv_k7_r2puppet_8u(131071,1987)
spiral completed in 1008.89 ms
generic completed in 32251.3 ms
offset 0 in1: 1 in2: 0 tolerance was: 0
offset 1 in1: 1 in2: 0 tolerance was: 0
offset 2 in1: 1 in2: 0 tolerance was: 0
offset 4 in1: 1 in2: 0 tolerance was: 0
offset 5 in1: 1 in2: 0 tolerance was: 0
offset 6 in1: 1 in2: 0 tolerance was: 0
offset 8 in1: 1 in2: 0 tolerance was: 0
volk_8u_conv_k7_r2puppet_8u: fail on arch spiral
Best aligned arch: generic
Best unaligned arch: generic
```

jdemel:
Is this issue related to #117 ?

hcab14:
AFAIC this issue can be closed .

michaelld:
Yes guessing so. I'll reopen if when I update Volk I still see this issue.

Issue #117, opened 2017-08-05
volk_8u_conv_k7_r2puppet_8u fails in Ubuntu 16.04


dmiralles2009:
I cloned the volk repo and perform the installation steps as described in the README file with the caveat that I am not running the last step of `$ sudo make install` since I want a local copy only. I am, however, running the `volk_profile` executable from within the build directory as `./apps/volk_profile`. While checking the output of the code I noticed the following:

RUN_VOLK_TESTS:volk_8u_conv_k7_r2puppet_8u(131071,198)
spiral completed in 191.857ms
generic completed in 2393.31ms
offset 0 in1: 1 in2: 0 tolerance was: 0
offset 1 in1: 1 in2: 0 tolerance was: 0
offset 2 in1: 1 in2: 0 tolerance was: 0
offset 3 in1: 1 in2: 0 tolerance was: 0
offset 4 in1: 1 in2: 0 tolerance was: 0
offset 5 in1: 1 in2: 0 tolerance was: 0
offset 6 in1: 1 in2: 0 tolerance was: 0
offset 7 in1: 1 in2: 0 tolerance was: 0
offset 8 in1: 1 in2: 0 tolerance was: 0
offset 9 in1: 1 in2: 0 tolerance was: 0
volk_8u_conv_k7_r2puppet_8u: fail on arch spiral
Best aligned arch: generic
Best unaligned arch: generic

Notice that the proto kernel fails and there are some print messages regarding the tolerance, not really sure I fully understand it. For additional information, I also included the architecture of the machine in question:

Architecture:          x86_64
CPU op-mode(s):        32-bit, 64-bit
Byte Order:            Little Endian
CPU(s):                8
On-line CPU(s) list:   0-7
Thread(s) per core:    2
Core(s) per socket:    4
Socket(s):             1
NUMA node(s):          1
Vendor ID:             GenuineIntel
CPU family:            6
Model:                 58
Model name:            Intel(R) Core(TM) i7-3632QM CPU @ 2.20GHz
Stepping:              9
CPU MHz:               1217.304
CPU max MHz:           3200.0000
CPU min MHz:           1200.0000
BogoMIPS:              4389.84
Virtualization:        VT-x
L1d cache:             32K
L1i cache:             32K
L2 cache:              256K
L3 cache:              6144K
NUMA node0 CPU(s):     0-7

Thanks in advance for the help

n-west:
[...]

PR #294, opened 2019-10-27
volk_8u_conv_k7_r2puppet_8u: initialize decisions to zero

hcab14:
The array of decisions it not initialized. When running volk_profile with this kernel alone I did not observe any failures, but when running volk_profile for all kernels, `volk_8u_conv_k7_r2puppet_8u` fails quite consistently. In the latter case it is more likely that memory is reused and its contents are not guaranteed to be zero.

@jessica-iwamoto: is it correct to initialize it to zero? (I guess yes, because these are updated using `|=`)

This PR is supposed to fix the following issues: 
https://github.com/gnuradio/volk/issues/266
https://github.com/gnuradio/volk/issues/117

### DIFF
--- a/kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
+++ b/kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
@@ -155,8 +155,11 @@ static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms, unsig
     once = 0;
   }
 
-    //unbias the old_metrics
+  //unbias the old_metrics
   memset(X, 31, d_numstates);
+
+  // initialize decisions
+  memset(D, 0, (d_numstates/8) * (framebits + 6));
 
   volk_8u_x4_conv_k7_r2_8u_spiral(Y, X, syms, D, framebits/2 - excess, excess, Branchtab);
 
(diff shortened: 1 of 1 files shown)


kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
```diff
@@ -155,9 +155,12 @@ static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms, unsig
     once = 0;
   }
 
-    //unbias the old_metrics
+  //unbias the old_metrics
   memset(X, 31, d_numstates);
 
+  // initialize decisions
+  memset(D, 0, (d_numstates/8) * (framebits + 6));
```
michaelld, line 162: I think it makes sense to 0 these. `volk_alloc` doesn't guarantee that the allocated memory is 0'd, and since this variable isn't set to anything otherwise, 0 seems wise!

michaelld (approved):
LGTM

michaelld:
CI failure has nothing to do with this PR.

hcab14:
> CI failure has nothing to do with this PR.

Yes... Note that the 2nd commit is for a different kernel (also an issue with missing initialization) which _seems_ to fix it.

michaelld:
OK good to know. That code reads OK to me too & seems to help.

michaelld:
Waiting for @jdemel 's input.

kernels/volk/volk_32f_8u_polarbutterflypuppet_32f.h
```diff
@@ -96,6 +96,7 @@ volk_32f_8u_polarbutterflypuppet_32f_generic(float* llrs, const float* input, un
   unsigned int frame_exp = log2_of_power_of_2(frame_size);
 
   sanitize_bytes(u, elements);
+  clean_up_intermediate_values(llrs, u, frame_size, elements);
```
jdemel: That's a good addition to make things more reliable. I'm OK to fix this in a PR that deals with another issue. But we should try to have one PR per issue. Even if it is just for verbosity.

hcab14: Done. See https://github.com/gnuradio/volk/pull/295

jdemel (approved):
Looks good to me.

PR #296, opened 2019-10-29
Fix `volk_32fc_convert_16ic_neon`

hcab14:
The problem was that `rintf` uses the current rounding mode while the NEON implementation had hard-coded `FE_DOWNWARD`.

Unfortunately 
 - the new neon implementation is ~1.66 times slower than the original one but still ~5 times faster than the generic implementation on an RPi3
 - for the NEONv8 implementation the inline assembly code used for NEONv7 does not work (compile error)
 - there is quite a lot of copy-and-paste between then v7 and v8 NEON implementations

If requested, I can split this PR into two PRs.

### DIFF
--- a/kernels/volk/volk_32fc_convert_16ic.h
+++ b/kernels/volk/volk_32fc_convert_16ic.h
@@ -159,6 +159,12 @@ static inline void volk_32fc_convert_16ic_a_sse2(lv_16sc_t* outputVector, const 
 #ifdef LV_HAVE_NEON
 #include <arm_neon.h>
 
+#define VCVTRQ_S32_F32(result, value)					\
+  __VOLK_ASM("VCVTR.S32.F32 %0, %1" : "=t"(result[0]) : "t"(value[0]) :); \
+  __VOLK_ASM("VCVTR.S32.F32 %0, %1" : "=t"(result[1]) : "t"(value[1]) :); \
+  __VOLK_ASM("VCVTR.S32.F32 %0, %1" : "=t"(result[2]) : "t"(value[2]) :); \
+  __VOLK_ASM("VCVTR.S32.F32 %0, %1" : "=t"(result[3]) : "t"(value[3]) :);
+
 static inline void volk_32fc_convert_16ic_neon(lv_16sc_t* outputVector, const lv_32fc_t* inputVector, unsigned int num_points)
 {
     const unsigned int neon_iters = num_points / 4;
@@ -173,8 +179,7 @@ static inline void volk_32fc_convert_16ic_neon(lv_16sc_t* outputVector, const lv
 
     const float32x4_t min_val = vmovq_n_f32(min_val_f);
     const float32x4_t max_val = vmovq_n_f32(max_val_f);
-    float32x4_t half = vdupq_n_f32(0.5f);
-    float32x4_t ret1, ret2, a, b, sign, PlusHalf, Round;
+    float32x4_t ret1, ret2, a, b;
 
     int32x4_t toint_a, toint_b;
     int16x4_t intInputVal1, intInputVal2;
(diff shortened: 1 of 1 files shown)


kernels/volk/volk_32fc_convert_16ic.h
```diff
@@ -189,16 +195,67 @@ static inline void volk_32fc_convert_16ic_neon(lv_16sc_t* outputVector, const lv
...
+    float32x4_t ret1, ret2, a, b;
+
+    int32x4_t toint_a={0,0,0,0}, toint_b={0,0,0,0};
+    int16x4_t intInputVal1, intInputVal2;
+    int16x8_t res;
+
+    for(i = 0; i < neon_iters; i++)
+        {
+            a = vld1q_f32((const float32_t*)(inputVectorPtr)); inputVectorPtr += 4;
+            b = vld1q_f32((const float32_t*)(inputVectorPtr)); inputVectorPtr += 4;
```
jdemel: Could you make this one statement per line? This would make it easier to read.

kernels/volk/volk_32fc_convert_16ic.h
```diff
@@ -173,10 +180,9 @@ static inline void volk_32fc_convert_16ic_neon(lv_16sc_t* outputVector, const lv
 
     const float32x4_t min_val = vmovq_n_f32(min_val_f);
     const float32x4_t max_val = vmovq_n_f32(max_val_f);
-    float32x4_t half = vdupq_n_f32(0.5f);
-    float32x4_t ret1, ret2, a, b, sign, PlusHalf, Round;
+    float32x4_t ret1, ret2, a, b;
 
-    int32x4_t toint_a, toint_b;
+    int32x4_t toint_a={0,0,0,0}, toint_b={0,0,0,0};
```
jdemel: Could you split this into 2 lines?

jdemel (commented):
So the performance hit comes from the fact that we respect the current rounding mode? That's a bummer.

michaelld (approved):
LGTM; please do tweak the CR/LF for initial variables settings per @jdemel 's notes.

kernels/volk/volk_32fc_convert_16ic.h
hcab14: Done.

What do you think about doing a clang-reformat of VOLK like what was done for GNURadio? Especially in the `kernels/volk` subdirectory, the code formatting is quite erratic.

michaelld:
Thanks for the tweaks. Let's let CI do its thing & @jdemel finalize his review.

kernels/volk/volk_32fc_convert_16ic.h
hcab14: I agree. It's done.

michaelld:
CI looks good. @jdemel what sayeth thou?

kernels/volk/volk_32fc_convert_16ic.h
michaelld: I'm all for doing a clang-format for code consistency!

michaelld: We also need to get the license code header to be consistent.

jdemel: Off the top of my head I think GR contains a script for license header formatting. Should we open 2 issues for that? `Make license headers more consistent` and `Apply clang-format`?

jdemel (approved):
Looks good to me

kernels/volk/volk_32fc_convert_16ic.h
michaelld: I'd say: yes! If you have time to do this, it would be great!

PR #297, opened 2019-10-29
volk_32fc_s32f_magnitude_16i MSVC

hcab14:
https://github.com/gnuradio/volk/pull/285

Hopefully fixed the MSVC error reported in https://github.com/gnuradio/volk/pull/285

### DIFF
--- a/kernels/volk/volk_32fc_s32f_magnitude_16i.h
+++ b/kernels/volk/volk_32fc_s32f_magnitude_16i.h
@@ -209,7 +209,7 @@ volk_32fc_s32f_magnitude_16i_a_sse(int16_t* magnitudeVector, const lv_32fc_t* co
   __m128 vScalar = _mm_set_ps1(scalar);
 
   __m128 cplxValue1, cplxValue2, result;
-  volatile __m128 iValue, qValue;
+  __m128 iValue, qValue;
 
   __VOLK_ATTR_ALIGNED(16) float floatBuffer[4];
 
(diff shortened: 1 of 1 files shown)


michaelld (approved):
I have no issues with this code change; it -can- make sense to use new variables to store intermediate results.

Question: Why is the `volatile` required at all here? From the new code using the new variables to store intermediate results, I don't see a good reason for this keyword. Maybe someone else does?

michaelld:
CI failure is for `qa_volk_32fc_convert_16ic` NEON ... so not this PR's fix. Do we have a PR for this failure yet?

hcab14:
> CI failure is for `qa_volk_32fc_convert_16ic` NEON ... so not this PR's fix. Do we have a PR for this failure yet?

Yes: https://github.com/gnuradio/volk/pull/296

michaelld:
Yup ... just got there myself ... thx!

hcab14:
> I have no issues with this code change; it -can- make sense to use new variables to store intermediate results.
> 
> Question: Why is the `volatile` required at all here? From the new code using the new variables to store intermediate results, I don't see a good reason for this keyword. Maybe someone else does?

Without the `volatile` keywords the compiler optimizes the order of FP operations in a way which breaks the agreement between the AVX2 and the other SIMD kernels+the generic kernel.

michaelld:
> Without the `volatile` keywords the compiler optimizes the order of FP operations in a way which breaks the agreement between the AVX2 and the other SIMD kernels+the generic kernel.

Interesting. I can see how that might hold true for the prior code, where the variables were reused. Assuming you've tested / verified its the same for the new code, even though the variables are specified in the ordering desired?

kernels/volk/volk_32fc_s32f_magnitude_16i.h
```diff
@@ -225,10 +225,10 @@ volk_32fc_s32f_magnitude_16i_a_sse(int16_t* magnitudeVector, const lv_32fc_t* co
     // Arrange in q1q2q3q4 format
     qValue = _mm_shuffle_ps(cplxValue1, cplxValue2, _MM_SHUFFLE(3,1,3,1));
 
-    iValue = _mm_mul_ps(iValue, iValue); // Square the I values
-    qValue = _mm_mul_ps(qValue, qValue); // Square the Q Values
+    volatile __m128 iValue2 = _mm_mul_ps(iValue, iValue); // Square the I values
+    volatile __m128 qValue2 = _mm_mul_ps(qValue, qValue); // Square the Q Values
```
jdemel, line 229: Looks good to me. Though, I thought that MSVC fails to compile because of the `volatile` keyword. I wonder how that PR changes that situation and how that PR might change our CI.

michaelld:
CI looks fine. I'm merging & we'll see what comes of it!

hcab14:
Can someone restart https://travis-ci.org/gnuradio/volk/jobs/604498622 which errored because of a timeout when downloading boost?

michaelld:
done 🤞

PR #300, opened 2019-11-05
feat: calc cos in AVX512F

fritterhoff:
This PR provides calculation of the cosinus values using AVX512F

### DIFF
--- a/kernels/volk/volk_32f_cos_32f.h
+++ b/kernels/volk/volk_32f_cos_32f.h
@@ -115,7 +115,7 @@ volk_32f_cos_32f_a_avx512f(float* cosVector, const float* inVector, unsigned int
 
       aVal = _mm512_load_ps(inPtr);
       // s = fabs(aVal)
-      s = _mm512_abs_ps(aVal);
+      s = _mm512_castsi512_ps(_mm512_and_si512(_mm512_castps_si512(aVal), _mm512_set1_epi32(0x7fffffff)));;
       // q = (int) (s * (4/pi)), floor(aVal / (pi/4))
       q = _mm512_cvtps_epi32(_mm512_floor_ps(_mm512_mul_ps(s, m4pi)));
       // r = q + q&1, q indicates quadrant, r gives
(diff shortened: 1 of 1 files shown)


fritterhoff:
Anybody an idea why using _mm512_and_si512 fails?

jdemel:
In TravisCI it fails for GCC 4.9. Am I right? 
This makes this whole thing difficult. I guess we need to figure out if we still support GCC 4.9 before we can merge this PR.
One general remark: Could you make sure all lines are below 80 characters. At least roughly.

fritterhoff:
Jep its GCC 4.9 in Travis. 
A possible solution would be to build a "wrapper" or something like that in a if clause for GCC 4.9. 

I'll watch out for the line length :+1:

hcab14:
The two missing functions do not point to any intrinsics but just cast `__m512` to `__m512i` and vice versa, see avx512fintrin.h

```c
extern __inline __m512i
__attribute__ ((__gnu_inline__, __always_inline__, __artificial__))
_mm512_castps_si512 (__m512 __A)
{
  return (__m512i) (__A);
}

extern __inline __m512
__attribute__ ((__gnu_inline__, __always_inline__, __artificial__))
_mm512_castsi512_ps (__m512i __A)
{
  return (__m512) (__A);
}
```

So you could do the cast just yourself and it should be ok.

jdemel:
At this point GCC 4.9 is an ancient compiler. Only some really longlong term support distros use it. I read about SLES 11 and RHEL 6? Ubuntu 16.04 (the oldest supported one) ships with GCC 5.4.
The question really is, should we require newer compilers or allow the code to be a bit more complex.

fritterhoff:
Maybe it makes sense to provide the missing methods in a seperate header and put a #if arround it, so it gets only included in this special case (GCC 4.9)?

Currently I'm trying to implement more and more kernels as AVX512-Version so probably I'll get some similar problems again and again

hcab14:
You may have a look at `include/volk/volk_avx2_intrinsics.h` and create a
similar file for AVX512.

michaelld:
Any updates on this PR?

fritterhoff:
I'll leave the current structure and will proceed with the CLA and change it probably later.

jdemel:
@flori-uni Is there any update on the CLA?
We start to see merge conflicts and I'd be more than happy to avoid them by merging your PRs.

PR #302, opened 2019-11-06
Add __clang__ specific attributes

michaelld:
From the comment in the code: AppleClang also defines __GNUC__, so do this check first.  These will probably be the same as for __GNUC__, but let's keep them separate just to be safe.

### DIFF
--- a/include/volk/volk_common.h
+++ b/include/volk/volk_common.h
@@ -4,7 +4,20 @@
 ////////////////////////////////////////////////////////////////////////
 // Cross-platform attribute macros
 ////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


jdemel (approved):
LGTM. I don't have any experience with AppleClang. It doesn't change anything for other compilers.

michaelld:
> LGTM. I don't have any experience with AppleClang. It doesn't change anything for other compilers.

Clang is starting to show up outside OSX, which is looking positive for its broader use. So I'll continue supporting it in these little ways!

michaelld:
All CI checks pass; positive review; compiler-specific change -> merging

PR #303, opened 2019-11-07
Fix liborc headers not found on Ubuntu server

amrbekhit:
When compiling Volk on Ubuntu server 19.10 for the Raspberry Pi, CMake is unable to find the liborc header files despite them being installed. This patch fixes this issue by providing PATH_SUFFIXES allowing CMake to find the headers.


jdemel (approved):
LGTM. I just tested this PR. It doesn't break things on my machine.

I assume we don't need a CLA for such a small PR @bhilburn ?

michaelld (changes requested):
This is fine. One observation that might reduce the code a little: we add `PATH_SUFFIXES` as done here, but remove the `orc-0.4` from the other 2 paths. I think that should work, and it tightens up the code a bit. @amrbekhit can you test to verify that this change works? If not, we can go with this as-is.

michaelld:
We don't need a CLA for such a small change.

amrbekhit:
> This is fine. One observation that might reduce the code a little: we add `PATH_SUFFIXES` as done here, but remove the `orc-0.4` from the other 2 paths. I think that should work, and it tightens up the code a bit. @amrbekhit can you test to verify that this change works? If not, we can go with this as-is.

Done and tested on the Raspberry Pi with no problems.

michaelld (approved):
LGTM; thanks for your contribution & quick turn around on my request!

michaelld:
All CI passes all checks (CI includes `orc` support); 2 positive reviews. Simple change. Merging.

PR #304, opened 2019-11-08
 CI cache / Intel SDE

hcab14:
- cache cross-compiled boost (ARM NEON)
- use the Intel SDE (also cached), see https://github.com/gnuradio/volk/issues/216


jdemel (approved):
I like that you removed copypasta. It's a good addition to cache Boost if it needs to be downloaded and built. Though, this makes this PR more difficult to review.

It seems like this is the way to go to use SDE for tests. All tests still pass and SDE is used. I'd say let's use it.

michaelld:
Everything looks positive here overall. Let me take a quick read through the Intel license to see what it says about this sort of use. Unless it explicitly / strictly prohibits such use, I'm OK with using it in this manner.

michaelld:
Just to check: (1) we're not distributing the Intel download, nor are we making it available for others to obtain? (2) we're not disassembling or reverse-engineering any part or portion of the Intel download? (3) we're not modifying anything in the Intel download? (4) we're just using the Intel download as if it were "static const"?

If we can honestly say "yes" to all of those, then that's roughly what the license boils down to. I find no further conditions imposed on how one may or may not agree to the licens,e nor use the Intel download once obtained.

jdemel:
(1) Yes (2) Yes! (3) Yes.
I can't see any reason to answer otherwise.

michaelld (approved):
LGTM let's try it & see if/how it works!

michaelld:
All CI passes; multiple positive reviews; nothing obvious why not to use this Intel download per its license -> merging!

PR #305, opened 2019-11-09
configpath: Abort read/write if path not found

### LINKED ISSUE #259: gracefully handle if $HOME and $APPDATA is empty and /etc/volk/volk_config is nonexistent
Currently if both environment variables and no system-wide config are empty `volk_profile` tries to use an empty path to save results. 
In any case `volk_profile` should also not save into the system-wide config.

Edit: Bug is in volk_prefs.c in the `volk_get_config_path` function.

jdemel:
This commit fixes #259.
After a call to `volk_get_config_path` check if `path[0] == 0`. If yes,
return because we could not find a volk config path.

### DIFF
--- a/apps/volk_profile.cc
+++ b/apps/volk_profile.cc
@@ -157,6 +157,10 @@ void read_results(std::vector<volk_test_results_t> *results)
 {
     char path[1024];
     volk_get_config_path(path, true);
+    if(path[0] == 0){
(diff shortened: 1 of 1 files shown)


michaelld (approved):
LGTM

michaelld:
All CI passes; small fix; positive review. Merging.

Issue #195, opened 2018-07-31
let cmake prefer Py3k for builds


jdemel:
Since the latest VOLK version builds with Py3k, I suggest to make it the preferred Python version for `cmake`.

There are some advantages
- `Mako` seems to be installed by default on Ubuntu 18.04. Thus, it cuts dependency installs.
- Py3k was identified to be the future. So let's try to make the latest software the default and use everything else as a fallback. This might also be beneficial to flesh out bugs.

How do we get there? Let `cmake` prefer Py3k over Py2k.

noc0lour:
We should definitively tell our downstream about that.

michaelld:
I'd be game for this change. Doing so keeps Volk better aligned with GNU Radio changes in this regard, beyond other benefits.

jdemel:
This issue is fixed with #306. Thus, I close this issue.

PR #306, opened 2019-11-14
Py3k: Prefer Py3k over Py2k if available

jdemel:
This is a simple change that pushes CMake to search for Python3 first.
Only if we can't find a Py3k install, we move on to search for Py2k.

This PR deals with #195

### DIFF
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -40,7 +40,7 @@ else(PYTHON_EXECUTABLE)
     find_package(PythonInterp 3)
 
     if(NOT PYTHONINTERP_FOUND)
-        find_package(PythonInterp 2)
+        find_package(PythonInterp 2.7)
     endif(NOT PYTHONINTERP_FOUND)
 
     #and if that fails use the find program routine
@@ -56,17 +56,6 @@ endif(PYTHON_EXECUTABLE)
 #make the path to the executable appear in the cmake gui
 set(PYTHON_EXECUTABLE ${PYTHON_EXECUTABLE} CACHE FILEPATH "python interpreter")
 
-#make sure we can use -B with python (introduced in 2.6)
-if(PYTHON_EXECUTABLE)
-    execute_process(
-        COMMAND ${PYTHON_EXECUTABLE} -B -c ""
-        OUTPUT_QUIET ERROR_QUIET
-        RESULT_VARIABLE PYTHON_HAS_DASH_B_RESULT
-    )
-    if(PYTHON_HAS_DASH_B_RESULT EQUAL 0)
-        set(PYTHON_DASH_B "-B")
-    endif()
-endif(PYTHON_EXECUTABLE)
 
 ########################################################################
 # Check for the existence of a python module:
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -36,12 +36,16 @@ if(PYTHON_EXECUTABLE)
 else(PYTHON_EXECUTABLE)
 
     #use the built-in find script
-    set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6)
-    find_package(PythonInterp 2)
+    set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6 3.7 3.8)
+    find_package(PythonInterp 3)
+
+    if(NOT PYTHONINTERP_FOUND)
+        find_package(PythonInterp 2)
+    endif(NOT PYTHONINTERP_FOUND)
 
     #and if that fails use the find program routine
     if(NOT PYTHONINTERP_FOUND)
-        find_program(PYTHON_EXECUTABLE NAMES python python2 python2.7 python3)
+        find_program(PYTHON_EXECUTABLE NAMES python3 python python2 python2.7)
         if(PYTHON_EXECUTABLE)
             set(PYTHONINTERP_FOUND TRUE)
         endif(PYTHON_EXECUTABLE)
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -28,6 +28,18 @@ set(__INCLUDED_VOLK_PYTHON_CMAKE TRUE)
 # or finds the interpreter via the built-in cmake module.
 ########################################################################
 #this allows the user to override PYTHON_EXECUTABLE
+
+# FUTURE TODO: With CMake 3.12+ we can simply do:
+#if(PYTHON_EXECUTABLE)
+#    set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
+#else(PYTHON_EXECUTABLE)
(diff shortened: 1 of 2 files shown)


michaelld (changes requested):
I like this change. We do need to keep Py2 around at least for Volk 2, IMHO. As such, would you add some logic to determine which `PythonInterp` was found (2 or 3), and then look just for the `PYTHON_EXECUTABLE` of that version? Thus, for example, of `PythonInterp` found is version 2, then don't look for the Py3 `PYTHON_EXECUTABLE`. Should be straight forward, I think, and not a huge change.

jdemel:
Basically, we want 2 cmake variables to be set
1. `PYTHONINTERP_FOUND` to indicate we can use Python
2. `PYTHON_EXECUTABLE` to use this specific Python version.

Previously, the code worked like this:
If `PYTHON_EXECUTABLE` is already set, e.g. by user, we just use it and `set(PYTHONINTERP_FOUND TRUE)`. This is still the same.

If we did not set `PYTHON_EXECUTABLE` we try to find it. This is where I my PR takes effect.
With `find_package(PythonInterp 3)` we search for a Py3k. This will set both cmake variables. No further action required.
Only if we can't find Py3k, we go on and search for Py2k.

Finally, if that fails as well, we try:
`find_program(PYTHON_EXECUTABLE NAMES python3 python python2 python2.7)`
This should search for executable names in order of appearance. Thus, we search for `python3` first.

Since we require Python 2.7+, we can remove 
```
#make sure we can use -B with python (introduced in 2.6)
if(PYTHON_EXECUTABLE)
    execute_process(
        COMMAND ${PYTHON_EXECUTABLE} -B -c ""
        OUTPUT_QUIET ERROR_QUIET
        RESULT_VARIABLE PYTHON_HAS_DASH_B_RESULT
    )
    if(PYTHON_HAS_DASH_B_RESULT EQUAL 0)
        set(PYTHON_DASH_B "-B")
    endif()
endif(PYTHON_EXECUTABLE)
```
in `cmake/Modules/VolkPython.cmake

I updated `.travis.yml` as well I hope the CI system does work again with my patch.

michaelld (commented):
Gotcha. Yes, that all makes sense. Good changes! Let me test locally with different Python 2 and 3 & verify before finalizing.

cmake/Modules/VolkPython.cmake
```diff
@@ -36,12 +36,16 @@ if(PYTHON_EXECUTABLE)
 else(PYTHON_EXECUTABLE)
 
     #use the built-in find script
-    set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6)
-    find_package(PythonInterp 2)
+    set(Python_ADDITIONAL_VERSIONS 3.4 3.5 3.6 3.7 3.8)
+    find_package(PythonInterp 3)
```
michaelld, line 52: according to CMake 3.16.0's `FindPythonInterp.cmake`:
```
deprecated:: 3.12
  Use :module:`FindPython3`, :module:`FindPython2` or :module:`FindPython` instead.
```
Looks like the new usage is `find_package(Python3 COMPONENTS Interpreter)`

Same for the 2.7 call.

michaelld: Looks like instead of `` we now do `set(Python3_FIND_STRATEGY VERSION)` ... which is the default so we  can just remove the `Python_ADDITIONAL_VERSIONS` I think.
```
``Python3_FIND_STRATEGY``
  This variable defines how lookup will be done.
  The ``Python3_FIND_STRATEGY`` variable can be set to empty or one of the
  following:

  * ``VERSION``: Try to find the most recent version in all specified
    locations.
    This is the default if policy :policy:`CMP0094` is undefined or set to
    ``OLD``.
  * ``LOCATION``: Stops lookup as soon as a version satisfying version
    constraints is founded.
    This is the default if policy :policy:`CMP0094` is set to ``NEW``.
```

cmake/Modules/VolkPython.cmake
```diff
@@ -36,12 +36,16 @@ if(PYTHON_EXECUTABLE)
...
+
+    if(NOT PYTHONINTERP_FOUND)
```
michaelld, line 54: This would be `Python3_Interpreter_FOUND`.

cmake/Modules/VolkPython.cmake
```diff
@@ -36,12 +36,16 @@ if(PYTHON_EXECUTABLE)
...
+        find_package(PythonInterp 2.7)
```
michaelld, line 55: Thinking somewhere in here you'd want logic setting a common `PYTHON_INTERPRETER_FOUND` and `PYTHON_EXECUTABLE` ... whichever of Py3 or Py2 was found .... variables that will be common once set regardless of which Py is found and how (old / deprecated or current new, or some future way).

michaelld (changes requested):
A few changes will bring the cmake up to date, but it's a little more complex than I'd hoped.

cmake/Modules/VolkPython.cmake
jdemel: Yeah, the new, in fact too new, way would be `find_package(Python COMPONENTS Interpreter)` [cf here](https://cmake.org/cmake/help/v3.16/module/FindPython.html#module:FindPython). This call would prefer Py3k over Py2k, thus effectively implement the same behavior I intend to implement.
I browsed through the CMake documentation and it seems this function was only added in CMake 3.12. And we currently require CMake 3.8. 

With CMake 3.12+ we could simply do this:
```
if(PYTHON_EXECUTABLE)
    set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
else(PYTHON_EXECUTABLE)
    find_package(Python COMPONENTS Interpreter)
endif(PYTHON_EXECUTABLE)

#make the path to the executable appear in the cmake gui
set(PYTHON_EXECUTABLE ${Python_EXECUTABLE} CACHE FILEPATH "python interpreter")
```
Since `grep -irnI PYTHONINTERP_FOUND .` did not reveal any use of this variable outside this section, we don't need to set this variable anymore.

I suggest to stick with my implementation for now. We can update this section after we move to CMake 3.12+ as a minimum. I don't like to make this section version dependent. This would make it more difficult to read.
@michaelld how would you proceed?

michaelld: OK yeah given the CMake min version, let's leave this code as it is. Would you please add a note about changing up the code when CMake >= 3.12? That way we're leaving something useful behind!

michaelld (changes requested):
LGTM after that comment addition.

jdemel:
I added a comment `FUTURE TODO` with a short hint on how to handle this with CMake 3.12+. I hope it'll help us then. Since Ubuntu 18.04 ships with CMake 3.11, I doubt we'll move to a more recent CMake version any time soon.

michaelld (approved):
Thanks for the comment. LGTM.

michaelld:
CI passes that we care about; failures are not due to this PR. Maintainer approves. Merging.

PR #307, opened 2019-11-15
cmake: use sha1 instead of md5 for unique target name hash

### LINKED ISSUE #238: MD5 usage causes build failure on FIPS enabled CentOS 7.4 system
On a CentOS 7.4 system with FIPS mode enabled, the MD5 hash algorithm cannot be executed and building volk fails because of the use of MD5 here:

https://github.com/gnuradio/volk/blob/dd5e293c85975d4db6718aef685b1a15de88a74f/cmake/Modules/VolkPython.cmake#L117

Changing 'md5' to either 'sha1', 'sha256', or 'sha512' resolves the problem.

michaelld:
title says it all.

Closes #238

### DIFF
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -114,7 +114,7 @@ file(TO_CMAKE_PATH ${VOLK_PYTHON_DIR} VOLK_PYTHON_DIR)
 function(VOLK_UNIQUE_TARGET desc)
     file(RELATIVE_PATH reldir ${CMAKE_BINARY_DIR} ${CMAKE_CURRENT_BINARY_DIR})
     execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "import re, hashlib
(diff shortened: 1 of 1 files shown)


jdemel (approved):
LGTM.
According to the docs `sha1` as well as `md5` are always available in `hashlib`. Let's move to a newer algorithm here.

argilo:
In cryptographic contexts, SHA1 has already been abandoned for similar reasons to MD5. So I wouldn't be surprised if `hashlib.sha1` is soon removed from FIPS-mode RHEL / CentOS as well. It would probably be safer to go with `hashlib.sha256`.

michaelld:
`sha256` it is ... available in both Py27 and Py3 `hashlib` ... I'll fix that shortly.

michaelld:
Done. When CI finished & shows success, let's merge.

michaelld:
CI passes. simple change. maintainer approval. -> merging.

Issue #90, opened 2016-06-02
cannot convert ‘ unsigned int* ’ to ‘ uint16_t* {aka short unsigned int*} ’ in argument passing


sgardner340:
I have had this error using the latest Gunradio since 05/26/2016. Looks like Volk was updated and the update conflicts with Gnuradio's gr-dtv component 

Error Log:
/home/hudson/workspace/PyBombs_Build_All_Ubuntu_Matrix/os/Ubuntu_14.04_Desktop_x86_64_VM_R/satisfy_order/deb-src/src/gr-wimax/lib/phy/phy_estimate_fine_offsets_shmidl_and_cox.cc: In function ‘ bool phy_estimate_fine_offsets_shmidl_and_cox(unsigned int, unsigned int, unsigned int, float, const gr_complex_, unsigned int, unsigned int&, double&, gr_complex_, size_t&) ’:
/home/hudson/workspace/PyBombs_Build_All_Ubuntu_Matrix/os/Ubuntu_14.04_Desktop_x86_64_VM_R/satisfy_order/deb-src/src/gr-wimax/lib/phy/phy_estimate_fine_offsets_shmidl_and_cox.cc: 45: 78: error: cannot convert ‘ unsigned int\* ’ to ‘ uint16_t\* {aka short unsigned int_} ’ in argument passing
volk_32fc_index_max_16u( &M_max, (lv_32fc_t_) &crossvec[0], num_crossvals);
^
make[2]: **\* [lib/CMakeFiles/ofdma.dir/phy/phy_estimate_fine_offsets_shmidl_and_cox.cc.o] Error 1
make[2]: **\* Waiting for unfinished jobs....
make[1]: **\* [lib/CMakeFiles/ofdma.dir/all] Error 2
make: **\* [all] Error 2

n-west:
This was fixed

PR #309, opened 2019-11-19
travis: Add native tests on arm64, s390x and ppc64le architectures

velichkov:
See https://docs.travis-ci.com/user/multi-cpu-architectures

- Test with both gcc and clang
- Add ppc64le and s390x to allow_failures as currently two tests fail
  on these two architectures

 89: Segmentation fault (core dumped)
 89/130 Test  #89: qa_volk_32fc_s32fc_multiply_32fc ...................***Failed    1.13 sec

 90: Segmentation fault (core dumped)
 90/130 Test  #90: qa_volk_32fc_s32fc_rotatorpuppet_32fc ..............***Failed    0.15 sec

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: cpp
 
+os: linux
 dist: trusty
 sudo: false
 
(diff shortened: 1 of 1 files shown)


michaelld (approved):
LGTM; thx!

Wondering what the default "dist:" is if not specified, as is (or seems to be) the case for your "Linux arm64" additions.

velichkov:
> Wondering what the default "dist:" is if not specified, as is (or seems to be) the case for your "Linux arm64" additions.

Oops, I just added `distro: bionic` to arm64 builds and forced pushed. 

According to [this](https://docs.travis-ci.com/user/reference/linux) and also my test the default `dist` is `xenial` for all architectures but because `dist: trusty` is specified at the top and probably because they don't have LXD containers with trusty for the new architectures they fallback to `bionic` and not to the default `xenial`. 

I just tested [`xenial` builds](https://travis-ci.org/velichkov/volk/builds/614819253) but the cmake from the [apt repository is too old](https://packages.ubuntu.com/xenial-updates/cmake) so on the new architectures all builds failed.  It's also interesting that there are a lot of difference between their Xenial VM and Xenial LXD images, in the VM a lot of the preinstall software seems to be updated not from the apt repository.

jdemel (commented):
Is it possible to move those `os: linux` lines to the top? Otherwise it looks good to me.  I like that you give different Jobs more meaningful names. 
I didn't know IBM hardware supports SSE/AVX. Where can I find documentation on that?
Since this is not the only change to `.travis.yml`, we should consolidate this file after all pending PRs have been dealt with.

hcab14:
Note that arm64 is already tested in subjob .7 using qemu and cross-compilation. Anyway it is good having these native tests, also because different compilers are used.

velichkov:
> Is it possible to move those `os: linux` lines to the top? 

Done. [This](https://travis-ci.org/gnuradio/volk/jobs/615651393) travis build has failed but the failures seems to be unrelated to my changes.

- https://travis-ci.org/gnuradio/volk/jobs/615651393#L1516-L1525
- https://travis-ci.org/gnuradio/volk/jobs/615651393#L1884-L1899

> I didn't know IBM hardware supports SSE/AVX. Where can I find documentation on that?

They don't. The PowerPC (ppc64le) architecture has [AltiVec](https://en.wikipedia.org/wiki/AltiVec) instructions and the [IBM Z (s390x) architecture](http://publibfi.boulder.ibm.com/epubs/pdf/dz9zr010.pdf) has SIMD instructions known as ZVector but I can't find good free documentations about them, gcc has the `-mzvector` option that I'm going to test, and [here](http://archive.ibmsystemsmag.com/mainframe/tipstechniques/systemsmanagement/simd-101/) is some magazine article.

https://travis-ci.org/gnuradio/volk/jobs/615651404#L169-L184
https://travis-ci.org/gnuradio/volk/jobs/615651406#L168-L196

I'm aware that currently there are no VOLK kernels that utilize those instructions and only the generic implementation is currently being tested in Travis.

> Since this is not the only change to `.travis.yml`, we should consolidate this file after all pending PRs have been dealt with.

The only other open PR that modifies `.travis.yml` is #306,  it adds `python3-mako` and `python3-six` to the `&common_packages` and once it gets merged I could remove those two packages from my PR and rebase the PR.

> Note that arm64 is already tested in subjob .7 using qemu and cross-compilation. 

OK.

jdemel (approved):
Changes look good to me.
I think it is a good idea to test Neon with Qemu and natively. Very much like our SDE vs native tests.
It is interesting to see if VOLK works on IBM systems, though I do worry if it is necessary to test for them as long as we don't have any kernels for these architectures. All tests run on each PR.

michaelld:
I still like this addition. Strange that just a single CI failed out of the whole bunch. I've restarted it to see if it was an anomaly. The failures don't seem to have anything to do with this PR.

michaelld:
@velichkov do you have a CLA (code license assignment) in place?

@jdemel it's not clear to me that the CLA is required for this small of a change, and one that is just CI related. Do you have any thoughts about this?

michaelld:
OK so the restarted CI passed this time. We might have a flaky test or 2. @jdemel do you know if some of the tests use random data? I honestly haven't looked, and I know I should :)

velichkov:
> do you have a CLA (code license assignment) in place?

I do have CLA for gnuradio. Do I need a separate CLA for VOLK?

michaelld:
> > do you have a CLA (code license assignment) in place?
> 
> I do have CLA for gnuradio. Do I need a separate CLA for VOLK?

Good question! I don't know!

Hopefully @bhilburn does ... tagging him to bring him into this conversation.

jdemel:
Yes, we use random data. That's one of the reasons why we have flaky kernels. Though, one vectors is used for all implementations of a kernel.
I don't know if we would need a CLA for such a small change. Since VOLK is a GNU Radio project, I assume we don't need a separate CLA. At least I hope so. 

Besides, this PR looks good to me.

bhilburn:
No further CLA needed.

michaelld:
@bhilburn :+1:

michaelld:
All CI passes. 2 approvals. No other issues. Merging.

PR #314, opened 2019-12-11
Adds NEON versions of cos sin and tan.

ast:
Adds NEON versions of cos sin and tan. The reason they are in the same commit is that they are calculated by the same underlying function.

RUN_VOLK_TESTS: volk_32f_cos_32f(131071,1)
generic_fast completed in 10.0344 ms
generic completed in 3.34385 ms
neon completed in 1.44828 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 3.34994 ms
neon completed in 1.46552 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_tan_32f(131071,1)
generic completed in 9.04345 ms
neon completed in 1.99319 ms
Best aligned arch: neon
Best unaligned arch: neon

### DIFF
--- a/include/volk/volk_neon_intrinsics.h
+++ b/include/volk/volk_neon_intrinsics.h
@@ -48,6 +48,30 @@
  *
  */
 
+/* Copyright (C) 2011  Julien Pommier
+
+  This software is provided 'as-is', without any express or implied
+  warranty.  In no event will the authors be held liable for any damages
+  arising from the use of this software.
+
+  Permission is granted to anyone to use this software for any purpose,
+  including commercial applications, and to alter it and redistribute it
+  freely, subject to the following restrictions:
+
+  1. The origin of this software must not be misrepresented; you must not
+     claim that you wrote the original software. If you use this software
+     in a product, an acknowledgment in the product documentation would be
+     appreciated but is not required.
+  2. Altered source versions must be plainly marked as such, and must not be
+     misrepresented as being the original software.
+  3. This notice may not be removed or altered from any source distribution.
+
+  (this is the zlib license)
+ 
+  _vsincosq_f32
+ 
+*/
+
 /*
  * This file is intended to hold NEON intrinsics of intrinsics.
  * They should be used in VOLK kernels to avoid copy-pasta.
@@ -77,6 +101,16 @@ static inline float32x4_t _vinvsqrtq_f32(float32x4_t x)
     sqrt_reciprocal = vmulq_f32(vrsqrtsq_f32(vmulq_f32(x, sqrt_reciprocal), sqrt_reciprocal), sqrt_reciprocal);
     
     return sqrt_reciprocal;
+}
+
+/* Inverse */
+static inline float32x4_t _vinvq_f32(float32x4_t x)
+{
+    // Newton's method
+    float32x4_t recip = vrecpeq_f32(x);
+    recip             = vmulq_f32(vrecpsq_f32(x, recip), recip);
+    recip             = vmulq_f32(vrecpsq_f32(x, recip), recip);
+    return recip;
 }
 
 /* Complex multiplication for float32x4x2_t */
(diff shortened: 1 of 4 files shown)


kernels/volk/volk_32f_cos_32f.h
```diff
@@ -673,4 +673,41 @@ volk_32f_cos_32f_generic(float* bVector, const float* aVector, unsigned int num_
...
+    float* bVectorPtr = bVector;
+    const float* aVectorPtr = aVector;
+    
+    float32x4_t b_vec;
+    float32x4_t a_vec;
+    
+    for(number = 0; number < quarter_points; number++) {
+        a_vec = vld1q_f32(aVectorPtr);
+        // Prefetch next one, speeds things up
+        __VOLK_PREFETCH(aVectorPtr+4);
```
michaelld, line 696: On the final iteration of this loop, `aVectorPtr+4` will point to the memory just beyond the end of the end of the `aVectorPtr`. There's no guarantee that this memory will have been allocated contiguously, and hence this prefetch could, maybe possibly, generate a runtime memory access error. We need a better way to do this prefetch! How much of a hit does this code take when the prefetch is disabled?

include/volk/volk_neon_intrinsics.h
```diff
@@ -147,4 +181,98 @@ static inline float32x4_t _vlogq_f32(float32x4_t x)
     return poly;
 }
 
+/* Evaluation of 4 sines & cosines at once.
```
michaelld, line 184: I'd like to start restructuring this file, starting with this PR:

When we add new code with a new license, let's append the whole piece of code including the license, inside a delineation comment such as:
```
/******************************************************************/
```
In this manner, we make it clear which sections of code are related to which license.

Once this PR is in place, I'll go through this file & rearrange it for how the code was added historically, keeping the licenses and related codes together.

michaelld (changes requested):
Great (re)start! One minor change (formatting), and one more serious change. Thx!

kernels/volk/volk_32f_cos_32f.h
ast: Good find. Probably not a big hit.

A quick `grep` reveals this bug is present in other kernels (also kernels I didn't touch).

I can't think of a way of pre-fetching without using a test. Probably best to remove them.

hcab14:
As far as I understand, prefetching beyond the size of the array does not do any harm. Since in some cases prefetch helps I would keep it.

Also found this reference
https://developer.arm.com/docs/100616/latest/functional-description/level-1-memory-system/data-prefetching

where 3 different ARM prefetch instructions are mentioned: PLD, PLDW, PLI, PRFM. Might be worth to understand their difference and if they can help.

ast:
Well that certainly would be good if that's the case...

michaelld:
Right now `VOLK_PREFETCH` for GCC and Clang uses the compiler directive `__builtin_prefetch`, which according to < https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html >:
```
Data prefetch does not generate faults if addr is invalid, but the address expression itself must be valid. For example, a prefetch of p->next does not fault if p->next is not a valid address, but evaluation faults if p is not a valid address.
```
So since the pointer value is valid, it doesn't matter (at least for GCC) whether the dereferenced pointer is valid. I'd guess the same holds for Clang. Hence, let's keep the prefetching.

balister:
We should confirm the prefetches have benefit and if not remove them. Less code, less to go wrong.

jdemel:
@ast could you do a quick benchmark?
I hope this does help though. Synthetic benchmarks are always difficult to apply to real world problems.
Actually, I'd be in favor of leaving `VOLK_PREFETCH` in. Even if we can't confirm an immediate performance gain. Mostly because we can only run synthetic benchmarks and we could already confirm, that a prefetch does not cause any harm. Though, we could add a warning comment to that line in code. If it ever fails, we have a reminder why.

ast:
Here's some output of `volk_test_all volk_32f_sin_32f` there *might* be a tiny effect of `VOLK_PREFETCH` but to conclude anything I think we would need a much longer running benchmark presenting mean and sd.

It would be nice if it could output data for GNUPlot also...

## With prefetch

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.61041 ms
neon completed in 3.69726 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.61528 ms
neon completed in 3.70294 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 3.42371 ms
neon completed in 1.49339 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.49853 ms
neon completed in 3.65365 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 3.40036 ms
neon completed in 1.46633 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.5379 ms
neon completed in 3.65135 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 3.40914 ms
neon completed in 1.46822 ms
Best aligned arch: neon
Best unaligned arch: neon



## Without

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.64338 ms
neon completed in 3.76444 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.63676 ms
neon completed in 3.65867 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.48042 ms
neon completed in 3.65633 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 3.43347 ms
neon completed in 1.52261 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.52323 ms
neon completed in 3.65681 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.58769 ms
neon completed in 3.65761 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 3.40534 ms
neon completed in 1.49404 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1)
generic completed in 8.58091 ms
neon completed in 3.70993 ms
Best aligned arch: neon
Best unaligned arch: neon

hcab14:
Instead of ctest you can also run `volk_profile -R volk_32f_sin_32f` which uses more than one iteration.

Your timings show that the CPU frequency is changing. If you are using an RPi3 this can be avoided by adding the following two lines to `/boot/config.txt`
```
core_freq=400
core_freq_min=400
```

ast:
Ah ok, of course.

These are on a Raspberry Pi 4 (32 bit OS though). I suspect the difference between NEON and generic would be quite a bit larger on a Raspberry Pi 3.

## With prefetch 

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1987)
generic completed in 6704.08 ms
neon completed in 2902.71 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1987)
generic completed in 6709.87 ms
neon completed in 2922.11 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1987)
generic completed in 6710.97 ms
neon completed in 2905.6 ms
Best aligned arch: neon
Best unaligned arch: neon

## Without

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1987)
generic completed in 6655.48 ms
neon completed in 2906.17 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1987)
generic completed in 6723.19 ms
neon completed in 2903.95 ms
Best aligned arch: neon
Best unaligned arch: neon

RUN_VOLK_TESTS: volk_32f_sin_32f(131071,1987)
generic completed in 6706.33 ms
neon completed in 2903.89 ms
Best aligned arch: neon
Best unaligned arch: neon

michaelld:
Looks like there is nominal performance gain from the prefetch. I'm still in favor of leaving the `VOLK_PREFETCH` in.

michaelld:
Please fix the header per my request & I think we'll be ready to call this PR good to go!

michaelld:
I've restarted the CI; the errors were with the build system itself.

jdemel:
@ast could you fix the header as per @michaelld request? Then we'll be ready to merge your PR. I guess you'll be happy as soon as it is merged.

ast:
Sorry was on holiday and forgot about it! I will take a look at it!

michaelld:
No major rush .. but, ping on this PR ... any progress?

include/volk/volk_neon_intrinsics.h
ast: Ok sorry about the long delay, lots of things going on.

I will fix this now but rereading your comment I'm not 100% sure how you would like it. Do you want me to move the license down to right above the relevant function and enclose that whole block in a delineation comment? Then I will also have to refactor the MIT licensed code.

michaelld: This: Move the license down to right above the relevant function and enclose that whole block in a delineation comment. Yes that means refactoring a little. Hoping that won't be too difficult.

michaelld: @jdemel  and I discussed this briefly. My goal was to make it extremely clear which license/code went together; since the function is named in the license, it's not so critical to keep them together. Given that there's already another license / code combo split here (and probably elsewhere), and that the function is named for both licenses: I'm removing this request.

michaelld (approved):
LGTM

michaelld:
@jdemel please review and if you approve the merge

jdemel (approved):
LGTM

PR #315, opened 2019-12-11
Update README

ast:
Update README to reflect how to build on Raspberry Pi and the importance of running volk_profile

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,59 @@
 [![Build Status](https://travis-ci.org/gnuradio/volk.svg?branch=master)](https://travis-ci.org/gnuradio/volk) [![Build status](https://ci.appveyor.com/api/projects/status/5o56mgw0do20jlh3/branch/master?svg=true)](https://ci.appveyor.com/project/gnuradio/volk/branch/master)
+
+# Welcome to VOLK!
+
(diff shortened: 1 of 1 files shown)


ast:
@michaelld please consider this or something similar. No one understands how to build on Raspberry Pi and there are no instructions anywhere. Also people forget to run volk_profile.

michaelld:
@ast Yeah I like this idea. We don't provide a lot of README in Volk, so putting something into the main one makes sense to me.

michaelld (approved):
LGTM ; we might tweak it a bit down the road for other toolchains, but hopefully this provides a starting point for ARM-based users. Thx!

balister:
I don't see any problems with this, if you used the same toolchain for both, do you see any meaingful differences in performance?

ast:
@balister I think so but I haven't tested, at least for some kernels Cortex A72 has hardware sqrt and if I remember correctly some other relevant functions. I guess it really depends on the compiler...

Would be interesting to bench mark.

I have noticed that the Coretex-A72 is *significantly* faster with the generic kernels and is really closing the gap between hand optimized code quite a bit...

jdemel (approved):
LGTM.

michaelld:
CI doesn't matter since this is a comment change. CLA in place. Maintainers approve. Merging.

PR #319, opened 2019-12-20
Fix arm neon detection

### LINKED ISSUE #323: Build is broken for armv7 architecture
I've dissected this problem to commit https://github.com/gnuradio/volk/commit/4589103155a27f9e0805232fcb43d1b0a067c013

It seems that the new policies in cmake 3.8 breaks something.

Good cmake:
```
-- Performing Test neon_compile_result
-- Performing Test neon_compile_result - Success
-- Performing Test have_neonv7_result
-- Performing Test have_neonv7_result - Success
-- Performing Test have_neonv8_result
-- Performing Test have_neonv8_result - Failed
-- CPU is armv7, Overruled arch neonv8
-- Available architectures: generic;hardfp;neon;neonv7;orc;norc
-- Available machines: generic_orc;neon_orc;neonv7_hardfp_orc
-- BUILD TYPE = RELEASE
```

Bad cmake:
```
-- Performing Test neon_compile_result
-- Performing Test neon_compile_result - Success
-- Performing Test have_neonv7_result
-- Performing Test have_neonv7_result - Failed
-- Performing Test have_neonv8_result
-- Performing Test have_neonv8_result - Failed
-- Available architectures: generic;hardfp;neon;neonv7;neonv8;orc;norc
-- Available machines: generic_orc;neon_orc;neonv7_hardfp_orc;neonv8
-- BUILD TYPE = RELEASE
```

The build fails when volk_machine_neonv8.c is attempted to be compiled.

A workaround for this is to change line 21 in the top level CMakeLists.txt from:
cmake_minimum_required(VERSION 3.8)
to:
cmake_minimum_required(VERSION 3.7)

trabucayre:
Closes: https://github.com/gnuradio/volk/issues/323

For ARM target (ARMV7 for example), compiled, for example with buildroot and cmake-3.15, volk compile fails with something like:

volk/kernels/volk/volk_32u_reverse_32u.h:352:49: warning: implicit declaration of function ‘vrbitq_u8’; did you mean ‘vbicq_u8’? [-Wimplicit-function-declaration]
volk/kernels/volk/volk_32u_reverse_32u.h:352:38: error: incompatible type for argument 1 of ‘vreinterpretq_u32_u8’
  uint32x4_t z = vreinterpretq_u32_u8(vqtbl1q_u8(vrbitq_u8(vreinterpretq_u8_u32 (x)),

This failure is due to wrong NEON detection:
-- Performing Test neon_compile_result
-- Performing Test neon_compile_result - Success
-- Performing Test have_neonv7_result
-- Performing Test have_neonv7_result - Failed
-- Performing Test have_neonv8_result
-- Performing Test have_neonv8_result - Failed

CMakeError.log shows:
buildroot-build/CMakeFiles/CMakeTmp/src.c:1:12: error: « asm » not declared (first use in the function)
 int main(){asm volatile("vrev32.8 q0, q0");}

This error is due to -std=c11 passed to the compiler.

To fix this issue, the first commit changes asm by __asm__

This issue reveal a second issue: If cmake fails for both architecture, neonv7 and neonv8 are left in available_archs.
This second commit check this case and suppress both flavour.


balister:
Nice catch. Does __asm__ work for builds where -std=c11 is not passed to the compiler?

trabucayre:
I've tried manually to compile both C code:
- with -std=c11 -> fails
- without -std=c11 -> success
- with -std=gnu11 -> success

But, depending on the build system, It may be impossible to avoid -std=c11, and, finally, everywhere in volk __asm__ is used.

hcab14:
> But, depending on the build system, It may be impossible to avoid -std=c11, and, finally, everywhere in volk **asm** is used.

in the kernels `__VOLK_ASM` is/should be used.

You may consider to use `#include <volk/volk_common.h>`, like here:
https://github.com/hcab14/volk/commit/a09183c387f2a90db8c2c874b484445a5552c782

Does this help? (I have not tested it for ARM NEON)

trabucayre:
I've just tested with __VOLK_ASM and the neonvX detection work fine.
Would you contribute this commit and I align my PR on this or I update my PR with __VOLK_ASM?
Thanks

jdemel (changes requested):
Thanks for taking this up.

Since `neonv7` and `neonv8` are XOR requirements and currently, we prefer `neonv8`, I suggest to comment on this.

jdemel:
Which compiler did you use? I assume GCC. I'm surprised it works with gnu11 but not with c11. Did GCC change the c11 flag?

balister:
What is blocking this? It has tripped up another person.

Regarding preferring v8, no machines support both.

lib/CMakeLists.txt
```diff
@@ -267,6 +267,11 @@ if(neon_compile_result)
     if (have_neonv8_result)
         OVERRULE_ARCH(neonv7 "CPU is armv8")
     endif()
+
```
jdemel, line 266: According to @balister ARM CPUs may support both architectures. In this case, I'd suggest to separately check for both architectures and call `OVERRULE_ARCH` on each one separately. We should be able to have both architectures compiled and let VOLK choose the preferred one according to `volk_profile` results.

jdemel (changes requested):
Currently the control flow is as follows
1. check if NEON compiles
2. check if v7 and/or v8 compiles.
3. Assume only `have_neonv8_result` XOR `have_neonv7_result` is true.
4. Exclude the other architecture.

@balister since CPUs can have both, we should allow both architectures and let `volk_profile` do a performance test and follow that result.
Otherwise, in case both architectures are available, we end up with neither being compiled.
Or is the assumption 3. correct?

Either way, we need to clarify this and put in a comment on the assumptions we make here.

lib/CMakeLists.txt
```diff
@@ -255,17 +255,25 @@ check_c_source_compiles("#include <arm_neon.h>\nint main(){ uint8_t *dest; uint8
...
+    check_c_source_compiles("#include <volk/volk_common.h>\n int main(){__VOLK_ASM volatile(\"vrev32.8 q0, q0\");}"
                             have_neonv7_result )
-    check_c_source_compiles("int main(){asm volatile(\"sub v1.4s,v1.4s,v1.4s\");}"
+    check_c_source_compiles("#include <volk/volk_common.h>\n int main(){__VOLK_ASM volatile(\"sub v1.4s,v1.4s,v1.4s\");}"
                             have_neonv8_result )
 
-    if (have_neonv7_result)
-        OVERRULE_ARCH(neonv8 "CPU is armv7")
+    # if both are supported we let VOLK choose the preferred one
+    if (NOT (have_neonv7_result EQUAL have_neonv8_result))
```
michaelld: Seems like the logic here should get cleaned up, otherwise I like what this PR does!

michaelld (changes requested):
change request is in the comment

lib/CMakeLists.txt
trabucayre: Hi.
Sorry, but I'm unable to see which change I need to add. When neonv7 xor neonv8 is true (ie only one version supported) we need to unset the neonvX not supported no?
Or maybe it's the comment just before the if?

jdemel: According to @balister ARM CPUs may support both architectures.
I hope he can comment on this. Options I'm aware of are:
1. Test for NEON versions separately and only run `OVERRULE_ARCH` if particular version is not found. We may have both versions enabled.
2. Decide which arch we prefer. Test for this arch, if available prune the other arch. No further test necessary. If only fallback arch is available, use this one.
3. Something else.

In any case, we need to make a decision. I hope we can make an informed decision.

lib/CMakeLists.txt
balister: I think you misunderstood what I wrote, or I wrote the wrong thing. I can not imagine a situation where compile test for neonv7 and neonv8 both succeed.

lib/CMakeLists.txt
jdemel: So only one of those tests will succeed. I suggest to test for both archs and just call `OVERRULE_ARCH` on any arch where the test fails.

Now I wonder why we did this general NEON check first and then tried to deduce the version. I'd expect the code to check for each version and if it is not available just prune it from the list.

Can you put in a comment that clearly states, we expect only one of those archs to be present at any given time?

Are there any other implications?

I've read a bunch on NEON and ARM so far. Basically, we can have 2 states:
1. 32bit with aarch32 and neonv7
2. 64bit with aarch64 and neonv8

They exclude each other. It wouldn't be reasonable to try to compile for both archs. My conclusion is that we can just check for specific NEON versions and call `OVERRULE_ARCH` on that architecture if it is not available. I hope that simplifies this code.

lib/CMakeLists.txt
jdemel: This solution looks good to me now. @balister, do you agree too?

michaelld: This solution LGTM now

michaelld (approved):
LGTM ; let's see what @balister says

jdemel (approved):
This PR LGTM too. @michaelld approved it as well.
@balister, do you have any further comments or do you approve it as well?

balister:
Looks OK to me.

jdemel:
Great. Everyone approves this PR. Merging.

PR #318, opened 2019-12-13
Add volk_32fc_accumulator_s32fc

Ka-zam:
Adding accumulator for complex vectors.

~/src/volk$ ~/local/bin/volk_profile -n -R 32fc_accu

Warning: this IS a dry-run. Config will not be written!
RUN_VOLK_TESTS: volk_32fc_accumulator_s32fc(131071,1987)
u_avx completed in 63.012 ms
u_sse completed in 122.686 ms
generic completed in 245.437 ms
Best aligned arch: u_avx
Best unaligned arch: u_avx
Warning: this was a dry-run. Config not generated

PR #322, opened 2020-01-03
Implement volk_32_ stddev_and_mean_32f_x2

### LINKED ISSUE #290: volk_32f_stddev_and_mean_32f_x2 inaccurate kernel
When given input with E[X]^2 ~= E[X^2], which is not particularly uncommon, the stddev calculation fails with a -nan result for both
1) volk_32f_stddev_and_mean_32f_x2
2) volk_32f_s32f_stddev_32f

Upon inspection of the code this happens because the following identity is used:
Var(X)  = E[(X−mu)^2] = E[X^2]−E[X]^2

which according to [https://dl.acm.org/citation.cfm?id=3221269.3223036](url) should never be used in practice as it suffers from catastrophic cancellation.

Test run where naive_f is the two-pass approach of first calculating the mean and then estimating the variance give the following output
~~~~
in[0]: 1.000000000
in[1]: 0.999998987
in[2]: 1.000002027
in[3]: 0.999997020
in[4]: 1.000004053
in[5]: 0.999994993
in[6]: 1.000005960
in[7]: 0.999993026
in[8]: 1.000007987
in[9]: 0.999991000

naive_f
mu:
0.999999523
omega:
0.000005311

volk_std_and_mean
mu:
0.999999523
omega:
-nan
~~~~

Also of note might be the fact that we should probably compute the sample variance and not the population variance. 
[https://en.wikipedia.org/wiki/Variance#Sample_variance](url)

Ka-zam:
I think this is ready to be merged now. Still haven't heard anything about a CLA though.

Performance on Ryzen 1700
```
VOLK version: 2.0
vlen :2025
iters:10000

NAME              TIME      MEAN         STDDEV
Naive Double      5.88 10000.0400391    0.9831967
Naive Float       5.18 10000.0361328    0.9832074
VOLK GEN          3.19 10000.0361328    0.9831734
VOLK U_SSE        0.58 10000.0390625    0.9832364
VOLK U_AVX        0.57 10000.0410156    0.9831885
```

Performance on rpi4 aarch64
```
VOLK version: 2.0
vlen :2025
iters:10000

NAME              TIME      MEAN         STDDEV
Naive Double      8.20 10000.0000000    0.9896870
Naive Float       8.16 10000.0068359    0.9897105
VOLK GEN          9.62 10000.0068359    0.9896759
VOLK NEON         3.08 10000.0000000    0.9897031
```

As a comparison here is the performance of the existing kernel:
```
VOLK version: 2.0
vlen :2025
iters:10000

NAME              TIME      MEAN         STDDEV
Naive Double      5.93 9999.9931641    0.9961484
Naive Float       5.17 9999.9912109    0.9961522
VOLK GEN          3.22 9999.9912109         -nan
VOLK A_SSE        0.83 9999.9951172         -nan
VOLK U_AVX        0.53 9999.9941406         -nan
```


jdemel:
Thanks for working on this!
I assume we still need to wait for your CLA as discussed in #318 .
I think we can still do a review here and merge this PR as soon as the CLA is in place.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2012, 2014 Free Software Foundation, Inc.
+ * Copyright 2019 Free Software Foundation, Inc.
```
jdemel: Just add 2019 and don't delete previous years.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
-      newMean += *aPtr++;
-    }
-    newMean /= num_points;
-    stdDev /= num_points;
-    stdDev -= (newMean * newMean);
-    stdDev = sqrtf(stdDev);
-  }
-  *stddev = stdDev;
-  *mean = newMean;
+update_square_sum_1_val(float* S, const float* T, const uint32_t* N, const float* val){
```
jdemel: could you be more verbose on parameter names? This increases readability.

Since only scalars are used, I suggest to pass parameters by value. This does improve readability and this shouldn't incur a performance penalty. Thus, I suggest to return the value to be added to `S` and add it in the calling function.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+  float n = (float) (*N);
+  (*S) += 1.f/( n*(n + 1.f) ) * ( n*(*val) - (*T) ) * ( n*(*val) - (*T) );
+}
 
+static inline void
+square_add(float* S,  const float* T0, const float* S1, const float* T1, const uint32_t* N){
+  float n = (float) (*N);
+  (*S) += (*S1);
+  (*S) += .5f/n*( (*T0) - (*T1) )*( (*T0) - (*T1) );
```
jdemel: Same as previous comment. Please use more verbose parameter names and call-by-value.

If you encounter a performance hit with the call-by-value change, we can still go back. At the moment I'd assume a compiler will honor the `inline` keyword. In this case the function is optimized out anyways.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
 }
-#endif /* LV_HAVE_AVX */
 
+static inline void
+accrue_result( float* S, float* T, const uint32_t N_accumulators, const uint32_t N_partition) {
```
jdemel: Please, be more verbose about parameter names.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -51,7 +51,7 @@
  *
  *   // Use a normal generator with 0 mean, stddev 1
```
jdemel, line 53: Can you add references to the employed algorithm here too?

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
-    }
-    newMean /= num_points;
-    stdDev /= num_points;
-    stdDev -= (newMean * newMean);
-    stdDev = sqrtf(stdDev);
+  // Youngs and Cramer's Algorithm for calculating std and mean
+  //   T is the running sum of values
+  //   S is the running square sum of values
+  //   Using the methods discussed here:
+  //   https://doi.org/10.1145/3221269.3223036
```
jdemel: This algorithm is used in all versions, correct? I suggest to move this comment to the top.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+  const float* in_ptr = inputBuffer;
+
+  unsigned int number = 1;
+
+  __VOLK_ATTR_ALIGNED(16) float T[8] = {0.f};
+  __VOLK_ATTR_ALIGNED(16) float S[8] = {0.f};
+
+  if (num_points < 8) {   
+    T[0] = (*in_ptr++);
+    goto FINALIZE; }
```
jdemel: Please use an `if else` statement or some other means to skip over your optimized loop. In either case don't use `goto`. A compiler does know how to optimize such `if else` statements. If someone wants to alter that code in the future, this person will have a hard time with those `goto`s. They make readability worse.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+    x1_reg = _mm_sub_ps(x1_reg, T1_acc);
+    x1_reg = _mm_mul_ps(x1_reg, x1_reg);
+    x1_reg = _mm_mul_ps(x1_reg, f_reg);
+    S1_acc = _mm_add_ps(S1_acc, x1_reg);
+  }
+
+  _mm_store_ps(&T[0], T0_acc);
+  _mm_store_ps(&T[4], T1_acc);
+  _mm_store_ps(&S[0], S0_acc);
+  _mm_store_ps(&S[4], S1_acc);
```
jdemel: Those variables `S` and `T` should have more verbose names. Preferably ones that make it clear that these are local variables. I started to wonder why an aligned store is used in an unaligned kernel but those variables are required to be aligned internally.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+
+  accrue_result( S, T, 8, eigth_points);
+
+  number = eigth_points*8;
+
+  FINALIZE:
+  for (; number < num_points; number++) {
+    update_square_sum_1_val(&S[0], &T[0], &number, in_ptr);
+    T[0] += (*in_ptr++);
+  }
```
jdemel: `S` is updated in the function, while `T` is update here, this makes this section more confusing. Please rewrite your function to call-by-value and update both `S` and `T` here.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+  __m128 T1_acc = _mm_loadu_ps(in_ptr);
+  in_ptr += 4;
+  __m128 S0_acc = _mm_setzero_ps();
+  __m128 S1_acc = _mm_setzero_ps();
+  __m128 v0_reg, v1_reg;
+  __m128 x0_reg, x1_reg;
+  __m128 f_reg;
+
+
+  for(;number < eigth_points; number++) {
```
jdemel: Since `number = 1` here, I suggest to use this variable only in this for loop and thus bind it to this scope.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+  _mm_store_ps(&T[4], T1_acc);
+  _mm_store_ps(&S[0], S0_acc);
+  _mm_store_ps(&S[4], S1_acc);
+
+  accrue_result( S, T, 8, eigth_points);
+
+  number = eigth_points*8;
+
+  FINALIZE:
+  for (; number < num_points; number++) {
```
jdemel: I suggest to bind `number` to the scope of this for loop. Also, I suggest to remove `FINALIZE` as the target for the previous `goto` together with `goto`.
Since this is a rather difficult to catch subtlety how `number` is defined, I suggest to find a more verbose control flow that doesn't rely on `goto` here.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+  const float* in_ptr = inputBuffer;
+
+  unsigned int number = 1;
+
+  __VOLK_ATTR_ALIGNED(16) float T[8] = {0.f};
+  __VOLK_ATTR_ALIGNED(16) float S[8] = {0.f};
+
+  if (num_points < 8) {   
+    T[0] = (*in_ptr++);
+    goto FINALIZE; }
```
jdemel: This is only the case if SIMD can't be used. This should be equivalent to a call to `generic`. I suggest to move this check to the top, call `generic` in case we do not have enough input and then just return.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+
+    f_reg = _mm256_set1_ps(  1.f/( n*np1 ) );
+    x0_reg = _mm256_set1_ps(np1);
+    
+    T0_acc = _mm256_add_ps(T0_acc, v0_reg);
+    
+    x0_reg = _mm256_mul_ps(x0_reg, v0_reg);
+    x0_reg = _mm256_sub_ps(x0_reg, T0_acc);
+    x0_reg = _mm256_mul_ps(x0_reg, x0_reg);
+    x0_reg = _mm256_mul_ps(x0_reg, f_reg);
```
jdemel: This section is executed twice on different data. A meta intrinsic would make the kernel more readable.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
```
jdemel: meta intrinsics for AVX would reduce copypasta from aligned to unaligned kernel. Same applies for SSE.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -73,325 +73,513 @@
...
+    __VOLK_PREFETCH(in_ptr + 8);
+
+    v1_reg = _mm256_load_ps(in_ptr);
+    in_ptr += 8;
+    __VOLK_PREFETCH(in_ptr + 8);
+
+    float n   = (float) number;
+    float np1 = n + 1.f;
+
+    f_reg = _mm256_set1_ps(  1.f/( n*np1 ) );
```
jdemel: I suggest spaces around operators.

lib/kernel_tests.h
```diff
@@ -113,7 +113,7 @@ std::vector<volk_test_case_t> init_test_list(volk_test_params_t test_params)
     QA(VOLK_INIT_TEST(volk_32f_s32f_power_32f,                        test_params))
     QA(VOLK_INIT_TEST(volk_32f_sqrt_32f,                              test_params_inacc))
     QA(VOLK_INIT_TEST(volk_32f_s32f_stddev_32f,                       test_params_inacc))
-    QA(VOLK_INIT_TEST(volk_32f_stddev_and_mean_32f_x2,                test_params_inacc))
+    QA(VOLK_INIT_TEST(volk_32f_stddev_and_mean_32f_x2,                test_params.make_tol(1e-4)))
```
jdemel, line 116: So the kernels are now more accurate? That's really nice.

jdemel (changes requested):
Thanks for working on this!

I made quite a couple of comments. Mostly, I suggest to get rid of `goto`. It took me quite a while to just understand why it was there. I'd argue it is unnecessary and should be replaced.

Intrinsics that capture the bulk of the SIMD kernels would also be a suggestion. This would reduce copypasta between aligned and unaligned versions. Also, the suggested meta intrinsic would be called twice in every iteration and further reduce copypasta.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: :+1:

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: Agree

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: Agree

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: OK

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: ok

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: i just branch to generic instead

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: started but i won't have time to finish for some time

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: OK, but I rewrote it to return float now (the new S) like discussed above. In that case we have a function that returns a value and also alters the value of one of it's parameters. That seems strange. It also has to be in that order or the algorithm won't work. 

Is this better?
```
  for (; number < num_points; number++) {
    float val = (*in_ptr);
    PSquareSumLoc[0] = update_square_sum_1_val(PSquareSumLoc[0], PSumsLoc[0], number, val);
    PSumsLoc[0]     += val;
    in_ptr++;
  }
```

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: I'm changing all instances to 
```
  size_t points_done = eigth_points*8;

  for (; points_done < num_points; points_done++) {
    float val = (*in_ptr);
    PSquareSumLoc[0] = update_square_sum_1_val(PSquareSumLoc[0], PSumsLoc[0], points_done, val);
    PSumsLoc[0]     += val;
    in_ptr++;
  }
```

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: ok

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: I don't know how to write these, are there examples?

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: ok

lib/kernel_tests.h
Ka-zam: Yes, they are more accurate and most of all for signals with high mean to variance ratio. That's what made me realize there was a problem in the first place. I made measurements what were too good, e.g. low variance.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: ok

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: ok

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
jdemel: Indeed it's strange to update a parameter in a function. 
My suggestion was to rewrite your function such that it only computes a return value and does not do anything else. Outside this function you update both variables.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
jdemel: [AVX intrinsics](https://github.com/gnuradio/volk/blob/master/include/volk/volk_avx_intrinsics.h) can be found here. I suggest you add your intrinsics in the appropriate files and just include them in the appropriate sections.

lib/kernel_tests.h
jdemel: That's good.

michaelld:
Any word on the CLA?

Ka-zam:
No word

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: This should be fine now

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: Done

Ka-zam:
> Any word on the CLA?

I sent another email to Ben.

include/volk/volk_avx_intrinsics.h
```diff
@@ -125,4 +125,14 @@ _mm256_polar_fsign_add_llrs(__m256 src0, __m256 src1, __m128i fbits){
     return dst;
 }
 
+static inline __m256
+_mm256_accumulate_square_sum(__m256 sq_acc, __m256 acc, __m256 val, __m256 rec, __m256 aux)
```
jdemel: The function name should end with `_ps` like the corresponding intrinsics do. That'll make it more in line with those intrinsics.

include/volk/volk_neon_intrinsics.h
```diff
@@ -147,4 +147,13 @@ static inline float32x4_t _vlogq_f32(float32x4_t x)
     return poly;
 }
 
+static inline float32x4_t 
+_neon_accumulate_square_sum(float32x4_t sq_acc, float32x4_t acc, float32x4_t val,
```
jdemel: Here, the function name ending should be `_f32`.

include/volk/volk_sse_intrinsics.h
```diff
@@ -46,4 +46,14 @@ _mm_magnitude_ps(__m128 cplxValue1, __m128 cplxValue2){
   return _mm_sqrt_ps(_mm_magnitudesquared_ps(cplxValue1, cplxValue2));
 }
 
-#endif /* INCLUDE_VOLK_VOLK_SSE_INTRINSICS_H_ */
+static inline __m128
+_mm_accumulate_square_sum(__m128 sq_acc, __m128 acc, __m128 val, __m128 rec, __m128 aux)
```
jdemel: `_ps` ending.

jdemel (changes requested):
Thanks for the update. I hope you'll have your CLA in place soon and we'll be able to merge this PR.

michaelld:
There are conflicts with recent changes. Please rebase to the current Volk master.

jdemel (changes requested):
Thanks for working on this!
A minor thing I saw, we open function `{` on a new line. I'd like to stick to one code style.

jdemel:
@Ka-zam  Is there any update on the CLA?
We start to see merge conflicts and I'd be more than happy to avoid them by merging your PRs.
Could you rebase your PR onto current master? That would trigger more CI and would make it easier to review. Thanks!

jdemel:
@Ka-zam I assume there's still no update on your CLA?

Anyways, we adopted a new system: DCO. Thus, you only need to sign off your commits. `git commit -s` for new commits. The details are explained in `DCO.txt`. With `git commit -s` you add a `Signed off by ...` line to the end of your commit message. You can add

I got this approach to sign previous commits from [stackoverflow](https://stackoverflow.com/a/58599218)
```
git rebase --exec 'git commit --amend --no-edit -n -s' -i HEAD~25
```
I updated the option `-s` to lower case. Also, `HEAD~25` should refer to your 25 commits. It would be great if you could rebase your PR onto current master. In that case we could just merge it.

Ka-zam:
> @Ka-zam I assume there's still no update on your CLA?
> 
> Anyways, we adopted a new system: DCO. Thus, you only need to sign off your commits. `git commit -s` for new commits. The details are explained in `DCO.txt`. With `git commit -s` you add a `Signed off by ...` line to the end of your commit message. You can add
> 
> I got this approach to sign previous commits from [stackoverflow](https://stackoverflow.com/a/58599218)
> 
> ```
> git rebase --exec 'git commit --amend --no-edit -n -s' -i HEAD~25
> ```
> 
> I updated the option `-s` to lower case. Also, `HEAD~25` should refer to your 25 commits. It would be great if you could rebase your PR onto current master. In that case we could just merge it.

OK, I will rebase and do the PR. Sorry about the delay!

Ka-zam:
I think I will close this PR and open a new one based off of the master branch. The rebasing failed miserably.

jdemel:
Now that #457 is merged, I assume we can close this PR and also the corresponding issue #290 . If there's anything missing, feel free to re-open them.

PR #325, opened 2020-01-08
Clang Attribute Fix

michaelld:
For attribute, change from clang to "clang but not MSC".

MSC attributes are covered in a later clause

Simple / trivial change.

Ref: https://github.com/gnuradio/volk/commit/7abdaa5da2ab236ed8acfe361d04021050146778#commitcomment-36706063


jdemel:
It's a simple change. LGTM. Merging.

PR #334, opened 2020-01-26
alloc: Use C11 aligned_alloc

### LINKED ISSUE #193: Document tolerance for a passed null pointer in a specific function parameter
[The function “volk_free”](https://github.com/gnuradio/volk/blob/17277ed3130e9b2592da06c681657e63b5590c56/lib/volk_malloc.c#L61 "Implementation of volk_free()") could be implemented by a simple call of [the function “free”](http://pubs.opengroup.org/onlinepubs/9699919799/functions/free.html "Free allocated memory").
Thus it seems that this function implementation inherits the property to tolerate a null pointer which was received in the parameter “aptr”.

:bulb: How do you think about to express this detail also in [the corresponding programming interface description](https://github.com/gnuradio/volk/blob/ddd9c348440f11317619da2802c369a8e980bef9/include/volk/volk_malloc.h#L59 "Update candidate in the API description")?

jdemel:
We require C11. Thus, we can now rely on `aligned_alloc` and `free`.
[aligned_alloc](https://en.cppreference.com/w/c/memory/aligned_alloc)
Also, the documentation is updated accordingly.

Closes #193

### DIFF
--- a/include/volk/volk_malloc.h
+++ b/include/volk/volk_malloc.h
@@ -32,8 +32,20 @@ __VOLK_DECL_BEGIN
  * \brief Allocate \p size bytes of data aligned to \p alignment.
  *
  * \details
- * Internally, we use C11 `aligned_alloc` to allocate aligned memory.
- * Thus, we rely on C11 syntax and compilers.
+ * We use C11 and want to rely on C11 library features,
+ * namely we use `aligned_alloc` to allocate aligned memory.
+ * see: https://en.cppreference.com/w/c/memory/aligned_alloc
+ *
+ * Not all platforms support this feature.
+ * For Apple Clang, we fall back to `posix_memalign`.
+ * see: https://linux.die.net/man/3/aligned_alloc
+ * For MSVC, we fall back to `_aligned_malloc`.
+ * see: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=vs-2019
+ *
+ * Because of the ways in which volk_malloc may allocate memory, it is
+ * important to always free volk_malloc pointers using volk_free.
+ * Mainly, in case MSVC is used. Consult corresponding documentation
+ * in case you use MSVC.
  *
  * \param size The number of bytes to allocate.
  * \param alignment The byte alignment of the allocated memory.
--- a/include/volk/volk_malloc.h
+++ b/include/volk/volk_malloc.h
@@ -32,22 +32,8 @@ __VOLK_DECL_BEGIN
  * \brief Allocate \p size bytes of data aligned to \p alignment.
  *
  * \details
- * Because we don't have a standard method to allocate buffers in
- * memory that are guaranteed to be on an alignment, VOLK handles this
- * itself. The volk_malloc function behaves like malloc in that it
- * returns a pointer to the allocated memory. However, it also takes
- * in an alignment specification, which is usually something like 16 or
- * 32 to ensure that the aligned memory is located on a particular
- * byte boundary for use with SIMD.
- *
- * Internally, the volk_malloc first checks if the compiler is C11
- * compliant and uses the new aligned_alloc method. If not, it checks
- * if the system is POSIX compliant and uses posix_memalign. If that
- * fails, volk_malloc handles the memory allocation and alignment
- * internally.
- *
- * Because of the ways in which volk_malloc may allocate memory, it is
- * important to always free volk_malloc pointers using volk_free.
+ * Internally, we use C11 `aligned_alloc` to allocate aligned memory.
+ * Thus, we rely on C11 syntax and compilers.
  *
  * \param size The number of bytes to allocate.
  * \param alignment The byte alignment of the allocated memory.
@@ -57,7 +43,13 @@ VOLK_API void *volk_malloc(size_t size, size_t alignment);
 
 /*!
  * \brief Free's memory allocated by volk_malloc.
- * \param aptr The aligned pointer allocaed by volk_malloc.
+ *
+ * \details
+ * We rely on C11 syntax and compilers and just call `free`
+ * on memory that was allocated with `aligned_alloc`.
+ * Thus, `volk_free` inherits the same behavoir `free` exhibits.
+ *
+ * \param aptr The aligned pointer allocated by volk_malloc.
  */
 VOLK_API void volk_free(void *aptr);
 
--- a/lib/volk_malloc.c
+++ b/lib/volk_malloc.c
@@ -29,33 +29,23 @@
 /*
  * For #defines used to determine support for allocation functions,
  * see: http://linux.die.net/man/3/aligned_alloc
+ *
+ * C11 features:
+ * see: https://en.cppreference.com/w/c/memory/aligned_alloc
 */
 
-// Otherwise, test if we are a POSIX or X/Open system
-// This only has a restriction that alignment be a power of 2 and a
-// multiple of sizeof(void *).
-#if _POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600 || HAVE_POSIX_MEMALIGN
 
 void *volk_malloc(size_t size, size_t alignment)
 {
-  void *ptr;
-
-  // quoting posix_memalign() man page:
-  // "alignment must be a power of two and a multiple of sizeof(void *)"
-  // volk_get_alignment() could return 1 for some machines (e.g. generic_orc)
-  if (alignment == 1)
+  if (alignment == 1){
     return malloc(size);
+  }
 
-  int err = posix_memalign(&ptr, alignment, size);
-  if(err == 0) {
-    return ptr;
+  void *ptr = aligned_alloc(alignment, size);
+  if(ptr == NULL) {
+    fprintf(stderr, "VOLK: Error allocating memory (aligned_alloc was POSIX)\n");
   }
-  else {
-    fprintf(stderr,
-            "VOLK: Error allocating memory "
-            "(posix_memalign: error %d: %s)\n", err, strerror(err));
-    return NULL;
-  }
+  return ptr;
 }
 
 void volk_free(void *ptr)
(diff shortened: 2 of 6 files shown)


lib/volk_malloc.c
```diff
@@ -29,102 +29,30 @@
 /*
  * For #defines used to determine support for allocation functions,
  * see: http://linux.die.net/man/3/aligned_alloc
+ *
```
michaelld: I think the just above comment and link can now be removed too.

lib/volk_malloc.c
```diff
@@ -29,102 +29,30 @@
...
 
 void *volk_malloc(size_t size, size_t alignment)
 {
-  void *ptr;
-
-  // quoting posix_memalign() man page:
-  // "alignment must be a power of two and a multiple of sizeof(void *)"
-  // volk_get_alignment() could return 1 for some machines (e.g. generic_orc)
-  if (alignment == 1)
+  if (alignment == 1){
```
michaelld, line 56: (1) Thinking to just remove this statement, since one can use `aligned_alloc` with alignment of 1. I'd guess internal to that function it checks the alignment & if == 1 then uses `malloc`, so I'd bet we're not saving a lot of CPU cycles here. And, it cleans up our code.
(2) Thinking making both `volk_malloc` and `volk_free` inline functions, since they are so short to begin with; no real need to have them be proper functions, IMHO. Maybe `__VOLK_ATTR_INLINE` since that already exists in `volk_common.h`? If not that, then we can define a `__VOLK_FUNC_INLINE` or something like it & use that here.

lib/volk_malloc.c
```diff
@@ -29,102 +29,30 @@
...
-  /* Get the real pointer */
-  info = (struct block_info *)(((uintptr_t)ptr) - sizeof(struct block_info));
-
-  /* Release real pointer */
-  free(info->real);
-}
 
-#endif // _POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600 || HAVE_POSIX_MEMALIGN
 
 //#endif // _ISOC11_SOURCE
```
michaelld: I'd say to remove these as well. They seem like cruft with all of the macros above removed.

michaelld (changes requested):
Good first shot! A few changes I think are in order. Or at least discussion of how we might remove more code and other comments.

include/volk/volk_malloc.h
```diff
@@ -32,22 +32,8 @@ __VOLK_DECL_BEGIN
...
- * Internally, the volk_malloc first checks if the compiler is C11
- * compliant and uses the new aligned_alloc method. If not, it checks
- * if the system is POSIX compliant and uses posix_memalign. If that
- * fails, volk_malloc handles the memory allocation and alignment
- * internally.
- *
- * Because of the ways in which volk_malloc may allocate memory, it is
- * important to always free volk_malloc pointers using volk_free.
+ * Internally, we use C11 `aligned_alloc` to allocate aligned memory.
+ * Thus, we rely on C11 syntax and compilers.
```
rear1019: MSVC never fails to disappoint when it comes to (lacking) support of “old” features [1]:
> C11 The Universal CRT implemented the parts of the C11 Standard Library that are required by C++17, with the exception of C99 strftime() E/O alternative conversion specifiers, C11 fopen() exclusive mode, and C11 aligned_alloc(). The latter is unlikely to be implemented, because C11 specified aligned_alloc() in a way that's incompatible with the Microsoft implementation of free(): namely, that free() must be able to handle highly aligned allocations.

In other words: Special handling for MSVC is required.

[1] https://docs.microsoft.com/en-us/cpp/overview/visual-cpp-language-conformance?view=vs-2019

jdemel: It is really disappointing. Thanks for the link.
I added a workaround, i.e. a preprocessor macro. I really don't like it but as long as we want non-standard compliant compilers to be able to build VOLK that's what we have to do.

lib/volk_malloc.c
jdemel: Since GCC throws this warning, I'd suggest we don't inline this function. 
```
warning: always_inline function might not be inlinable [-Wattributes]
 __VOLK_ATTR_INLINE void volk_free(void *ptr)
                         ^~~~~~~~~
warning: always_inline function might not be inlinable [-Wattributes]
 __VOLK_ATTR_INLINE void *volk_malloc(size_t size, size_t alignment)
```

That would also be consistent with its current implementation. Further, I'd assume current compilers inline such functions whenever appropriate and possible.

jdemel:
Thanks for your comments. I updated this PR accordingly. Apparently, CI passes again.

carlesfernandez:
I think function `print_malloc()` in [volk_config_info.cc](https://github.com/gnuradio/volk/blob/d3e9614cb9b0f5b2f65a7f9b77ede7421557c530/apps/volk-config-info.cc#L39) should be changed to something like:

```
void print_malloc()
{
    // You don't want to change the volk_malloc code, so just copy the if/else
    // structure from there and give an explanation for the implementations
    std::cout << "Used malloc implementation: ";
#if defined(_MSC_VER)
    std::cout << "aligned_malloc" << std::endl;
#else
    std::cout << "aligned_alloc" << std::endl;
#endif
}
```

carlesfernandez:
Also, [lines 68 to 78](https://github.com/gnuradio/volk/blob/d3e9614cb9b0f5b2f65a7f9b77ede7421557c530/lib/CMakeLists.txt#L68-L78) in `lib/CMakelists.txt` should be removed, since `HAVE_POSIX_MEMALIGN` is not used elsewhere.

michaelld:
Good catches, @carlesfernandez ! Yes, let's get those changes into this PR as well. They all fit together nicely.

jdemel:
Thanks for those hints @carlesfernandez . I pushed an update with your proposed changes.

carlesfernandez:
I'm getting segfaults when trying this on macOS.  Apparently, aligned_alloc is not available in Apple's C11  :-(

carlesfernandez:
Please check https://github.com/carlesfernandez/volk/tree/aligned-alloc-macos for a solution that works for macOS

michaelld:
Well carp. Let me test on my macOS box (10.14) ...

michaelld:
Indeed there is no `aligned_alloc` on macOS 10.14 latest Xcode, which means it just doesn't exist on macOS. Hence: instead of assuming it works, let's add a test for it & augment the current `volk_malloc` codes to use it if found, but otherwise do whatever it does right now. I think this means we'll need to check in `volk_free` for if the argument is NULL or not & deal with it accordingly.

jdemel:
@carlesfernandez thanks for your hints.

[Our old doc reference](https://linux.die.net/man/3/aligned_alloc) states, that `posix_memalign`'ed pointers can just be passed to `free`.
Thus, I'd really want to keep the `volk_free` function as simple as possible. Only MSVC gets a special case.

I looked through the proposed changes. (sidenote: I'm all in favor of auto-formatting code but it makes it harder to get to the essentials of the commit.) 

GCC or Clang will return True for this call:
```
include(CheckSymbolExists)
check_symbol_exists(posix_memalign stdlib.h HAVE_POSIX_MEMALIGN)
```
But usually, we want to use `aligned_alloc` which is also defined in [`stdlib.h`](https://en.cppreference.com/w/c/memory/aligned_alloc).
Thus, I'll check for a solution that prefers `aligned_alloc` and just uses `posix_memalign` as a fallback.

jdemel:
So, all tests pass with the newest version. Unfortunately, we don't have CI for macos yet. Thus, I'd like to ask you to test the newest version, @carlesfernandez @michaelld .
I tried to follow @carlesfernandez solution but with some differences that reflect my considerations.

carlesfernandez:
Hi @jdemel, I confirm that your solution works on macOS 10.15, latest Xcode.

michaelld (approved):
LGTM

michaelld:
@rear1019 do these changes work for you / on Windows?

carlesfernandez:
Damn! 

In macOS, ```CHECK_SYMBOL_EXISTS(aligned_alloc stdlib.h USE_ALIGNED_ALLOC)``` sets  ```USE_ALIGNED_ALLOC``` to ```1```,  so it happily compiles but then you get runtime errors.

A possible solution is:

```
if(NOT (${CMAKE_SYSTEM_NAME} MATCHES "Darwin"))
    CHECK_SYMBOL_EXISTS(aligned_alloc stdlib.h USE_ALIGNED_ALLOC)
endif()
```

 since posix_memalign seems to be there on macOS, and works well.

michaelld:
well drat! so we need a "runtime" check the symbol ... OSX can be so crappy sometimes!

jdemel:
Since AppVeyor passes, I assume we have a fix for Windows MSVC. At least it runs the generic kernels and thus, it must use `volk_malloc`.
But why would `CHECK_SYMBOL_EXISTS(aligned_alloc stdlib.h USE_ALIGNED_ALLOC)` basically return true if this is not the case? At the very least I'd expect a system to clearly communicate its capabilities.

@michaelld, @carlesfernandez  could you add a commit to this PR or publish it on GitHub that fixes the issue for MacOS? I don't have a MacOS box and thus, I can't test it.
I'm still

michaelld:
@jdemel yeah we'll figure out how to get OSX stuff working. Apple has a bad habit of using default compiler settings that don't barf on bad symbols, but either can't be linked or don't work in runtime.

carlesfernandez:
I pushed a change that seems to work well (runtime-tested on macOS 10.15 and 10.14) at https://github.com/jdemel/volk/pull/1

posix_memalign *should* work fine also in older macOS versions, but I have nowhere to test that.

michaelld:
:+1: from me to merge in @carlesfernandez's PR as noted. It will work with pretty much any usable OSX install.

jdemel:
Thanks for that fix @carlesfernandez I merged it. Let's see how our CI works and I assume this PR is ready to merge. fingers crossed.

michaelld (approved):
Works for me on OSX

michaelld:
Relevant CI passes. Maintainer approval. Merging.

Issue #329, opened 2020-01-14
volk_32fc_s32fc_x2_rotator_32fc 10 dB noise in results


ghostop14:
When testing the rotator block with a straight signal source I noticed that the volk implementation volk_32fc_s32fc_x2_rotator_32fc has almost 10 dB of noise that builds between -133 and -123 dB that is not present in the CPU-only version.  Testing Flowgraph: signal source -> throttle -> rotator -> Frequency Sink, samp_rate=2.4e6, frequency rotation 100e3.  Running on an i7-7700HQ CPU on Ubuntu 18 64-bit.  You'll see it start okay, then build, reach a peak, then "break", then start the cycle over again.

To see the difference, compile the gr-blocks/rotator_cc_impl.cc with the "#if 0" and with "#if 1" in the work function (test code looks like it's already there to flip back and forth between CPU and volk).

Just a suspicion but I saw similar behavior on GPUs with floats where the CPU is maintaining double precision accuracy but the GPU was using true IEEE spec 9 decimal point floats.  The difference in accuracy was enough to introduce similar noise.

jdemel:
Is this related to #331 ?
-123 dB noise is still way below our usual target. We compromise speed for accuracy sometimes. But usually with below `10^-7` deviation from our reference. -123dB would translate into smth like `10^-12`.
Also, -123dB is way above the dynamic range of a 16bit ADC and the high speed ADCs we see in hardware are around 12 bit. Just assuming a 16bit DAC, we would probably see a noise floor at -100db and on the receiver side, we would need an SNR in the range of 100dB to even get close to a situation where this inaccuracy would effect our system.
Correct me if I'm wrong but this is an issue we can only see in theory. Maybe my whole train of thought regarding ADCs etc. is wrong. Also, I've personally never done any simulation that is even close to 100dB SNR and I can't picture a situation where this is currently useful.

ghostop14:
Hi @jdemel , not sure if it's related to 331 or not.  It could be.  I've tested on both GR 3.9 and GR 3.8 versions of volk (I think 3.9 uses a newer version) just to confirm.  It's not just theoretical, if you create a new flowgraph, set samp_rate to 2.4M, then just connect a signal source to a throttle to a rotator (freq shift -100e3 in my test), then a GUI Frequency Sink (without adjusting the dB range so you see 0-> -140), you'll see it.  It'll start fine, then the whole signal will get an oscillation that builds, peaks, then appears to "break" or snap back to normal then cycle again.

[...]

PR #341, opened 2020-02-02
Increase rotator precision test to 1e-6

ghostop14:
During testing, it appeared that the rotator was showing accuracy
to 1e-5 for the avx kernel.  (the generic kernel matched the CPU
calculations well, so I don't think there's any issue there).
This shorter precision was resulting in cycling noise of
about 10dB when tested in GNU Radio with a flowgraph consisting of
signal source->throttle->rotator->frequency sink.  While the noise
starts to appear at about -100 dB it is very prominent around
-120 dB.  If the rotator is used to frequency shift very weak
satellite or radio astronomy signals, this low-level noise
could potentially manifest in the integrated data.  This PR is
suggesting considering rotator accuracy to 1e-6 to trivialize
the resulting noise.


ghostop14:
This PR is the followup to issue #329 at @jdemel's recommendation to open a PR to start the conversation on rotator precision.

jdemel (changes requested):
Unfortunately, this breaks CI. Thus, we can't just increase our precision requirement.
The generic and SIMD versions seem to deviate after a few undred samples. My first suggestions would be to incorporate #331 in your PR because these to issues align.
Secondly, I suggest to align the logic in all kernels. Mostly, use the normalize def variable in all kernels to renormalize after the same number of samples.
Then, test different renormalize values. According to our CI 128 instead of 512 would do the trick. I suggest this approach and I hope it becomes clear why I suggest to incorporate a solution to #331 in this PR.

Further, I suggest to update the `volk_32fc_s32fc_rotatorpuppet_32fc` to normalize `phase` as well as `phase_inc` before a call to the actual implementation. Since all reported errors appear before first renormalization, this might be an issue as well. The scalar value `std::polar(1.0f, 0.1f)` is obviously not normalized.

I hope the suggested changes get us closer to a solution. If we need to do renormalization more often,  we need to discuss how precise we need this kernel to be.

ghostop14:
Definitely sounds reasonable that they're related.  The flowgraph described in the PR is the best (and easiest) way to see the issue.  I'll leave it to folks with the volk skillz to determine the most appropriate solution.   Unfortunately I'm out of my element in the non-generic code.

jdemel:
I had a look at this issue and [my work](https://github.com/jdemel/volk/commit/5c43de76ee0a219e8ce9c06f5b26b8b91bfec567) so far is inconclusive.

My observations so far:
1. Make sure all input scalars are normalized -> no improvement
2. Trigger renormalization after same num_points -> no improvement
3. Lower tolerance values move the first point that breaks this bound further to the front. It is not obvious why this would happen so far.

Since SSE kernels tend to break later, approx. factor 2, maybe the
generic kernels accumulates errors faster. That's just a wild guess
though.

This does probably need a MWE that gets us closer to the root. For now, we can't just lower the tolerance value.

ghostop14:
Agreed.  Something's going on.  I'm glad you guys are on it.

PR #342, opened 2020-02-02
Improve performance of generic kernels with complex multiply

ghostop14:
This PR incorporates the cx-limited-range checks for complex
multiply similar to GR pull request 3114.  Native complex
multiply math has a significant amount of processing overhead
ensuring that INF and NAN conditions are not encountered.
In some tests on a modern processor, these extra tests cause
a simle c = a * b complex calculation to take as much as 20%
of the processing time in a loop.  The -fcx-limited-range
compiler flag, when available ensures that the c = a * b
calculation is reduced to just the appropriate math.  The
changes to the makefile ensure that the compiler supports it
before enabling it.  On embedded debian-based platforms where
the generic kernel may get chosen, this should have a significant
positive impact performance for complex multiply kernels.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,19 @@ set(CMAKE_CXX_STANDARD 11)
 
 enable_testing()
 
+# Disable complex math NaN/INFO range checking for performance
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag(-fcx-limited-range HAVE_CX_LIMITED_RANGE)
+if(HAVE_CX_LIMITED_RANGE)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcx-limited-range")
+endif(HAVE_CX_LIMITED_RANGE)
+
+include(CheckCCompilerFlag)
+check_c_compiler_flag(-fcx-limited-range HAVE_C_LIMITED_RANGE)
+if(HAVE_C_LIMITED_RANGE)
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fcx-limited-range")
+endif(HAVE_C_LIMITED_RANGE)
+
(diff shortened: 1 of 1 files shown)


jdemel (changes requested):
This flag should be set for the C compiler as well. Besides, this seems like a reasonable change.
Also, we need to make sure this change does not introduce flaky tests. CI tells us that test 59 aka dot_prod fails on ARM.

ghostop14:
Added the C flag as well.  Did the arm test say why it failed?  I actually think the platforms where the most value will be seen is on platforms that choose the generic kernel like on some of the ARMs.  Was that the only one that failed?

CMakeLists.txt
```diff
@@ -30,6 +30,14 @@ set(CMAKE_CXX_STANDARD 11)
 
 enable_testing()
 
+# Disable complex math NaN/INFO range checking for performance
+include(CheckCXXCompilerFlag)
+check_cxx_compiler_flag(-fcx-limited-range HAVE_CX_LIMITED_RANGE)
```
jdemel, line 35: I would assume C and C++ compilers are basically the same piece of software (gcc/g++) but I'd still add an extra test for the C compiler.
Besides, this looks good.

ghostop14: Good call.  Added separate check.

jdemel (approved):
LGTM

jdemel:
@michaelld do you approve this PR?

michaelld:
Let me test on OSX ... guessing it'll be fine ... but who knows with Apple Clang?

michaelld (approved):
LGTM

michaelld:
The CMake code is sollid. On Apple Clang, these flags don't work, but that shouldn't stop us from using them. Let's integrate this & see what happens ...

michaelld:
Relevant CI passes. 2 approvals. Merging.

PR #344, opened 2020-02-09
rotator: Refactor AVX kernels

### LINKED ISSUE #331: generic volk_32fc_s32fc_x2_rotator_32fc does not normalize rotator on short sequences
Moved from https://github.com/gnuradio/gnuradio/issues/1019 .

jdemel:
Fix #331

This started as some work to track down the root cause of #329 . As a first result, it fixes #331 .
Apparently, these two issues are not as closely related as previously suspected.

Besides a fix for the generic kernel, this PR includes quite a bit of refactoring for better readibility of the AVX u/a rotator kernels.

### DIFF
--- a/include/volk/volk_avx_intrinsics.h
+++ b/include/volk/volk_avx_intrinsics.h
@@ -56,12 +56,11 @@ _mm256_complexconjugatemul_ps(__m256 x, __m256 y){
 static inline __m256
 _mm256_normalize_ps(__m256 val)
 {
-  __m256 tmp1, tmp2;
-  tmp1 = _mm256_mul_ps(val, val);
-  tmp2 = _mm256_hadd_ps(tmp1, tmp1);
-  tmp1 = _mm256_shuffle_ps(tmp2, tmp2, 0xD8);
-  tmp2 = _mm256_sqrt_ps(tmp1);
-  return _mm256_div_ps(val, tmp2);
+  __m256 tmp1 = _mm256_mul_ps(val, val);
+  tmp1 = _mm256_hadd_ps(tmp1, tmp1);
+  tmp1 = _mm256_shuffle_ps(tmp1, tmp1, _MM_SHUFFLE(3, 1, 2, 0)); // equals 0xD8
+  tmp1 = _mm256_sqrt_ps(tmp1);
+  return _mm256_div_ps(val, tmp1);
 }
 
 static inline __m256
--- a/include/volk/volk_avx_intrinsics.h
+++ b/include/volk/volk_avx_intrinsics.h
@@ -54,6 +54,17 @@ _mm256_complexconjugatemul_ps(__m256 x, __m256 y){
 }
 
 static inline __m256
+_mm256_normalize_ps(__m256 val)
+{
+  __m256 tmp1, tmp2;
+  tmp1 = _mm256_mul_ps(val, val);
+  tmp2 = _mm256_hadd_ps(tmp1, tmp1);
+  tmp1 = _mm256_shuffle_ps(tmp2, tmp2, 0xD8);
+  tmp2 = _mm256_sqrt_ps(tmp1);
+  return _mm256_div_ps(val, tmp2);
+}
+
+static inline __m256
 _mm256_magnitudesquared_ps(__m256 cplxValue1, __m256 cplxValue2){
   __m256 complex1, complex2;
   cplxValue1 = _mm256_mul_ps(cplxValue1, cplxValue1); // Square the values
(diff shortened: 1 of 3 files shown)


include/volk/volk_avx_intrinsics.h
```diff
@@ -53,6 +53,17 @@ _mm256_complexconjugatemul_ps(__m256 x, __m256 y){
   return _mm256_complexmul_ps(x, y);
 }
 
+static inline __m256
+_mm256_normalize_ps(__m256 val)
+{
+  __m256 tmp1, tmp2;
```
rear1019: There is no need for two temporary variables.

include/volk/volk_avx_intrinsics.h
```diff
@@ -53,6 +53,17 @@ _mm256_complexconjugatemul_ps(__m256 x, __m256 y){
...
+  tmp1 = _mm256_mul_ps(val, val);
+  tmp2 = _mm256_hadd_ps(tmp1, tmp1);
+  tmp1 = _mm256_shuffle_ps(tmp2, tmp2, 0xD8);
```
rear1019: You might consider using the ``_MM_SHUFFLE`` macro to make the shuffle mask more readable. ``_MM_SHUFFLE(3, 1, 2, 0)`` should be equal to ``0xD8``.

kernels/volk/volk_32fc_s32fc_x2_rotator_32fc.h
```diff
@@ -432,103 +432,64 @@ static inline void volk_32fc_s32fc_x2_rotator_32fc_a_avx(lv_32fc_t* outVector, c
...
+    __m256 aVal, phase_Val, z;
 
     phase_Val = _mm256_loadu_ps((float*)phase_Ptr);
-    inc_Val = _mm256_set_ps(lv_cimag(incr), lv_creal(incr),lv_cimag(incr), lv_creal(incr),lv_cimag(incr), lv_creal(incr),lv_cimag(incr), lv_creal(incr));
-    const unsigned int fourthPoints = num_points / 4;
+    
+    const __m256 inc_Val = _mm256_set_ps(lv_cimag(incr), lv_creal(incr),
+                                         lv_cimag(incr), lv_creal(incr),
+                                         lv_cimag(incr), lv_creal(incr),
+                                         lv_cimag(incr), lv_creal(incr));
```
rear1019, line 442: ``_mm256_castpd_ps(_mm256_broadcast_sd((const double*)&incr))`` (see e.g. ``volk_32fc_x2_s32f_square_dist_scalar_mult_32f_a_avx()``). On the other hand, the compiler probably figures this out.

Also applies to the unaligned version.

rear1019 (approved):
It’s up to you whether to ignore my nit-picking or not.

kernels/volk/volk_32fc_s32fc_x2_rotator_32fc.h
```diff
@@ -98,18 +98,17 @@ static inline void volk_32fc_s32fc_x2_rotator_32fc_generic(lv_32fc_t* outVector,
...
+
         (*phase) /= hypotf(lv_creal(*phase), lv_cimag(*phase));
-#endif
     }
     for(i = 0; i < num_points%ROTATOR_RELOAD; ++i) {
         *outVector++ = *inVector++ * (*phase);
         (*phase) *= phase_inc;
     }
-
+    if(i){
```
michaelld, line 108: Do you really mean `if(i)` here? Unless `num_points` or `ROTATOR_RELOAD` is 0, then this will always be true. Or am I misunderstanding?

kernels/volk/volk_32fc_s32fc_x2_rotator_32fc.h
jdemel: I didn't change the implementation here. But I like this solution because it is verbose without any assumptions. Thus, I'd prefer to leave it as is.

jdemel:
@rear1019 thanks for your comments. I've updated my PR accordingly.

michaelld (approved):
LGTM ; mostly wondering about the one place I commented.

kernels/volk/volk_32fc_s32fc_x2_rotator_32fc.h
jdemel: At first, I thought this is unnecessary but
`if(num_points%ROTATOR_RELOAD == 0)`
the loop just in front of it sets `i = 0` but then runs zero iterations. In this case, we just normalized phase in L102.

I could remove this `if` statement (and ignore that special case) but that would be inconsistent with the other kernels. Thus, I decided to leave it as is.

jdemel: @michaelld is this comment resolved? In that case, we can merge this PR.

michaelld: yes ... thx for the ping ... I'll do one final review & then deal with this

michaelld:
Relevant CI passes. Comments addressed. Maintainer Approval. Merging.

PR #349, opened 2020-02-21
Various tweaks

michaelld:
3 commits of small / simple tweaks. Overall functionality is not changed.

### DIFF
--- a/include/volk/volk_common.h
+++ b/include/volk/volk_common.h
@@ -4,7 +4,17 @@
 ////////////////////////////////////////////////////////////////////////
 // Cross-platform attribute macros
 ////////////////////////////////////////////////////////////////////////
-#if defined(__clang__) && (!defined(_MSC_VER))
+#if _MSC_VER
+#  define __VOLK_ATTR_ALIGNED(x) __declspec(align(x))
+#  define __VOLK_ATTR_UNUSED
+#  define __VOLK_ATTR_INLINE     __forceinline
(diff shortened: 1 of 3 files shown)


include/volk/volk_common.h
```diff
@@ -4,7 +4,17 @@
 ////////////////////////////////////////////////////////////////////////
 // Cross-platform attribute macros
 ////////////////////////////////////////////////////////////////////////
-#if defined(__clang__) && (!defined(_MSC_VER))
```
jdemel, line 7: I thought we had this line due to some bug with clang and MSVC.  Obviously, it works with Clang, GCC and MSVC at the moment (I rely on CI here). And I assume it works with Apple-Clang. I just want to double-check.

jdemel (commented):
LGTM. Just one line to check if it was there for a reason.

include/volk/volk_common.h
michaelld: Yes your memory is correct! So was I did was move this macro first, so that if MSVC then those attributes are used.  Next up is Clang (and thus, !MSVC), because -some- clang (but not all) also define __GNUC__; then finally __GNUC__ & then the default if none of them. The reordering takes care of our prior issue wit MSVC, and reduces macro complexity.

jdemel: Great. then, this is resolved.

jdemel (approved):
LGTM

PR #350, opened 2020-02-23
clang-format: Apply clang-format

### LINKED ISSUE #335: Apply clang-format
I suggest we take the clang format file from GR and apply it to all our C/C++ files. Further, we should set up a GitHub Action to apply clang-format to all PRs.

jdemel:
This commit adds `.clang-format` from GNU Radio and also adds and modifies the corresponding github worker to apply clang-format on all PRs.

Then run: `find . -regex '.*\.\(c\|cc\|cpp\|cxx\|h\|hh\)' -exec clang-format -style=file -i {} \;`


michaelld:
@jdemel can you fix the conflicts here, then I'll review & get this merged before anything else messes it up. Then we'll have to rebase pretty much all of the active PRs ... for better or for worse ... but I'd prefer to get this reformatting done ... so doing it first ...

jdemel:
I can redo this PR. That's no big deal. Though, I want to make sure we can set up a GitHub Action that runs clang-format on every PR like we have it for GR. In order to do so, we need to set up this worker. 

At first I tried to copy the GitHub action from GR and include it with this PR. But I have no idea what will happen if this is included because the action does not get triggered in that case. Currently my idea is to test this in my fork and have this worker ready as soon as we merge this PR.

jdemel:
@michaelld I updated this PR such that clang-format is applied to the most recent master and we have a GitHub Action in place that runs formats everything. Took me a while to figure out how it works.

michaelld:
Nice! There's no way I can actually review this ... so ... since CI all passes, including the new formatting check, I'm going to merge as is & we can clean up if/as necessary down the road. Thx!

PR #352, opened 2020-02-23
Refactor reported 'shadowed' variables

jdemel:
If VOLK is compiled with `-Wshadow` a lot of shadowed variables are reported. This PR is an effort to reduce the number of those reports. It would be great to eliminate all such occurrences and also compile with more warnings enabled.

### DIFF
--- a/apps/volk_option_helpers.cc
+++ b/apps/volk_option_helpers.cc
@@ -1,6 +1,25 @@
-//
-// Created by nathan on 2/1/18.
-//
+/* -*- c++ -*- */
+/*
+ * Copyright 2018-2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * GNU Radio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * GNU Radio is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with GNU Radio; see the file COPYING.  If not, write to
+ * the Free Software Foundation, Inc., 51 Franklin Street,
+ * Boston, MA 02110-1301, USA.
+ */
+
 
 #include "volk_option_helpers.h"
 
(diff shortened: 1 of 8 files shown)


jdemel:
I just pushed another commit which gets rid of all warnings due to shadowed variables.

I had to move the `volk_32f_s32f_s32f_mod_range_32f_generic` to the top of the file. Otherwise, it can't be used in the other kernels to reduce copypasta.

drmpeg:
You missed one in some NEON code.
```
In file included from /home/ubuntu/xfer/gnuradio/build/volk/lib/volk_machine_neonv7_hardfp_orc.c:62:
/home/ubuntu/xfer/gnuradio/volk/kernels/volk/volk_16u_byteswap.h: In function ‘volk_16u_byteswap_neon’:
/home/ubuntu/xfer/gnuradio/volk/kernels/volk/volk_16u_byteswap.h:239:14: warning: declaration of ‘output’ shadows a previous local [-Wshadow]
     uint16_t output = *inputPtr;
              ^~~~~~
/home/ubuntu/xfer/gnuradio/volk/kernels/volk/volk_16u_byteswap.h:227:21: note: shadowed declaration is here
   uint16x8_t input, output;
                     ^~~~~~
```

apps/volk_option_helpers.cc
```diff
@@ -1,6 +1,25 @@
-//
-// Created by nathan on 2/1/18.
-//
+/* -*- c++ -*- */
+/*
+ * Copyright 2018 Free Software Foundation, Inc.
```
michaelld: I'd suggest changing this to 2018-2020 ... there are enough changes herein to warrant it!

apps/volk_option_helpers.h
```diff
@@ -1,6 +1,24 @@
-//
-// Created by nathan on 2/1/18.
-//
+/* -*- c++ -*- */
+/*
+ * Copyright 2018 Free Software Foundation, Inc.
```
michaelld: 2018-2020

apps/volk_option_helpers.h
```diff
@@ -23,12 +41,12 @@ typedef enum
...
-    option_t(std::string longform, std::string shortform, std::string msg, void (*callback)(std::string));
-  option_t(std::string longform, std::string shortform, std::string msg, std::string printval);
+  option_t(std::string t_longform, std::string t_shortform, std::string t_msg, void (*t_callback)());
+    option_t(std::string t_longform, std::string t_shortform, std::string t_msg, void (*t_callback)(int));
+    option_t(std::string t_longform, std::string t_shortform, std::string t_msg, void (*t_callback)(float));
+    option_t(std::string t_longform, std::string t_shortform, std::string t_msg, void (*t_callback)(bool));
+    option_t(std::string t_longform, std::string t_shortform, std::string t_msg, void (*t_callback)(std::string));
+  option_t(std::string t_longform, std::string t_shortform, std::string t_msg, std::string t_printval);
 
   std::string longform;
```
michaelld: Another option here that might be fewer changes is to move these to `d_` and leave the instantiation arguments alone. This would better fit into the "GR/Volk class variable naming" way of doing things ... but clearly either would work.

michaelld (changes requested):
Overall LGTM ; see about adding in DrMPEG's ARM code find, as well as my suggestions.

apps/volk_option_helpers.h
jdemel: I opted to `t_` because members are used in quite a few places and I tried to keep the changes to a minimum.

jdemel:
> You missed one in some NEON code.
> 
> ```
> In file included from /home/ubuntu/xfer/gnuradio/build/volk/lib/volk_machine_neonv7_hardfp_orc.c:62:
> /home/ubuntu/xfer/gnuradio/volk/kernels/volk/volk_16u_byteswap.h: In function ‘volk_16u_byteswap_neon’:
> /home/ubuntu/xfer/gnuradio/volk/kernels/volk/volk_16u_byteswap.h:239:14: warning: declaration of ‘output’ shadows a previous local [-Wshadow]
>      uint16_t output = *inputPtr;
>               ^~~~~~
> /home/ubuntu/xfer/gnuradio/volk/kernels/volk/volk_16u_byteswap.h:227:21: note: shadowed declaration is here
>    uint16x8_t input, output;
>                      ^~~~~~
> ```

That's a good catch. I compiled for x86 only. Thus, this code is not compiled on my machine. I fix it anyways.

apps/volk_option_helpers.h
michaelld: works for me

jdemel:
As far as I can tell, all issues are resolved. @michaelld

jdemel:
@michaelld @drmpeg I rebased this PR on current master. Though, this rewrites all commits. 

I assume that the GCC5 CI test fails for this PR as well. That's smth we need to discuss in another issue. cf. #375

michaelld:
> If VOLK is compiled with `-Wshadow` a lot of shadowed variables are reported. This PR is an effort to reduce the number of those reports. It would be great to eliminate all such occurrences and also compile with more warnings enabled.

Yes!

michaelld (approved):
LGTM

michaelld:
Useful CI passes. Maintainer approval. Requests addressed. We can fix up more/other is/as necessary in another PR. Merging before this goes stale!

PR #353, opened 2020-02-23
volk: accurate exp kernel. rebased #12

jdemel:
A more accurate exp VOLK kernel than volk_32f_expfast_32f. Taken from code licensed with zlib.

This is a continuation of #12 . I just rebased this PR onto current master and added the GPL header.

### DIFF
--- a/kernels/volk/volk_32f_exp_32f.h
+++ b/kernels/volk/volk_32f_exp_32f.h
@@ -1,0 +1,298 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2015-2020 Free Software Foundation, Inc.
+ *
+ * This file is part of GNU Radio
+ *
+ * GNU Radio is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3, or (at your option)
+ * any later version.
+ *
+ * GNU Radio is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


kernels/volk/volk_32f_exp_32f.h
```diff
@@ -0,0 +1,298 @@
...
+
+#include <stdio.h>
+#include <math.h>
+#include <inttypes.h>
+
+#ifndef INCLUDED_volk_32f_exp_32f_a_H
+#define INCLUDED_volk_32f_exp_32f_a_H
+
+#ifdef LV_HAVE_SSE4_1
+#include <smmintrin.h>
```
rear1019: The code is actually SSE2. I don’t see any non-SSE2 instructions and just compiled using ``LV_HAVE_SSE2`` and including ``<emmintrin.h>`` instead of ``<smmintrin.h>``.

rear1019 (changes requested):
The code should be safe to compile with SSE2 instead of requiring SSE4.1.

kernels/volk/volk_32f_exp_32f.h
jdemel: That's a surprise, since I just resubmitted this PR and we already marked it as reviewed. Hmm...

jdemel:
@michaelld since this PR was already reviewed and is just a rebase. I'd say it is ready to be merged.

michaelld (approved):
LGTM

PR #360, opened 2020-02-27
Fix the broken index max kernels

### LINKED ISSUE #359: AVX2 index max kernels fail when vector length is odd
As noted in https://github.com/gnuradio/volk/issues/356#issuecomment-590150800 and https://github.com/gnuradio/volk/issues/356#issuecomment-590320253, the AVX2 index max kernels sometimes fail. I had a look at the code, and the reason is that they fail to process the last element of the vector when `num_points` is odd. There is special handling for trailing groups of 4 or 2 elements, but not 1.

argilo:
Fixes https://github.com/gnuradio/volk/issues/359.

The `32fc_index_max_32u` and `32fc_index_max_16u` kernels for `a_avx2` and `u_avx2` currently fail for vector lengths that are not a multiple of 8, due to several bugs in the calculations for the tail end of the vector.

To solve the problem, I've replaced the vectorized tail-end processing with ordinary loops, similar to what's done in the `32f_index_max` kernels.

I also did a bunch of cleanup in some separate commits.

I tested this code with vector lengths 0-32 as follows:

`for l in {0..32}; do for i in {1..50}; do apps/volk_profile -v $l -i 1 -R 32fc_index_max_16u | grep -B 1 fail; apps/volk_profile -v $l -i 1 -R 32fc_index_max_32u | grep -B 1 fail; done; done`

After these changes, all errors are gone.

### DIFF
--- a/kernels/volk/volk_32fc_index_max_16u.h
+++ b/kernels/volk/volk_32fc_index_max_16u.h
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2012, 2014 Free Software Foundation, Inc.
+ * Copyright 2012, 2014-2016, 2018-2020 Free Software Foundation, Inc.
  *
  * This file is part of GNU Radio
  *
@@ -36,8 +36,8 @@
  *
  * <b>Dispatcher Prototype</b>
  * \code
- * void volk_32fc_index_max_16u(uint16_t* target, lv_32fc_t* src0, uint32_t num_points)
- * \endcode
+ * void volk_32fc_index_max_16u(uint16_t* target, lv_32fc_t* src0, uint32_t
+ * num_points) \endcode
  *
  * \b Inputs
  * \li src0: The complex input vector.
@@ -89,33 +89,32 @@ static inline void
 volk_32fc_index_max_16u_a_avx2(uint16_t* target, lv_32fc_t* src0, uint32_t num_points)
 {
     num_points = (num_points > USHRT_MAX) ? USHRT_MAX : num_points;
-    // Branchless version, if we think it'll make a difference
-    // num_points = USHRT_MAX ^ ((num_points ^ USHRT_MAX) & -(num_points < USHRT_MAX));
-
     const uint32_t num_bytes = num_points * 8;
 
     union bit256 holderf;
     union bit256 holderi;
     float sq_dist = 0.0;
+    float max = 0.0;
+    uint16_t index = 0;
 
     union bit256 xmm5, xmm4;
     __m256 xmm1, xmm2, xmm3;
-    __m256i xmm8, xmm11, xmm12, xmmfive, xmmfour, xmm9, holder0, holder1, xmm10;
+    __m256i xmm8, xmm11, xmm12, xmm9, xmm10;
 
-    xmm5.int_vec = xmmfive = _mm256_setzero_si256();
-    xmm4.int_vec = xmmfour = _mm256_setzero_si256();
-    holderf.int_vec = holder0 = _mm256_setzero_si256();
-    holderi.int_vec = holder1 = _mm256_setzero_si256();
+    xmm5.int_vec = _mm256_setzero_si256();
+    xmm4.int_vec = _mm256_setzero_si256();
+    holderf.int_vec = _mm256_setzero_si256();
+    holderi.int_vec = _mm256_setzero_si256();
 
     int bound = num_bytes >> 6;
     int i = 0;
 
-    xmm8 = _mm256_set_epi32(7, 6, 5, 4, 3, 2, 1, 0);
-    xmm9 = _mm256_setzero_si256(); //=xmm8
+    xmm8 = _mm256_setr_epi32(0, 1, 2, 3, 4, 5, 6, 7);
+    xmm9 = _mm256_setzero_si256();
     xmm10 = _mm256_set1_epi32(8);
     xmm3 = _mm256_setzero_ps();
 
-    __m256i idx = _mm256_set_epi32(7, 6, 3, 2, 5, 4, 1, 0);
+    __m256i idx = _mm256_setr_epi32(0, 1, 4, 5, 2, 3, 6, 7);
     for (; i < bound; ++i) {
         xmm1 = _mm256_load_ps((float*)src0);
         xmm2 = _mm256_load_ps((float*)&src0[4]);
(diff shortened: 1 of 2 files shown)


argilo:
It may still be possible to squeeze a bit more performance out of the code that processes the tail end, but for now it seems better to have slow code that works than fast code that doesn't. :smile:

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -141,104 +138,33 @@ volk_32fc_index_max_16u_a_avx2(uint16_t* target, lv_32fc_t* src0,
...
 
+  for(i = bound * 8; i < num_points; i++, src0++)
+  {
+    float new_sq_dist = lv_creal(src0[0]) * lv_creal(src0[0]) + lv_cimag(src0[0]) * lv_cimag(src0[0]);
+    if(new_sq_dist > sq_dist)
+    {
+      target[0] = i;
+      sq_dist = new_sq_dist;
+    }
+  }
```
jdemel: I really prefer this solution over the old one. Since we generally aim for large vectors, this small performance bump that might be introduced here, is probably neglectable.

Besides, this code should be copypasta from the generic kernel. Could you just call the generic kernel on the tail instead?

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -263,21 +186,20 @@ volk_32fc_index_max_16u_a_sse3(uint16_t* target, lv_32fc_t* src0,
...
+  xmm5.int_vec = _mm_setzero_si128();
+  xmm4.int_vec = _mm_setzero_si128();
+  holderf.int_vec = _mm_setzero_si128();
+  holderi.int_vec = _mm_setzero_si128();
 
   int bound = num_bytes >> 5;
   int i = 0;
 
-  xmm8 = _mm_set_epi32(3, 2, 1, 0);//remember the crazy reverse order!
+  xmm8 = _mm_set_epi32(3, 2, 1, 0);
```
jdemel: There should be a `_mm_setr_epi32` intrinsic, which hides this reverse order insanity.

jdemel (changes requested):
I didn't comment on all code occasions but I assume it would be helpful to update the code as suggested.

Also, you can update the copyright year. `-2020`.

Besides, this PR is really difficult to review because it rewrites so much of these kernels. Still, I like the effort to make it more verbose.

kernels/volk/volk_32fc_index_max_16u.h
argilo: > Could you just call the generic kernel on the tail instead?

Not easily, because the kernel only returns the index and not the maximum value itself (which would then need to be recomputed to check whether it's larger than the beginning of the array). That's why I went with the copy & paste approach. The code is actually adapted from the `32f_index_max` AVX kernels, but I'll see if I can make it more like the `32fc_index_max` generic kernels.

kernels/volk/volk_32fc_index_max_16u.h
argilo: True, I'll add add another commit to change that.

argilo:
> Besides, this PR is really difficult to review because it rewrites so much of these kernels.

It may be easier to review the commits individually. I tried to keep the various changes separate to simplify review.

kernels/volk/volk_32fc_index_max_16u.h
jdemel: You could write a function that also returns the maximum value. Then just call that function in the generic kernel.

argilo:
I made the requested changes. The tail processing code should now be closer to the generic version, and I switched from `set` intrinsics to `setr` in a separate commit.

jdemel:
@argilo I just noticed, the copyright year is not updated yet. That's something that should definitely go into this PR.

argilo:
Oops, I forgot that. I won't be able to get to it until tonight.

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -141,104 +140,27 @@ volk_32fc_index_max_16u_a_avx2(uint16_t* target, lv_32fc_t* src0,
...
-  target[0] = (holderf.f[4] > sq_dist) ? holderi.i[4] : target[0];
-  sq_dist = (holderf.f[4] > sq_dist) ? holderf.f[4] : sq_dist;
-  target[0] = (holderf.f[5] > sq_dist) ? holderi.i[5] : target[0];
-  sq_dist = (holderf.f[5] > sq_dist) ? holderf.f[5] : sq_dist;
-  target[0] = (holderf.f[6] > sq_dist) ? holderi.i[6] : target[0];
-  sq_dist = (holderf.f[6] > sq_dist) ? holderf.f[6] : sq_dist;
-  target[0] = (holderf.f[7] > sq_dist) ? holderi.i[7] : target[0];
-  sq_dist = (holderf.f[7] > sq_dist) ? holderf.f[7] : sq_dist;
-
+    index = sq_dist > max ? i : index;
```
michaelld: Because there are 2 things being set, I think this would be more clear as:
```
if (sq_dist > max) {
  index = i;
  max = sq_dist;
}
```

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -492,77 +381,27 @@ volk_32fc_index_max_16u_u_avx2(uint16_t* target, lv_32fc_t* src0,
...
-      xmm11 = _mm256_and_si256(xmm8, xmm5.int_vec);
-      xmm12 = _mm256_and_si256(xmm9, xmm4.int_vec);
-
-      xmm9 = _mm256_add_epi32(xmm11, xmm12);
+  for(i = bound * 8; i < num_points; i++, src0++)
+  {
+    sq_dist = lv_creal(src0[0]) * lv_creal(src0[0]) + lv_cimag(src0[0]) * lv_cimag(src0[0]);
 
-      xmm8 = _mm256_add_epi32(xmm8, xmm10);
+    index = sq_dist > max ? i : index;
```
michaelld: ditto my comment above here

kernels/volk/volk_32fc_index_max_32u.h
```diff
@@ -131,83 +133,33 @@ volk_32fc_index_max_32u_a_avx2(uint32_t* target, lv_32fc_t* src0,
...
-    xmm11 = _mm256_and_si256(xmm8, xmm5.int_vec);
-    xmm12 = _mm256_and_si256(xmm9, xmm4.int_vec);
-
-    xmm9 = _mm256_add_epi32(xmm11, xmm12);
+  for(i = bound * 8; i < num_points; i++, src0++)
+  {
+    sq_dist = lv_creal(src0[0]) * lv_creal(src0[0]) + lv_cimag(src0[0]) * lv_cimag(src0[0]);
 
-    xmm8 = _mm256_add_epi32(xmm8, xmm10);
+    index = sq_dist > max ? i : index;
```
michaelld: ditto here

kernels/volk/volk_32fc_index_max_32u.h
```diff
@@ -445,76 +369,26 @@ volk_32fc_index_max_32u_u_avx2(uint32_t* target, lv_32fc_t* src0,
...
```
michaelld: ditto here

michaelld (changes requested):
LGTM overall; good to get these kernels fixed! Code reads well & makes sense. Consider my change request ... go with it or not; I think it makes cleans up the code just a touch, but I'm good either way really. Once the Copyright date is updated we can merge IMHO.

argilo:
I added a commit with copyright date changes (including the years that had changes according to `git log`) and another commit with the changes suggested by @michaelld. I also updated the generic kernels to match, since that code was copied from them. To make sure I didn't break anything, I first updated the AVX2 kernels, then ran the tests, then updated the generic kernels, and ran tests again.

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2012, 2014 Free Software Foundation, Inc.
+ * Copyright 2012, 2014-2016, 2018-2020 Free Software Foundation, Inc.
```
michaelld, line 3: LOL I'm getting sloppy and just moving to 2012-2020 ... I've seen other projects doing this, too, and it kinda makes sense ... Let's go with your version for now!

michaelld (approved):
LGTM; thx!

kernels/volk/volk_32fc_index_max_32u.h
```diff
@@ -385,12 +313,12 @@ static inline void
 #define INCLUDED_volk_32fc_index_max_32u_u_H
```
jdemel, line 305: I suggest to simplify this include guard. i.e. only one include guard for all kernel versions.

kernels/volk/volk_32fc_index_max_32u.h
```diff
@@ -445,76 +375,29 @@ volk_32fc_index_max_32u_u_avx2(uint32_t* target, lv_32fc_t* src0,
...
+  _mm256_storeu_si256(&(holderi.int_vec), xmm9);
 
-    xmm8 = _mm256_add_epi32(xmm8, xmm10);
+  for(i = 0; i < 8; i++)
+  {
+    if(holderf.f[i] > max)
+    {
+      index = holderi.i[i];
+      max = holderf.f[i];
+    }
```
jdemel: I'm curious about formatting here. I'd start braces on a new line for functions and at the end of a line for `for` and `if`. I thought that would be the clang-format way.

jdemel (approved):
LGTM. Thanks for your patience and your work!

kernels/volk/volk_32fc_index_max_32u.h
argilo: I rebased (to pick up the recent clang-format commit) and ran clang-format on this branch, so the formatting should now be fixed.

jdemel:
2 approvals, CI passes, merging.

PR #364, opened 2020-03-05
Updated log10 calcs to use faster log2 approach

### LINKED ISSUE #361: power_spectral_density kernel speedup
Log10 is a very slow calculation on a generic CPU.  There is a way to rewrite the log10 calculations to leverage volk's log2 kernel for a 300-400% performance speedup.  If you change the log10 calc over like this: n*log10(x) = n*log2(x)/log2(10) = (n/log2(10)) * log2(x), you can do the same math in 2 volk calls since there's a volk log2 by precomputing the n/log2(10) as log2To10Factor:
volk_32f_log2_32f(out,in,noi);
volk_32f_s32f_multiply_32f(out,out,log2To10Factor,noi);

There's a number of GR GUI controls calling the PSD kernel that would benefit from the performance increases, so this issue is a request to change the PSD kernel to take this approach instead of straight log10.

ghostop14:
Calculating n*log10(x) is faster as:
n*log2(x)/log2(10) = (n/log2(10)) * log2(x)
This update incorporates that change.

Testing before updates gave the following performance (3664 ms -> 270 ms):
RUN_VOLK_TESTS: volk_32fc_s32f_power_spectrum_32f(131071,1987)
a_sse3 completed in 3656.11 ms
generic completed in **3664 ms**
Best aligned arch: a_sse3
Best unaligned arch: generic

After updates:
RUN_VOLK_TESTS: volk_32fc_s32f_power_spectrum_32f(131071,1987)
a_sse3 completed in 2049.2 ms
generic completed in **270.026 ms**
Best aligned arch: generic
Best unaligned arch: generic


ghostop14:
Pushed another update to the spectral kernels to leverage the volk multiply kernel for the one CPU loop I had.  Now only a scalar add has to be done with a standard loop (doesn't look like there's a volk 32f scalar add kernel).  Squeezed a little more performance out of it.

ghostop14:
Looks like there's a build failure in qa_volk_32f_x3_sum_of_poly_32f, unrelated to these changes.

include/volk/volk_common.h
```diff
@@ -5,84 +5,86 @@
...
-#  define __VOLK_ATTR_ALIGNED(x) __declspec(align(x))
-#  define __VOLK_ATTR_UNUSED
-#  define __VOLK_ATTR_INLINE     __forceinline
-#  define __VOLK_ATTR_DEPRECATED __declspec(deprecated)
-#  define __VOLK_ATTR_EXPORT     __declspec(dllexport)
-#  define __VOLK_ATTR_IMPORT     __declspec(dllimport)
-#  define __VOLK_PREFETCH(addr)
-#  define __VOLK_ASM __asm
-#  define __VOLK_VOLATILE
+#define __VOLK_ATTR_ALIGNED(x) __declspec(align(x))
```
michaelld: did you run clang-format?

michaelld: I don't mind; but please make it a separate commit, so that we can easily review the critical changes.

kernels/volk/volk_32f_log2_32f.h
```diff
@@ -92,29 +94,23 @@
...
 
-static inline void
-volk_32f_log2_32f_generic(float* bVector, const float* aVector, unsigned int num_points)
-{
-  float* bPtr = bVector;
-  const float* aPtr = aVector;
+static inline void volk_32f_log2_32f_generic(float *bVector,
+                                             const float *aVector,
+                                             unsigned int num_points) {
+  float *bPtr = bVector;
```
michaelld: i don't know clang-format; like really at all. can we change it to do `float* foo;` rather than `float *foo;`? My preference is keeping the whole type all together.

michaelld (changes requested):
this is tough to review because of the clang-format changes. Please split them up & redo this PR to make our review doable easily. thx!

kernels/volk/volk_32fc_s32f_power_spectrum_32f.h
```diff
@@ -128,7 +128,10 @@ volk_32fc_s32f_power_spectrum_32f_a_sse3(float* logPowerOutput, const lv_32fc_t*
     const float real = *inputPtr++ * iNormalizationFactor;
     const float imag = *inputPtr++ * iNormalizationFactor;
 
-    *destPtr = 10.0*log10f(((real * real) + (imag * imag)) + 1e-20);
+    // *destPtr = 10.0*log10f(((real * real) + (imag * imag)) + 1e-20);
+    // For performance rewrite log10 to use log2
```
argilo: I don't think these comments are needed. (Likewise for the other places they appear below.)

kernels/volk/volk_32fc_s32f_power_spectrum_32f.h
```diff
@@ -128,7 +128,10 @@ volk_32fc_s32f_power_spectrum_32f_a_sse3(float* logPowerOutput, const lv_32fc_t*
...
+    *destPtr = volk_log2to10factor *
+               log2f_non_ieee(((real * real) + (imag * imag)) + 1e-20);
```
argilo: The `+ 1e-20` should no longer be necessary, because `log2f_non_ieee` already returns -127 instead of -Inf when the input is zero.

argilo: There are a bunch more places below where `+ 1e-20` is used and I think they can all be removed.

ghostop14:
New updates have been pushed.  Thanks to @argilo we eliminated a little more unnecessary code from the calculations.

argilo:
The entire `volk_32fc_s32f_x2_power_spectral_density_32f` implementation could be replaced by a single call to `volk_32fc_s32f_power_spectrum_32f(logPowerOutput, complexFFTInput, normalizationFactor * sqrt(rbw), num_points)`.

argilo:
Since power spectrum can be implemented as four calls to other volk kernels, it hardly seems worth keeping the sse3 and avx kernels, which are an order of magnitude slower.

argilo:
I've implemented the above two suggestions in this commit: https://github.com/argilo/volk/commit/a5a23f77b370e3989384302e21a5e9902031c6cb

ghostop14:
Before I merge the commit in, @michaelld, @jdemel: we were talking if it made sense to leave the other non-generic kernels for now to allow tests to run since it was such a big change, then in a future update remove them and consolidate the spectral calls to just an inline function?  Or would it make sense to do this within this PR?   Any suggestions?

michaelld:
My vote is to remove those kernels in a different commit.

michaelld:
Let's get this issue resolved & see where the speedups happen once it's all said and done. We can then decide what the correct path is moving forward.

ghostop14:
I think we should be good-to-go then.  All tests now passed and I incorporated the volk_32fc_s32f_x2_power_spectral_density_32f into this version that @argilo had in the commit.

ghostop14:
@jdemel , @michaelld , hope all is well with everything going on.  Any updates on these and the other PR's for me to address before merging?  Obviously have a little more spare time than usual in the evenings to clean up any issues given all the closures.

kernels/volk/volk_32fc_s32f_power_spectrum_32f.h
```diff
@@ -190,23 +193,25 @@ volk_32fc_s32f_power_spectrum_32f_generic(float* logPowerOutput, const lv_32fc_t
...
+  volk_32fc_magnitude_squared_32f(logPowerOutput, complexFFTInput, num_points);
+
+  // Finish ((real * real) + (imag * imag)) calculation:
+  volk_32f_s32f_multiply_32f(logPowerOutput, logPowerOutput, normFactSq, num_points);
+
+  // The following calculates 10*log10(x) = 10*log2(x)/log2(10) = (10/log2(10))
+  // * log2(x)
+  volk_32f_log2_32f(logPowerOutput, logPowerOutput, num_points);
+  volk_32f_s32f_multiply_32f(logPowerOutput, logPowerOutput, volk_log2to10factor,
+                             num_points);
```
jdemel: I like the "meta" approach here.

Since this is the generic kernel, I wonder if we should call the generic kernels here as well.

jdemel (commented):
LGTM. We should discuss what to do about my comment and then I'm all in to merge this PR.

kernels/volk/volk_32fc_s32f_power_spectrum_32f.h
ghostop14: We'd lose all the performance benefits if we forcibly call the generics.  I think the benefit here is that this generic calls the best-performing version of each of the called kernels.

jdemel: That's true. Also, a combination of different kernels for different parts might be beneficial in terms of performance. In this case, I wonder if we should just remove all the other kernels. Or go for a generic simple implemenentation and a generic optimized version that just calls into the other kernels.

ghostop14:
I thought about some options too, but this one's kind of a one-off since it's not a foundational kernel per se.  It just calls other kernels.  It's more of an inline wrapper than a kernel?  In that respect I don't know taking the time to write a pure generic to see the implementation would help.  Maybe we just add the implementation calculations back in to a comment to show the logic?

jdemel:
@ghostop14 do you mind to rebase your PR? That would be great.

ghostop14:
Rebased, fixed merge issues, clang-formatted my modified files, moved generic to top of file to align with other suggestions, and re-pushed.  Doesn't look like CI re-ran after force push but the conflicts are resolved in the repo.

michaelld (approved):
LGTM

michaelld:
@jdemel thoughts on this PR? I'm good with it.

jdemel:
I approve this PR too. Merging.

PR #366, opened 2020-03-11
Fixed Broken Rotator

### LINKED ISSUE #329: volk_32fc_s32fc_x2_rotator_32fc 10 dB noise in results

ghostop14:
Fixes #329 

The non-generic rotator blocks were using incorrect loop maximums
for the outer, inner, and remainder loops that were causing
calculation issues.  This commit corrects those issues and restores
the rotator accuracy to the default 1e-6 precision.


ghostop14:
Issue 329 can be linked to this PR.

michaelld:
> Issue 329 can be linked to this PR.

done

michaelld:
Relevant CI passes, which is always good! Have you tried specifically stressing the rotator? I think there's a discussion of how to do this somewhere in Volk github space ...

michaelld:
The CI error is with GCC7 in `qa_volk_16ic_x2_dot_prod_16ic` ... 1-off errors again!

ghostop14:
Yeah I noticed that.  These changes are nowhere near that code.  Had happened on a couple other PR's.  If it runs again sometimes it goes through.  I figured that wasn't on my end.

ghostop14:
@michaelld I did a quick search and I couldn't find any past discussion.  I had done something like that when I had originally opened the linked issue.  I did run updated tests with the new kernel in GR flowgraphs.  Some very targeted rotator comparisons against old results, as well as rotating a recorded signal and all seemed well.  There's still some noise that shows up if you set the frequency y-min to like -150 and just pump a signal source into the rotator->frequency sink, but it's way better than it was before.  I'm sure that has to do with the volk tolerances mostly being good only to 1e-6 (if you set it to 1e-7 in the qa_test for all kernels, multiple existing kernels start to fail, probably why it's set to 1e-6).

jdemel (approved):
LGTM. Thanks for taking the time to figure out the details.

ghostop14:
Rebased, fixed merge issues, clang-formatted my modified files, and re-pushed.  Doesn't look like CI re-ran after force push but the conflicts are resolved in the repo.

michaelld (approved):
LGTM; thx!

jdemel (approved):
LGTM. Thanks for figuring out the correct fix.

jdemel:
2 approvals. Merging.

PR #367, opened 2020-03-22
cmake: Remove the ORC from the VOLK public link interface

velichkov:
The ORC is an internal dependency that is used to generate SIMD
implementations of some the kernels and no ORC types or functions are
exposed by the VOLK library so adding it to the public link interface is
unnecessary when linking dynamically.

Currently the ORC is added to the INTERFACE_LINK_LIBRARIES property of
the Volk::volk target in VolkTargets.cmake and you need to have the ORC
development files (liborc-*-dev) installed on your system in order to
successfully link a program or library that uses VOLK.

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -465,7 +465,6 @@ if(ORC_FOUND)
     #setup orc library usage
     include_directories(${ORC_INCLUDE_DIRS})
     link_directories(${ORC_LIBRARY_DIRS})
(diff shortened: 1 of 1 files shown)


jdemel (approved):
LGTM

michaelld (approved):
LGTM; makes sense; thx!

michaelld:
relevant CI passes. 2 approvals. simple and useful fix. merging.

PR #368, opened 2020-03-24
32fc_index_max: Improve speed of AVX2 versions

rear1019:
The title says it all. Review carefully. Because, you know, human virtual machines for the execution of x86 and C code do fail sometimes :wink:.

[1] [``volk_profile`` results](https://gist.github.com/rear1019/90502ab693e8e6a0a19fc7e201526d98)

### DIFF
--- a/include/volk/volk_avx2_intrinsics.h
+++ b/include/volk/volk_avx2_intrinsics.h
@@ -119,4 +119,116 @@ static inline __m256 _mm256_scaled_norm_dist_ps_avx2(const __m256 symbols0,
     return _mm256_mul_ps(norms, scalar);
 }
 
+/*
+ * The function below vectorizes the inner loop of the following code:
+ *
+ * float max_values[8] = {0.f};
+ * unsigned max_indices[8] = {0};
+ * unsigned current_indices[8] = {0, 1, 2, 3, 4, 5, 6, 7};
+ * for (unsigned i = 0; i < num_points / 8; ++i) {
+ *     for (unsigned j = 0; j < 8; ++j) {
+ *         float abs_squared = real(src0) * real(src0) + imag(src0) * imag(src1)
+ *         bool compare = abs_squared > max_values[j];
+ *         max_values[j] = compare ? abs_squared : max_values[j];
+ *         max_indices[j] = compare ? current_indices[j] > max_indices[j]
+ *         current_indices[j] += 8; // update for next outer loop iteration
+ *         ++src0;
+ *     }
+ * }
+ */
+static inline void vector_32fc_index_max_variant0(__m256 in0,
+                                                  __m256 in1,
+                                                  __m256* max_values,
+                                                  __m256i* max_indices,
+                                                  __m256i* current_indices,
+                                                  __m256i indices_increment)
+{
+    in0 = _mm256_mul_ps(in0, in0);
+    in1 = _mm256_mul_ps(in1, in1);
+
+    /*
+     * Given the vectors a = (a_7, a_6, …, a_1, a_0) and b = (b_7, b_6, …, b_1, b_0)
+     * hadd_ps(a, b) computes
+     * (b_7 + b_6,
+     *  b_5 + b_4,
+     *  ---------
+     *  a_7 + b_6,
+     *  a_5 + a_4,
+     *  ---------
+     *  b_3 + b_2,
+     *  b_1 + b_0,
+     *  ---------
+     *  a_3 + a_2,
+     *  a_1 + a_0).
+     * The result is the squared absolute value of complex numbers at index
+     * offsets (7, 6, 3, 2, 5, 4, 1, 0). This must be the initial value of
+     * current_indices!
+     */
+    __m256 abs_squared = _mm256_hadd_ps(in0, in1);
+
+    /*
+     * Compare the recently computed squared absolute values with the
+     * previously determined maximum values. cmp_ps(a, b) determines
+     * a > b ? 0xFFFFFFFF for each element in the vectors =>
+     * compare_mask = abs_squared > max_values ? 0xFFFFFFFF : 0
+     *
+     * If either operand is NaN, 0 is returned as an “ordered” comparision is
+     * used => the blend operation will select the value from *max_values.
+     */
+    __m256 compare_mask = _mm256_cmp_ps(abs_squared, *max_values, _CMP_GT_OS);
+
+    /* Select maximum by blending. This is the only line which differs from variant1 */
+    *max_values = _mm256_blendv_ps(*max_values, abs_squared, compare_mask);
+
+    /*
+     * Updates indices: blendv_ps(a, b, mask) determines mask ? b : a for
+     * each element in the vectors =>
+     * max_indices = compare_mask ? current_indices : max_indices
+     *
+     * Note: The casting of data types is required to make the compiler happy
+     * and does not change values.
+     */
+    *max_indices =
+        _mm256_castps_si256(_mm256_blendv_ps(_mm256_castsi256_ps(*max_indices),
+                                             _mm256_castsi256_ps(*current_indices),
+                                             compare_mask));
+
+    /* compute indices of complex numbers which will be loaded in the next iteration */
+    *current_indices = _mm256_add_epi32(*current_indices, indices_increment);
+}
+
+/* See _variant0 for details */
+static inline void vector_32fc_index_max_variant1(__m256 in0,
+                                                  __m256 in1,
+                                                  __m256* max_values,
+                                                  __m256i* max_indices,
+                                                  __m256i* current_indices,
+                                                  __m256i indices_increment)
+{
+    in0 = _mm256_mul_ps(in0, in0);
+    in1 = _mm256_mul_ps(in1, in1);
+
+    __m256 abs_squared = _mm256_hadd_ps(in0, in1);
+    __m256 compare_mask = _mm256_cmp_ps(abs_squared, *max_values, _CMP_GT_OS);
+
+    /*
+     * This is the only line which differs from variant0. Using maxps instead of
+     * blendvps is faster on Intel CPUs (on the ones tested with).
+     *
+     * Note: The order of arguments matters if a NaN is encountered in which
+     * case the value of the second argument is selected. This is consistent
+     * with the “ordered” comparision and the blend operation: The comparision
+     * returns false if a NaN is encountered and the blend operation
+     * consequently selects the value from max_indices.
+     */
+    *max_values = _mm256_max_ps(abs_squared, *max_values);
+
+    *max_indices =
+        _mm256_castps_si256(_mm256_blendv_ps(_mm256_castsi256_ps(*max_indices),
+                                             _mm256_castsi256_ps(*current_indices),
+                                             compare_mask));
+
+    *current_indices = _mm256_add_epi32(*current_indices, indices_increment);
+}
+
 #endif /* INCLUDE_VOLK_VOLK_AVX2_INTRINSICS_H_ */
--- a/kernels/volk/volk_32fc_index_max_16u.h
+++ b/kernels/volk/volk_32fc_index_max_16u.h
@@ -84,82 +84,120 @@
 
 #ifdef LV_HAVE_AVX2
 #include <immintrin.h>
+#include <volk/volk_avx2_intrinsics.h>
 
-static inline void
-volk_32fc_index_max_16u_a_avx2(uint16_t* target, lv_32fc_t* src0, uint32_t num_points)
+static inline void volk_32fc_index_max_16u_a_avx2_variant_0(uint16_t* target,
+                                                            lv_32fc_t* src0,
+                                                            uint32_t num_points)
 {
     num_points = (num_points > USHRT_MAX) ? USHRT_MAX : num_points;
-    const uint32_t num_bytes = num_points * 8;
 
-    union bit256 holderf;
-    union bit256 holderi;
-    float sq_dist = 0.0;
-    float max = 0.0;
-    uint16_t index = 0;
+    const __m256i indices_increment = _mm256_set1_epi32(8);
+    /*
+     * At the start of each loop iteration current_indices holds the indices of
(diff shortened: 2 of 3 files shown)


jdemel:
The results seem to be correct. In this specific case though, could you add a comments that explain the idea behind the algorithm? Especially why is it not necessary to call permutevars anymore. I'm sure that'll be helpful at some point in the future.

jdemel (changes requested):
The results seem to be correct. In this specific case though, could you add a comments that explain the idea behind the algorithm? Especially why is it not necessary to call permutevars anymore. I'm sure that'll be helpful at some point in the future.

rear1019:
I’ve changed the algorithm, making the AVX2 version even faster – at least on a Ryzen CPU. Someone should check whether Intel CPUs benefit as well. I think they will looking at the assembly code and instruction tables. If so, I will update the `_16u` version, add comments to the code to explain the algorithm and squash the commits.

``permutevar`` removed
```
RUN_VOLK_TESTS: volk_32fc_index_max_32u(1024,1000000)
a_avx2 completed in 356.048 ms
a_sse3 completed in 390.086 ms
generic completed in 813.806 ms
u_avx2 completed in 424.176 ms

RUN_VOLK_TESTS: volk_32fc_index_max_32u(256,200000)
a_avx2 completed in 19.314 ms
a_sse3 completed in 22.4622 ms
generic completed in 43.357 ms
u_avx2 completed in 24.558 ms
```
new algorthm
```
RUN_VOLK_TESTS: volk_32fc_index_max_32u(1024,1000000)
a_avx2 completed in 275.812 ms
a_sse3 completed in 369.187 ms
generic completed in 806.999 ms
u_avx2 completed in 343.29 ms

RUN_VOLK_TESTS: volk_32fc_index_max_32u(256,200000)
a_avx2 completed in 15.6729 ms
a_sse3 completed in 21.5369 ms
generic completed in 44.497 ms
u_avx2 completed in 21.4834 ms
```

See [1] for more results.

[1] [`volk_profile` results](https://gist.github.com/rear1019/90502ab693e8e6a0a19fc7e201526d98)

EDIT: Fixes labels.

jdemel (changes requested):
I couldn't test the code on Intel. The changes are promising though.

Could you add comments to your code that explain it better?

Could you run `clang-format` on your changed files please. Otherwise, we can't merge this PR because we want to enforce consistent code formatting.

rear1019:
Rebased against current master, added comments and ran clang-format.

jdemel:
I run: `./apps/volk_profile -n -R fc_index_max_32 -i 10000000 -v 2048`

I obtain the following results on `AMD Ryzen 3970X`
branch `master`:
```
RUN_VOLK_TESTS: volk_32fc_index_max_32u(2048,10000000)
a_avx2 completed in 4022.68 ms
a_sse3 completed in 6189.35 ms
generic completed in 12595.8 ms
u_avx2 completed in 4697.48 ms
```
branch `improve_index_max_avx2`:
```
a_avx2 completed in 2534.15 ms
a_sse3 completed in 6353.07 ms
generic completed in 12940.8 ms
u_avx2 completed in 2775.02 ms
```
RESULT: pretty consistent imrpovements. That's great.


I obtain the following results on `Intel Xeon Gold 6254`
branch `master`:
```
RUN_VOLK_TESTS: volk_32fc_index_max_32u(2048,10000000)
a_avx2 completed in 5016.42 ms
a_sse3 completed in 7272.35 ms
generic completed in 22704.3 ms
u_avx2 completed in 5793.55 ms
```
branch `improve_index_max_avx2`:
```
a_avx2 completed in 6267.95 ms
a_sse3 completed in 7238.15 ms
generic completed in 23722.6 ms
u_avx2 completed in 6465.34 ms
```
RESULT: your changes seem to make things slower. Though, I'm not 100% certain because results vary quite wildly.

I obtain the following results on `Intel Core i7-7700T`
branch `master`:
```
RUN_VOLK_TESTS: volk_32fc_index_max_32u(2048,10000000)
a_avx2 completed in 4542.01 ms
a_sse3 completed in 8097.81 ms
generic completed in 23887.7 ms
u_avx2 completed in 5400.66 ms
```
branch `improve_index_max_avx2`:
```
a_avx2 completed in 6286.81 ms
a_sse3 completed in 8015.46 ms
generic completed in 23757.7 ms
u_avx2 completed in 6443.72 ms
```
RESULT: I suspect you improve speed on AMD but Intel is a different thing entirely.

A couple of bullet points we need to clarify
- Why is the `_u16` version so much faster? >2x (not posted here)
- Do the kernel still work reliably? We had an issue with incorrect results recently. cf. PR #360 and the corresponding issue #359 
- Where do the differences between Intel and AMD stem from?

rear1019:
> RESULT: your changes seem to make things slower [on Intel]. Though, I'm not 100% certain because results vary quite wildly.

Disable frequency scaling and Turbo Boost
`cpupower -c all frequency-set -g performance`
`echo 1 > /sys/devices/system/cpu/intel_pstate/no_turbo` (Intel CPU)
`echo 0 > /sys/devices/system/cpu/cpufreq/boost` (Anything else?)

~~There are numerous reports that Intel CPUs scale the frequency down if you make heavy use of AVX-512. But I don’t remember reading this applies to AVX2 as well.~~ (EDIT: This is not relevant here, as it would affect the old and new code.)

I might get access to an Intel machine with AVX2 support next week and check what’s going.

~~Regarding `_u16` version: It only considers the first `USHOT_MAX` elements which is smaller than the vector size you supplied.~~ EDIT2: The vector size is 2048

rear1019:
I can confirm that on a Intel CPU the new algorithm is slower than the old. Looking at the instruction latencies and throughputs I am rather surprised by this. On the other hand I am by no means an expert in this area.

In my tests the `_u16` was as fast as the `_u32` version.

jdemel (commented):
The code LGTM. It improves readability quite significantly. Unfortunately this is an Intel vs AMD performance enhancement. Either it runs faster on Intel or AMD. We need to discuss this issue.

rear1019:
I have rebased the branch and added a change which fixes the performance regression on Intel CPUs (at least on the one I tested with): The code is now faster than master (only slightly, if I remember correctly). This change does not affect performance gains on Ryzen. Please double-check this.

However, looking at the change I wonder whether `max_ps` and `cmp_ps` behave differently with regard to `NaN`… Moreover, I wonder whether the new and old code return the same result as the generic implementation if a `NaN` is encountered.

hcab14:
> RESULT: I suspect you improve speed on AMD but Intel is a different thing entirely.

If that's the case, why not keep both versions of the kernel and let `volk_profile` sort it out?

jdemel:
@hcab14 , this is, of course, the obvious solution. Thanks! At least I didn't think of it before. We have such constructs in other places as well. I can think of `spiral` kernels.

@rear1019 If you don't mind, it would be great to merge your PR with both kernel versions. Could you try to move common code to there own functions? Especially meta intrinsics. That should improve maintainability. I suggest smth like `amd` and `intel` name extensions. Please double check if this code works with `volk_profile` and our other infrastructure. I know this seems unnecessary now that you figured out a way to improve speed without trashing Intel CPU performance. But it wouldn't hurt to have both.

rear1019:
Sorry for the long absence :(

Code is rebased and contains the two variants. On a Ryzen CPU (Zen 1 architecture) both variants are equally fast: The numbers are too close to call a winner, though it looks like `variant_1` is a tiny bit faster than `variant_0` with “large” vector sizes.

`variant_1` is the one faster on Intel, however, I do not have access to an Intel machine with AVX2 to test again.

How does Volk select an implementation if there are multiple ones for the same architecture level (e.g. AVX2)? We should make sure that `variant_1` is preferred, just in case `volk_profile` has not been run (I personally don’t…)

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -86,84 +86,35 @@
...
-        xmm8 = _mm256_add_epi32(xmm8, xmm10);
-    }
-
-    _mm256_store_ps((float*)&(holderf.f), xmm3);
-    _mm256_store_si256(&(holderi.int_vec), xmm9);
+#define ALIGNED 1
+#define VARIANT 0
+#include "volk_32fc_index_max_avx2.tmpl"
+#undef ALIGNED
+#undef VARIANT
```
jdemel: I understand your attempt to minimize code. Though, in this instance, I'd be hesitant to merge it because this template magic adds quite a bit of complexity and cmake interaction.

Would it be possible to move most code into intrinsics in [volk_avx2_intrinsics.h](https://github.com/gnuradio/volk/blob/master/include/volk/volk_avx2_intrinsics.h)?

kernels/volk/volk_32fc_index_max_avx2.tmpl
```diff
@@ -0,0 +1,131 @@
...
+    /*
+     * Updates indices: blendv_ps(a, b, mask) determines mask ? b : a for
+     * each element in the vectors =>
+     * max_indices = compare_mask ? current_indices : max_indices
+     *
+     * Note: The casting of data types is required to make the compiler happy
+     * and does not change values.
+     */
+    max_indices = (__m256i)_mm256_blendv_ps(
+        (__m256)max_indices, (__m256)current_indices, compare_mask);
```
jdemel: At least MSVC doesn't understand these typecasts. I assume the correct call would be `_mm256_castps_si256` and the inverse.

jdemel (changes requested):
Thanks for working on your PR.

Would it be possible to create an intrinsic (or multiple) in `volk_avx2_intrinsics.h` and move most of the common code into these functions. Then use them in all 4 kernels? I know this might add some copypasta code for load/store etc. but the current solution really adds complexity. And I worry that this will break in the future and cause quite a bit of a maintenance burden.

Our CI complains about formatting and the Windows build is broken too. I hope these issues are easy to fix.

kernels/volk/volk_32fc_index_max_avx2.tmpl
rear1019: Fixed

kernels/volk/volk_32fc_index_max_16u.h
rear1019: I moved some common code into that file. There is still duplicate code after the main loop, but as of now I do not know how to deal with that: Putting it into the intrinsics file feels wrong and header files in `kernels/volk` are parsed by generator scripts (i.e. they must contain kernel code). We need a way to have private helper functions…

rear1019: PS: The latest changes are not tested, I do not have access to machine with AVX2 right now.

lib/CMakeLists.txt
```diff
@@ -318,12 +318,13 @@ message(STATUS "Available machines: ${available_machines}")
 file(GLOB xml_files ${PROJECT_SOURCE_DIR}/gen/*.xml)
 file(GLOB py_files ${PROJECT_SOURCE_DIR}/gen/*.py)
 file(GLOB h_files ${PROJECT_SOURCE_DIR}/kernels/volk/*.h)
+file(GLOB tmpl_files ${PROJECT_SOURCE_DIR}/kernels/volk/*.tmpl)
```
jdemel: I assume the changes in `lib/CMakeLists.txt` are obsolete now.

jdemel (commented):
Thanks for working on this PR! It has seen quite a lot of changes by now. Could you fix the formatting issues https://github.com/gnuradio/volk/pull/368/checks?check_run_id=1645069266 and I'd say this PR is ready to be merged.

lib/CMakeLists.txt
rear1019: Fixed

rear1019:
* Squashed commits
* Added some comments regarding NaNs

One last thing to decide: Do we really want to include both versions? Variant 1 is the one which should be used on Intel (at least with the CPUs tested so far) and AMD’s Ryzen (Zen 1) doesn’t care. Intel users not running `volk_profile` might see a performance regression depending on how `volk` selects kernels. As far as I can tell, the first defined will be used (currently Variant 0), but I haven’t tested this.

> Could you fix the formatting issues

It looks like the `clang-format` version used by CI is not smart enough to recognize that `for (; i<num_bytes>> 3; ++i)` does not contain a templates related code :wink:

jdemel:
I've seen clang-format hiccups in the past as well. I'd like to stick to them nonetheless and hope the software improves over time. I'd benefit from having both versions. I'd say let's keep them both. Since `volk_profile` exists exactly because we can't reliably tell which version to run, we need to rely on user interaction here. I assume it would be helpful to make this a more prominent feature.

jdemel (approved):
LGTM, Thanks for working on it.

include/volk/volk_avx2_intrinsics.h
```diff
@@ -119,4 +119,116 @@ static inline __m256 _mm256_scaled_norm_dist_ps_avx2(const __m256 symbols0,
     return _mm256_mul_ps(norms, scalar);
 }
 
+/*
+ * The function below below vectorizes the inner loop of the following code:
```
michaelld: double "below"; please remove one of them

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -84,86 +84,125 @@
...
+        ++src0;
     }
-    target[0] = index;
+
+    *target = index;
 }
 
 #endif /*LV_HAVE_AVX2*/
 
+
```
michaelld: Extra \n; please remove

kernels/volk/volk_32fc_index_max_16u.h
```diff
@@ -323,84 +362,123 @@ volk_32fc_index_max_16u_generic(uint16_t* target, lv_32fc_t* src0, uint32_t num_
...
```
michaelld: Extra \n; please remove

michaelld (changes requested):
3 minor tweaks & it's good to go IMHO

michaelld (approved):
LGTM ; thx!

PR #370, opened 2020-03-29
Drop compile-time CPU detection

### LINKED ISSUE #369: Build results vary from build machine CPU
Originally filed at https://bugzilla.opensuse.org/show_bug.cgi?id=1152001

While working on reproducible builds for openSUSE, I found that
our volk-2.2.1 package varied depending on build machine CPU type.

In our build environment, I can see that some files are only created on newer CPUs:

```diff
+++ /home/abuild/rpmbuild/BUILD/volk-2.2.1/build/CMakeFiles/CMakeRuleHashes.txt	2035-04-29 10:36:47.097658418 +0000
@@ -30,10 +30,6 @@
 10ab5f8cd40dc80f123094a691e3846e lib/volk_8i_convert_16i_a_orc_impl.c
 7fed82588c050353f4725686c4c4d55c lib/volk_8i_s32f_convert_32f_a_orc_impl.c
 ab188bfb1a0db04e8ac5dcdd16bd45ca lib/volk_cpu.c
-f8783a9360e4177fad6877c9981cf5b3 lib/volk_machine_avx2_64_mmx_orc.c
-0156eb25d341afbcd7da28bffb2211cf lib/volk_machine_avx512cd_64_mmx_orc.c
-010fd1209e2dead02ca1286493e4b93d lib/volk_machine_avx512f_64_mmx_orc.c
-35a4ba48334a08077b745897eee21beb lib/volk_machine_avx_64_mmx_orc.c
 48ce9369600175205bc1ae3d1398fd8e lib/volk_machine_generic_orc.c
 ea76d4812482f5afdf1b85fd0515cd3d lib/volk_machine_sse2_64_mmx_orc.c
```

Here is a partial diff between -cpu host and -cpu kvm64 builds
on a avx2 capable host

```diff
  -- Performing Test have_mavx512f - Success
  -- Performing Test have_mavx512cd
  -- Performing Test have_mavx512cd - Success
- -- CPU missing xgetbv, Overruled arch avx
- -- Available architectures: generic;64;3dnow;abm;popcount;mmx;fma;sse;sse2;orc;norc;sse3;ssse3;sse4_a;sse4_1;sse4_2;avx2;avx512f;avx512cd
+ -- Available architectures: generic;64;3dnow;abm;popcount;mmx;fma;sse;sse2;orc;norc;sse3;ssse3;sse4_a;sse4_1;sse4_2;avx;avx2;avx512f;avx512cd
+ -- Available machines: generic_orc;sse2_64_mmx_orc;sse3_64_mmx_orc;ssse3_64_mmx_orc;sse4_a_64_mmx_orc;sse4_1_64_mmx_orc;sse4_2_64_mmx_orc;avx_64_mmx_orc;avx2_64_mmx_orc;avx512f_64_mmx_orc;avx512cd_64_mmx_orc
+ -- BUILD INFO ::: avx_64_mmx_orc ::: GNU ::: -O2 -g -DNDEBUG -O2 -Wall -D_FORTIFY_SOURCE=2 -fstack-protector-strong -funwind-tables -fasynchronous-unwind-tables -fstack-clash-protection -Werror=return-type -flto=auto -DNDEBUG -Wall -m64 -mmmx -msse -msse2 -msse3 -mssse3 -msse4.1 -msse4.2 -mpopcnt -mavx
```

bmwiedemann:
Drop compile-time CPU detection
So that we can generate AVX binaries on non-AVX machines.

See https://reproducible-builds.org/ for why this matters.

Fixes #369

This PR was done while working on reproducible builds for openSUSE.


jdemel:
LGTM. CI passes. Archs are still present in CI. So this seems to be a good change. Does that enable reproducable builds for VOLK? Or are there other issues pending?

@bmwiedemann  Do you have a CLA?

michaelld:
I think for this small a change we don't actually need a CLA ... but, if you want to do more PRs then it would be wise to have one.

michaelld:
Duplicating my comment on the issue:

We added this configure execution check a long time ago because Volk didn't have the runtime capable of failing on a compiled kernel ... I mean, why enable a kernel if it won't execute on the target CPU?

I agree that it would be better to allow any compilable kernel to be compiled, then do runtime detection and disabling of kernels based on the target CPU. I don't honestly know that Volk can do this yet.

bmwiedemann:
@jdemel it is the only reproducibility issue I found for openSUSE, but Debian people might find others later, because they vary more build environment parameters (build user, path etc)

https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/volk.html mentions "records build flags" (CFLAGS). E.g. do one build with `-Werror=return-type -Wall` and one without - should produce identical binaries if it would not include those strings in libvolk.so.2.x

Edit: it also captures build path in html docs (<title> and somewhere else)

bmwiedemann:
For the CLA, I do not forsee doing more patches.

I can state that I did all of these changes myself, my employer (SUSE) is totally fine with me contributing to open-source projects and therefore I declare my work CC0 licensed that allows FSF to claim copyright.

jdemel:
We know that we can merge this PR. Further, CI passes. Also, it gives us reproducible builds which we want too.

But before we can merge this PR, we need to figure out if this would break things for some users.

VOLK should dynamically detect CPU capabilities. Also, only the AVX kernel is not compiled. Even newer ones, e.g. AVX2, are.

@michaelld what are the exact circumstances under which VOLK fails? Why was this decision made?

michaelld:
> @michaelld what are the exact circumstances under which VOLK fails? Why was this decision made?

I honestly don't recall the -exact- circumstances, but here's the general memory:
{{{
Some now-older CPUs don't properly implement the `xgetbv` instruction (so, partial AVX, but not full AVX), but the compiler doesn't know this. If the compiler supports AVX, then it will compile AVX code that includes the `xgetbv` instruction -- regardless of whether it will execute. Executing this code on a CPU with the buggy `xgetbv` results in some sort of runtime error.

Thus, if the target CPU can't execute basic `xgetbv` code then we know to eliminate AVX as a supported SIMD kernel.
}}}

The above works really well when compiling from source for a specific target CPU.

But if we don't take the CPU into account -- just what the compiler supports -- the actual execution check should be removed. Other code that should be removed is (at least) < https://github.com/gnuradio/volk/blob/de93c2734821eefcd0de06d02e5291bb81b21472/lib/CMakeLists.txt#L179-L181 > ... and maybe others ... anything that is actually executed by CMake to check for whether the target CPU can execute the target instruction.

I'm not against this change. But I do want to make sure the Volk runtime will properly handle kernel failures such as what this specific PR's current code targets (the `xgetbv` instruction). I don't know if it does.

michaelld:
> Other code that should be removed is (at least) < https://github.com/gnuradio/volk/blob/de93c2734821eefcd0de06d02e5291bb81b21472/lib/CMakeLists.txt#L179-L181
> 
> ... and maybe others ... anything that is actually executed by CMake to check for whether the target CPU can execute the target instruction.

A quick code review ... looks like this is the only other such CPU execution check. Thinking it would be wise to add removing it to this PR ... to be complete.

velichkov (approved):
I just tested this patch successfully on an old server with a CPU that does not have AVX instructions (only SSE4.2). Executing `test_xgetbv` result in a `SIGILL`.
```
$gdb --ex run --args ./lib/test_xgetbv

Program received signal SIGILL, Illegal instruction.
0x0000000000401112 in _xgetbv (index=0) at /root/volk/build/lib/test_xgetbv.c:2
2	 unsigned long long _xgetbv(unsigned int index) { unsigned int eax, edx; __VOLK_ASM __VOLK_VOLATILE("xgetbv" : "=a"(eax), "=d"(edx) : "c"(index)); return ((unsigned long long)edx << 32) | eax; } int main (void) { (void) _xgetbv(0); return (0); }
(gdb) bt
#0  0x0000000000401112 in _xgetbv (index=0) at /root/volk/build/lib/test_xgetbv.c:2
#1  0x000000000040113b in main () at /root/volk/build/lib/test_xgetbv.c:2
(gdb) disassemble 
Dump of assembler code for function _xgetbv:
   0x0000000000401106 <+0>:	push   %rbp
   0x0000000000401107 <+1>:	mov    %rsp,%rbp
   0x000000000040110a <+4>:	mov    %edi,-0x14(%rbp)
   0x000000000040110d <+7>:	mov    -0x14(%rbp),%eax
   0x0000000000401110 <+10>:	mov    %eax,%ecx
=> 0x0000000000401112 <+12>:	xgetbv 
   0x0000000000401115 <+15>:	mov    %eax,-0x4(%rbp)
```
The runtime detection is implemented here
https://github.com/gnuradio/volk/blob/faf230e643e4910c884ccfbdcc27de8357b37c0a/tmpl/volk.tmpl.c#L47-L55
and `get_machine()` function is called in the various `__init*` function in `build/volk/lib/volk.c`.
```
(gdb) bt
#0  0x00007ffff74b665e in get_machine () at /home/vasko/gr38/src/gnuradio38/build/volk/lib/volk.c:7740
#1  __init_volk_32fc_s32fc_x2_rotator_32fc () at /home/vasko/gr38/src/gnuradio38/build/volk/lib/volk.c:7740
#2  0x00007ffff74b6919 in __volk_32fc_s32fc_x2_rotator_32fc (outVector=0x7ffff4eef010, inVector=0x7ffff4fb3010, phase_inc=<optimized out>, 
    phase=0x7fffffffb8c0, num_points=100000) at /home/vasko/gr38/src/gnuradio38/build/volk/lib/volk.c:7770
#3  0x0000000000402f4a in gr::blocks::rotator::rotateN (n=100000, in=0x7ffff4fb3010, out=0x7ffff4eef010, this=0x7fffffffb8c0)
    at /home/vasko/gr38/src/gnuradio38/gr-blocks/lib/../include/gnuradio/blocks/rotator.h:63
```
after applying this patch `volk-config-info` gives the following results
```
$ ./apps/volk-config-info --all-machines
generic;sse2_64_mmx;sse3_64_mmx;ssse3_64_mmx;sse4_a_64_mmx;sse4_1_64_mmx;sse4_2_64_mmx;avx_64_mmx;avx2_64_mmx;avx512f_64_mmx;avx512cd_64_mmx
$ ./apps/volk-config-info --avail-machines
generic;sse2_64_mmx;sse3_64_mmx;ssse3_64_mmx;sse4_1_64_mmx;sse4_2_64_mmx;
$ ./apps/volk-config-info --machine
sse4_2_64_mmx
```
so all x86_64 kernels are built, the best one gets selected, all test pass and `volk_profile` does finish successfully. 

And here is some more info about the CPU.
```
$cat /proc/cpuinfo 
processor	: 0
vendor_id	: GenuineIntel
cpu family	: 6
model		: 44
model name	: Intel(R) Xeon(R) CPU           E5620  @ 2.40GHz
stepping	: 2
microcode	: 0x1f
cpu MHz		: 2400.000
cache size	: 12288 KB
physical id	: 1
siblings	: 8
core id		: 0
cpu cores	: 4
apicid		: 32
initial apicid	: 32
fpu		: yes
fpu_exception	: yes
cpuid level	: 11
wp		: yes
flags		: fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc aperfmperf eagerfpu pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 cx16 xtpr pdcm pcid dca sse4_1 sse4_2 popcnt aes lahf_lm epb ssbd ibrs ibpb stibp tpr_shadow vnmi flexpriority ept vpid dtherm ida arat spec_ctrl intel_stibp flush_l1d
```

michaelld (approved):
LGTM ; makes sense too!

michaelld:
@velichkov thanks for that info ... very informative! I need to go collect my very old, pre-AVX laptop ... would help in verifying changes such as this!

michaelld:
@bmwiedemann do you want to add removing this to this PR too? It's similar to the current CPU test being removed already ... so thinking yes ...

https://github.com/gnuradio/volk/blob/de93c2734821eefcd0de06d02e5291bb81b21472/lib/CMakeLists.txt#L179-L181

bmwiedemann:
> do you want to add removing this to this PR too? It's similar to the current CPU test being removed already ... so thinking yes ...

I added this removal in a 2nd commit

jdemel:
PR still LGTM. Travis fails with `sum_of_poly` on ARM. It'd be helpful to be able to rerun jobs in such situations.

michaelld:
I've restarted CI / Travis. Guessing most will pass, some might not but I don't think it's because of this PR.

michaelld:
CI now passes. Multiple approvals and a good use case for why to remove this code. Merging.

PR #371, opened 2020-04-02
Fix endif macro name in comment

michaelld:
simple & quick tweak!


jdemel:
LGTM. Only comments change, merging.

This whole `simdmath.h` usage is suspicious though. We need to investigate its proper usage.

PR #372, opened 2020-04-08
fix: Use unaligned load

jdemel:
In the `multiply_conjugate_add` aligned AVX kernel, there was an aligned
load on a potentially unaligned vector. This tends to fail randomly.
This commit fixes the issue.


rear1019 (approved):
`set1_pd`, `load1_pd` or `broadcast_sd` (including some cast magic) could be used instead of the temporary array `scalar_v` + `loadu`. But I have not checked whether this actually works.

jdemel:
@rear1019 an alternative solution might be feasible but this is not the intent of this PR. I just want to make sure these functions never SegFault. A behavior I've seen quite a few times on Ryzen in particular.

michaelld (approved):
LGTM ... good catch!

michaelld:
Relevant CI passes. Maintainer approval. Merging.

PR #376, opened 2020-04-10
ci: Add Github CI Action

### LINKED ISSUE #336: Update CI infrastructure
Currently we have 3 CI systems that cover our codebase partially. And a fourth one to consider.
I have a list of suggestions for updates and issues that we should address. Additions and comments are welcome.

- Travis is probably the one we rely on the most. It needs to be updated! 
     - `distro: bionic` should be the baseline instead of `trusty`
     - All Jobs should have names to make them easier to distinguish
     - All jobs should be run on amd64 and aarch64 and maybe aarch32. That would probably imply to merge some jobs into one.
     - Use Python 3 only (remove six)
     - Include Boost only for old systems e.g. old enough GCC/Clang which do not provide `std::filesystem` or `experimental`
- AppVeyor to check if MSVC is broken
     - Needs an update to use Python3. 
     - Remove obsolete packages (e.g. Cheetah)
- BuildBot, the GR worker
     - Needs to run reliably.
     - Needs an update to use newer Ubuntu versions etc.
- GitHub Actions
     - Since we are on GitHub, we should use it as well to test at least on build

jdemel:
This commit adds CI tests for GitHub Actions for Linux, Windows and MacOS.

Fixes #336 

Since we use GitHub for code hosting, it is worthwhile to have CI tests that run in every VOLK repo on push. 

Be aware that Windows and MacOS CI does only run generic kernels at the moment because it cannot find `xgetbv`.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -1,0 +1,50 @@
+name: Run VOLK tests
+
+on: [push, pull_request]
+
+jobs:
+  build-ubuntu:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: dependencies
+      run: sudo apt install python3-mako liborc-dev
+    - name: configure
+      run: mkdir build && cd build && cmake -DCMAKE_CXX_FLAGS="-Werror" ..
+    - name: build
+      run: cmake --build build
+    - name: test
+      run: cd build && ctest -V
+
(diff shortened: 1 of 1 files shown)


michaelld (approved):
LGTM

michaelld:
I'm going to go ahead and merge this; we can fix it up down the road if/as need be. Better a little more CI that works rather than CI that doesn't!

PR #379, opened 2020-05-02
readme: Add section on supported platforms

jdemel:
So far, this is kind of a work in progress. I'd like to discuss the
exact wording here.
We run CI on Ubuntu, Windows and MacOS. For Windows and MacOS, we focus
on the latest release and do not make any attempt to support older
releases.
For Linux, we want to have support for distros other than Ubuntu and
older distros to some degree. Though, we decided to drop support for
Ubuntu 16.04 because some bugs make it hard work put a work around in
place for this old distro.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,6 @@ $ volk_profile
 ```
 
 ## Supported platforms
-
 VOLK aims to be portable to as many platforms as possible. We can only run tests on some platforms.
 
 ### Hardware architectures
@@ -65,11 +64,11 @@ Currently VOLK aims to run with optimized kernels on x86 with SSE/AVX and ARM wi
 
 ### OS / Distro
 We run tests on a variety of Ubuntu versions and aim to support as many current distros as possible.
-The same goal applies to differenct OSes. Although this does only happen rarely, it might occur that VOLK does not work on obsolete distros, e.g. Ubuntu 12.04.
+The same goal applies to different OSes. Although this does only happen rarely, it might occur that VOLK does not work on obsolete distros, e.g. Ubuntu 12.04.
 
 ### Compilers
-We want to make sure VOLK works with C/C++ standard compliant compilers. Most notably GCC and Clang.
-Other compilers that fail to be standard compliant, e.g. MSVC or AppleClang, do also receive attention but some functionality might be missing with these compilers due to lack of standard support.
+We want to make sure VOLK works with C/C++ standard compliant compilers. Of course, as an open source project we focus on open source compilers, most notably GCC and Clang.
+We work want to make sure VOLK compiles on a wide variety of compilers. Thus, we target AppleClang and MSVC as well. Mind that MSVC lacks `aligned_alloc` support for aligned arrays. We use MSVC specific instructions in this case which cannot be `free`'d with `free`.
 
 
 ## License
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ The same goal applies to different OSes. Although this does only happen rarely, 
 
 ### Compilers
 We want to make sure VOLK works with C/C++ standard compliant compilers. Of course, as an open source project we focus on open source compilers, most notably GCC and Clang.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,6 +56,22 @@ $ sudo make install
...
+### Hardware architectures
+Currently VOLK aims to run with optimized kernels on x86 with SSE/AVX and ARM with NEON.
+
+### OS / Distro
+We run tests on a variety of Ubuntu versions and aim to support as many current distros as possible.
+The same goal applies to differenct OSes. Although this does only happen rarely, it might occur that VOLK does not work on obsolete distros, e.g. Ubuntu 12.04.
+
+### Compilers
+We want to make sure VOLK works with C/C++ standard compliant compilers. Most notably GCC and Clang.
+Other compilers that fail to be standard compliant, e.g. MSVC or AppleClang, do also receive attention but some functionality might be missing with these compilers due to lack of standard support.
```
michaelld: VOLK builds using modern AppleClang, which correctly reports SIMD support. AppleClang is also C/C++ standards complaint.

MSVC is standards compliant, too, just as much as GCC or *Clang.

Most compilers provide non-standard features, e.g., `gnu++` for GCC, which is only partly supported on Clang or MSVC -- no guarantees!

I'd say to rewrite these 2 sentences.

README.md
```diff
@@ -54,6 +56,22 @@ $ sudo make install
 $ volk_profile
 ```
 
+## Supported platforms
+
```
michaelld: Extra CR/LF, please remove

michaelld (changes requested):
Good start!

README.md
jdemel: Sorry, only LF available. =D

README.md
jdemel: It would be great to tell our CI to use modern AppleClang then and build with proper SIMD support. Anyways, I'm happy to hear the situation is better than I thought according to CI.

I have [compiler support](https://en.cppreference.com/w/cpp/compiler_support) in mind whenever I refer to "standard compliant". Though, this is missing C language features. And I couldn't find such a table for C.
Anyways, I'm still very much disappointed of the lack of support for `aligned_alloc` in MSVC. Of course this frustrating experience motivated this sentence. 
I can rewrite this section to stress that we want to make sure, VOLK works best with open source compilers. What's the status of AppleClang? Are Apples' changes also open source just like Clang? For MSVC, I'd like to point out that it lacks `aligned_alloc` support but works otherwise.

jdemel: CMake can't find `xgetbv` on AppleClang and thus only generic kernels are compiled in our CI. A quick search revealed some discussion around that exact issue from 2013 by @michaelld . How did the situation evolve since then?

README.md
```diff
@@ -57,19 +57,18 @@ $ volk_profile
...
 ### OS / Distro
 We run tests on a variety of Ubuntu versions and aim to support as many current distros as possible.
-The same goal applies to differenct OSes. Although this does only happen rarely, it might occur that VOLK does not work on obsolete distros, e.g. Ubuntu 12.04.
+The same goal applies to different OSes. Although this does only happen rarely, it might occur that VOLK does not work on obsolete distros, e.g. Ubuntu 12.04.
 
 ### Compilers
-We want to make sure VOLK works with C/C++ standard compliant compilers. Most notably GCC and Clang.
-Other compilers that fail to be standard compliant, e.g. MSVC or AppleClang, do also receive attention but some functionality might be missing with these compilers due to lack of standard support.
+We want to make sure VOLK works with C/C++ standard compliant compilers. Of course, as an open source project we focus on open source compilers, most notably GCC and Clang.
+We work want to make sure VOLK compiles on a wide variety of compilers. Thus, we target AppleClang and MSVC as well. Mind that MSVC lacks `aligned_alloc` support for aligned arrays. We use MSVC specific instructions in this case which cannot be `free`'d with `free`.
```
michaelld: `work want` -> just `want`

michaelld (approved):
one change then go ahead & merge.

jdemel:
Approved and wording is fixed. Merging.

PR #391, opened 2020-06-15
fix: Adjust min/max for 32f_s32f_convert_8i kernel

### LINKED ISSUE #390: volk_32f_s32f_convert_8i behaviour differs between arm and x64
As detailed in https://github.com/gnuradio/gnuradio/issues/3571 it appears that `volk_32f_s32f_convert_8i` does not work as expected on ARM. The behaviour differs from that experienced in x86_64, and appears to unintentionally truncate negative values to zero.

jdemel:
This kernel used `CHAR_MIN` to determine the minimum value which may be
0. It is expected that this value is -127. Thus, we move to `INT8_MIN`.

Fix #390

### DIFF
--- a/kernels/volk/volk_32f_s32f_convert_8i.h
+++ b/kernels/volk/volk_32f_s32f_convert_8i.h
@@ -104,8 +104,8 @@ static inline void volk_32f_s32f_convert_8i_u_avx2(int8_t* outputVector,
     const float* inputVectorPtr = (const float*)inputVector;
     int8_t* outputVectorPtr = outputVector;
 
-    float min_val = CHAR_MIN;
-    float max_val = CHAR_MAX;
+    float min_val = INT8_MIN;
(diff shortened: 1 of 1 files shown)


drmpeg:
Tested on armv7 and works good. I also verified that CHAR_MIN = 0 and CHAR_MAX = 255 since char is unsigned on ARM.

For consistency (even though they'll never execute on ARM), should the other instances of CHAR_MIN and CHAR_MAX in the module be changed to INT8_MIN and INT8_MAX?

jdemel:
I'd agree that they need to be updated too. Thanks for pointing this out.

michaelld (approved):
Please change the comment on line 28 : "char" -> "int". After that, LGTM & let's merge this.

michaelld:
Relevant CI passes. Maintainer approval. Important fix. Merging.

PR #393, opened 2020-06-27
cpu_features: Switch to cpu_features

### LINKED ISSUE #311: Check Arch overrides on Windows
Check if all arch overrides are still valid for windows. E.g. AVX should be available on Appveyor, but it's ruled out.

### LINKED ISSUE #348: Make sure checks for AVX(2) and AVX-512 are correct
Here is how you check for avx2 https://github.com/webmproject/libvpx/blob/e2ade251b1231f59f3de3cd79ffbc01b2b2fec51/vpx_ports/x86.h#L213

```
if (max_cpuid_val >= 7) {
        /* Get the leaf 7 feature flags. Needed to check for AVX2 support */
        cpuid(7, 0, reg_eax, reg_ebx, reg_ecx, reg_edx);

        if (reg_ebx & BIT(5)) flags |= HAS_AVX2;
```

besides 

```
if ((reg_ecx & (BIT(27) | BIT(28))) == (BIT(27) | BIT(28))) {
    // Check for OS-support of YMM state. Necessary for AVX and AVX2.
    if ((xgetbv() & 0x6) == 0x6) {
      flags |= HAS_AVX;
```

https://github.com/gnuradio/volk/blob/1ec627c153a9625b588eef394c70af42811c53ab/tmpl/volk_cpu.tmpl.c#L114
It indeed looks like if `((xgetbv() & 0x6) == 0x6` should be done but you also check for ebx after cpuid 
And also look here, avx512 support is right or not... https://github.com/webmproject/libvpx/commit/602d25ec4344f96031ce1043c15490e79438c1bc

_Originally posted by @ValZapod in https://github.com/gnuradio/volk/issues/327#issuecomment-573787182_

### LINKED ISSUE #401: macOS volk_profile reports many "no architectures to test" and "generic"
Hi folks. I'm getting a lot of `no architectures to test` and `generic` in the `volk_profile` report generated. This is on macOS (AMD64 arch) built from source for volk tag 2.3.0. This is strange. I can do more debugging this week, but curious if folks have seen this before. Shall I explicitly declare arch?

Output: https://pastebin.com/f7sEXrRD

cc: https://github.com/Homebrew/homebrew-core/issues/59921

jdemel:
We used our own CPU feature detection methods. These were not totally
reliable and difficult to implement across compilers and OSes. We switch
to `cpu_features` as a subproject and build upon this expertise.

So far this is a work-in-progress. Known issues:
   - <s>code needs clean-up</s>
   - <s>remove unnecessary checks</s>
   - <s>make sure all NEON versions are still detected correctly</s> - CI indicates correct detection.
   - <s>incomplete `VOLK_CPU` feature checks:</s>
	- <s>`has_64`: check 64bit CPU.</s> We compile for one or the other. Not for both. No online check necessary. 
	- <s>`has_3dnow`</s> - FIXED: removed because it is unused
	- <s>`has_abm`</s> - FIXED: removed because it is unused
	- <s>`has_fma` now includes `fma3` and `fma4`.</s> - FIXED `fma4` removed because it is unused and obsolete
   - <s>Remove has `xgetbv` checks. We don't handle this anymore.</s> - Obsolete checks removed
   - <s>Make sure we don't break ABI.</s>
   - <s>AVX2 machine unavailable on Ryzen</s> - FIXED

Since this is a pretty invasive change, I'd like to discuss this PR and get feedback.

Fix #348 
Fix #311 
Fix #401


jdemel:
I updated this PR quite a lot. In order to proceed I'd really appreciate some feedback. I updated my original PR message with some additional info and worked out issues.

There are some issues, that need to be discussed:
- It does not compile with MSVC on Windows at the moment. This seems to be a configuration issue that  someone with said build environment needs to work out.
- Older compilers/linkers tend to fail. e.g. GCC4.8 or Clang 6 complain about missing -fPIC for cpu_features. It does work with Clang 10 and GCC7+ at least. Also, it seems to work well on ARM.
- QEMU seems to be problematic at the moment. CI fails.

ValZapod:
Where it compiles can you check with perf or some other profilers with check for vector engine load whether there is any difference in AVX2/512 code path?

noc0lour:
Hm, why is this not run on travis?

jdemel:
It seems the trigger is not reliable. But most previous commits are run with Travis.

noc0lour:
Weird, since travis is so tightly integrated with GitHub.

jdemel:
I assume those Travis issues stem from the fact that Travis is triggered via an old GitHub API and we should actually move to a new API.

lib/CMakeLists.txt
```diff
@@ -552,6 +552,9 @@ target_include_directories(volk
     PUBLIC $<INSTALL_INTERFACE:include>
 )
 
+include_directories(${PROJECT_SOURCE_DIR}/cpu_features/include)
```
ryanvolz: More of a style comment, but I think it is preferable from a CMake perspective to remove this and replace it with
```
target_link_libraries(volk_obj PRIVATE cpu_features)
```
above right after the `volk_obj` target is defined. That will not only set the include directory properly for `cpu_features` but also do anything else that might be needed in the future if there are changes to `cpu_features`

tmpl/volk_cpu.tmpl.c
```diff
@@ -24,189 +24,33 @@
 #include <stdlib.h>
 #include <string.h>
```
ryanvolz, line 26: Need
```
#if defined(__cplusplus)
using namespace cpu_features;
#endif
```
here to get this compiling with MSVC, where we are compiling the source as C++ to accommodate missing `complex.h` with C and MSVC. When the `cpu_features` headers are included from C++, they put everything in the `cpu_features` namespace.

ryanvolz (commented):
@jdemel asked me to take a look since I have experience fighting with MSVC and at least have access to a Windows VM. With a little digging, I have been able to get it compiling locally.

Here's the problem that I found. With MSVC, the volk_sources are compiled as C++ due to the lack of complex.h. So when cpu_features_macros.h gets included, it has a section starting on line 102 that detects the C++ compiler and puts the declarations in the "cpu_features" namespace. Thus "GetX86Info" is undefined, as it is instead known as "cpu_features::GetX86Info". I commented on the changes with my suggested solution, which is to use an `#if` to use that namespace when compiling as C++.

jdemel:
Thanks @ryanvolz !
Now, CI passes for Windows too.

[AppVeyor VS 12 2013 / python 3.4](https://ci.appveyor.com/project/gnuradio/volk/builds/34713299/job/fdrirq32gjigacs0) fails though. This is a really old version. I suggest to drop support for this old version. Just like my suggestion to drop support for GCC 4.8. We're looking at compilers from 2012/2013 here.

About Travis: @noc0lour, @n-west, @michaelld I'd like to update Travis integration and was successful with my fork. Though, it seems like I don't have permission to fix this in our official repo yet. I'd be great if you could help me out here.

noc0lour:
@jdemel  I clicked through the new travis-ci stuff and added it as a integration to the GNU Radio github org & migrated the volk repository to the "new" travis-ci.com site/functionality. Is that what you were refering to?

jdemel:
@noc0lour thanks! That's exactly what I was trying to do.

ValZapod:
> [AppVeyor VS 12 2013 / python 3.4](https://ci.appveyor.com/project/gnuradio/volk/builds/34713299/job/fdrirq32gjigacs0) fails though

Looks like there is no any error there, 😃. Ha. It just stops. You need to close and reopen pull request to repeat the build.
Is this
>Also, AppVeyor tells us that MSVC doesn't find any archs

Fixed now?

jdemel:
These are the last lines on the console:

> C:\projects\volk\cpu_features\src\string_view.c(174): error C2440: 'initializing' : cannot convert from 'int' to 'const StringView' [C:\projects\volk\cpu_features\utils.vcxproj]
>   Generating Code...
> Command exited with code 1

Thus, I assume it is still broken. It looks very specific as well. Of course, we need to figure out how to deal with this error.

I want to make sure, this PR works for all toolchains that we target. At the moment I worry about old compilers and QEMU. I hope we'll be able to fix this soon. Afterwards, this PR needs some clean-up. Refactoring etc. I'll try to remove unnecessary checks...

ryanvolz:
MSVC 2013 [doesn't have an `inline` keyword](https://stackoverflow.com/questions/24736304/unable-to-use-inline-in-declaration-get-error-c2054) and that causes the bulk of the errors with that AppVeyor build. It seems that `cpu_features` targets at least MSVC 2015, although there may be workarounds (e.g. `-Dinline=__inline`).

There's also this error:
```
C:\projects\volk\cpu_features\src\filesystem.c(29): error C2065: '_SH_DENYWR' : undeclared identifier
```

which indicates that the [`share.h` header is needed](https://docs.microsoft.com/en-us/cpp/c-runtime-library/sharing-constants), although I can't find documentation for MSVC 2013 to verify and figure out why it's different from 2015+.

lib/CMakeLists.txt
```diff
@@ -528,6 +528,8 @@ target_include_directories(volk_obj
     PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}
 )
 
+target_link_libraries(volk_obj PRIVATE cpu_features)
```
ryanvolz: Sorry for this suggestion, it is apparently only valid with CMake 3.12+, otherwise you get this error as seen at least with the Ubuntu Trusty build which presumably uses an older CMake:
```
CMake Error at lib/CMakeLists.txt:531 (target_link_libraries):

  Object library target "volk_obj" may not link to anything.
```

Instead, this should work:
```
target_include_directories(volk_obj PRIVATE $<TARGET_PROPERTY:cpu_features,INTERFACE_INCLUDE_DIRECTORIES>)
```

which can be added to the previous `target_include_directories` command.

noc0lour:
> MSVC 2013 [doesn't have an `inline` keyword](https://stackoverflow.com/questions/24736304/unable-to-use-inline-in-declaration-get-error-c2054) and that causes the bulk of the errors with that AppVeyor build. It seems that `cpu_features` targets at least MSVC 2015, although there may be workarounds (e.g. `-Dinline=__inline`).
> 
> There's also this error:
> 
> ```
> C:\projects\volk\cpu_features\src\filesystem.c(29): error C2065: '_SH_DENYWR' : undeclared identifier
> ```
> 
> which indicates that the [`share.h` header is needed](https://docs.microsoft.com/en-us/cpp/c-runtime-library/sharing-constants), although I can't find documentation for MSVC 2013 to verify and figure out why it's different from 2015+.

Oops, pressed the wrong button .... Maybe with this feature it would be advisable to drop support for MSVC 2013 and require MSVC 2015 and newer.

jdemel:
For some reason `set(BUILD_PIC ON)` is sometimes ignored in the `cpu_features` subdirectory. Setting it manually helps to overcome this problem. The reason for this behavior is yet unknown. "manually" implies to set it in the Travis CI config file.

Thus, the only failing tests are the QEMU ones. Since they are kind of redundant at this point, I suggest to remove them. If we want to keep a cross-compile target test, we need to figure out how to tell cpu_features which target it is supposed to find.

noc0lour:
Maybe adding more info as cpu_features has it in their travis will help? https://github.com/google/cpu_features/pull/27/files

noc0lour:
Ye they hardcode the `CMAKE_SYSTEM_PROCESSOR` matching to the travis config. maybe that's the way to go

.travis.yml
```diff
@@ -49,17 +49,20 @@ matrix:
...
+      env: MATRIX_EVAL="CC=gcc-7 && CXX=g++-7 && CMAKE_ARG=\"-DBUILD_PIC=ON -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/arm-linux-gnueabihf.cmake\" && QEMU_ARCH=arm"
       addons: {apt: {sources: "ubuntu-toolchain-r-test", packages: [*common_packages, g++-arm-linux-gnueabihf, qemu-user]}}
 
     # Job 8 ... ARMv8 (aarch64) cross compile
     - name: Linux ARMv8 (aarch64) Qemu GCC 7
-      env: MATRIX_EVAL="CC=gcc-7 && CXX=g++-7 && CMAKE_ARG=-DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/aarch64-linux-gnu.cmake"
+      env: MATRIX_EVAL="CC=gcc-7 && CXX=g++-7 && CMAKE_ARG=\"-DBUILD_PIC=ON -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/aarch64-linux-gnu.cmake\""
+           TOOLCHAIN=LINARO
+           TARGET=aarch64-linux-gnu
+           QEMU_ARCH=aarch64
```
jdemel: @noc0lour So we should do this:
https://github.com/google/cpu_features/blob/6d0767e0d9269919452dfa20a326cb4a6529171a/.travis.yml#L56

here as well? For some reason inline code links don't work.

noc0lour:
Well, that's only one part, since they have a bash script to run everything else https://github.com/google/cpu_features/blob/6d0767e0d9269919452dfa20a326cb4a6529171a/scripts/run_integration.sh

Edit: I do not suggest to copy everything from there, but maybe it can serve as inspiration on how we can add the correct arm cmake statements. Just check out what they pass to cmake in their travis builds on their latest commits.

jdemel:
I rebased this PR against the latest `master`.
I squashed some commits together to remove some of my trial and error tests with the CI system.

CMakeLists.txt
```diff
@@ -114,6 +114,10 @@ endif(MSVC)
 # Dependencies setup
 ########################################################################
 
+# cpu_features
+set(BUILD_PIC ON)
+add_subdirectory(cpu_features)
```
jdemel: This seems to be the offending part for all the PIC issues.
If you add a message here:
```cmake
set(BUILD_PIC ON)
message(STATUS "pic='${BUILD_PIC}'")
add_subdirectory(cpu_features)
```
It'll print `pic=ON`. But CMake will ignore this setting in the subdirectory `cpu_features`. There, this variable is unset again and consequently let's the build fail. If you run cmake with `-DBUILD_PIC=ON` it'll work. This is a really frustrating experience at the moment. 
@noc0lour, do you have any suggestion how to fix this?

michaelld: `CACHE` it. That's what the `-D` setting is, for all practical purposes. That way the variable is set in any subdirectory that might reference it, unless it has been explicitly removed from the cache.

jdemel: Thanks. That works. It did work without cache on my machine though.

.travis.yml
jdemel: These env variables were a really naive approach. The cross-compile etc. build needs more attention.

@noc0lour thanks for your suggestions.

jdemel:
Currently, this PR hangs on 2 issues:
- Cross-compile/QEMU CI. We have 2 cross-compile checks that would be great if they'd pass to validate that cross-compiles still work. Also, we'd know better how to cross-compile
  - I'd appreciate some help to solve this cross-compile issue because I don't have any prior experience with it.
- Remove obsolete checks and make sure all compilers are dealt with correctly.
  - I suspect we need more compile flag definitions in `gen/archs.xml` for MSVC.
  - We need to discuss a path forward to solve the `AppleClang` vs `Clang` confusion. I assume we can just say `AppleClang == Clang`. But I'd like to verify that.

michaelld:
> We need to discuss a path forward to solve the AppleClang vs Clang confusion. I assume we can just say AppleClang == Clang. But I'd like to verify that.

To a first order this is correct. There are very subtle differences, but from our Volk perspective I don't think we care. Thus, IMHO, we can use `MATCHES "*clang*"` instead of `STREQUAL` or whatever. I think we do this in some CMake scripts already.

lib/CMakeLists.txt
```diff
@@ -48,6 +48,13 @@ if(MSVC) #its not set otherwise
     set(COMPILER_NAME MSVC)
 endif()
 
+# Assume "AppleClang == Clang".
+string(TOLOWER ${COMPILER_NAME} COMPILER_NAME_LOWER)
+string(REGEX MATCH "clang" COMPILER_NAME_LOWER ${COMPILER_NAME_LOWER})
+if(${COMPILER_NAME_LOWER} MATCHES "clang")
+    set(COMPILER_NAME "Clang")
+endif()
+
```
jdemel, line 57: This is the necessary change to compile with optimized kernels on MacOS. The CI passes and shows that AVX/SSE etc. kernels are compiled and tested as well.

jdemel:
According to our CI on Windows, the architectures: 64, mmx and sse are not detected while sse2 etc are. This is an issue because missing sse implies that all SSE and AVX archs are disabled and thus, we only compile the generic machine.

This doesn't really change the situation if we still don't compile any optimized kernels for Windows but it would be great to finally be able to do so.

hcab14:
> * Cross-compile/QEMU CI. We have 2 cross-compile checks that would be great if they'd pass to validate that cross-compiles still work. Also, we'd know better how to cross-compile
>   
>   * I'd appreciate some help to solve this cross-compile issue because I don't have any prior experience with it.

It's important to keep checking the ARM SIMD kernels.

I'd be happy to have a look at this as I was one of the people adding `qemu`. Of course if someone else wants to have a look that's great, too.

ryanvolz:
`cpu_features` has seen some updates recently: https://github.com/google/cpu_features/releases. Maybe updating would help with the detections on Windows?

hcab14:
> > * Cross-compile/QEMU CI. We have 2 cross-compile checks that would be great if they'd pass to validate that cross-compiles still work. Also, we'd know better how to cross-compile

> I'd be happy to have a look at this as I was one of the people adding `qemu`. Of course if someone else wants to have a look that's great, too.

See: https://github.com/jdemel/volk/compare/cpu-features...hcab14:cpu-features
https://travis-ci.org/github/hcab14/volk/builds/736985570

Should I make a PR w.r.t. your cpu-features branch or w.r.t. volk master?

jdemel:
> > * Cross-compile/QEMU CI. We have 2 cross-compile checks that would be great if they'd pass to validate that cross-compiles still work. Also, we'd know better how to cross-compile
> >   
> >   * I'd appreciate some help to solve this cross-compile issue because I don't have any prior experience with it.
> 
> It's important to keep checking the ARM SIMD kernels.
> 
> I'd be happy to have a look at this as I was one of the people adding `qemu`. Of course if someone else wants to have a look that's great, too.

We do have native `aarch64` tests. TravisCI offers CI for different architectures. Nevertheless, the QEMU cross-compile tests help to verify that we are able to do cross-compiles. And I assume this is an important feature. Thus, I'd like to update these tests as well.
@hcab14 I'd be happy if you could have a look. A PR against `cpu-features` would be the way to go I assume.

jdemel:
> `cpu_features` has seen some updates recently: https://github.com/google/cpu_features/releases. Maybe updating would help with the detections on Windows?

I'll update the submodule pointer. I thought I'd done this but I might have missed it.
@ryanvolz currently one of our appveyor tests fails. Since this test uses really old versions at this point, do you think it is reasonable to remove it? It's based on python3.4 and VS12 2013 and everything before 3.6 is EOL as far as I know.

hcab14:
> @hcab14 I'd be happy if you could have a look. A PR against `cpu-features` would be the way to go I assume.

It's done, see https://github.com/jdemel/volk/pull/3
and https://github.com/gnuradio/volk/pull/393#issuecomment-711788909

ryanvolz:
> > `cpu_features` has seen some updates recently: https://github.com/google/cpu_features/releases. Maybe updating would help with the detections on Windows?
> 
> I'll update the submodule pointer. I thought I'd done this but I might have missed it.
> @ryanvolz currently one of our appveyor tests fails. Since this test uses really old versions at this point, do you think it is reasonable to remove it? It's based on python3.4 and VS12 2013 and everything before 3.6 is EOL as far as I know.

I think it is fine to drop this support. It won't affect my conda-forge builds (VS 2017 and minimum Python 3.6) or future gncdevelopment.com binaries (VS 2015 for previous release), and it's reasonable to ask anyone else building from source (I don't think they exist) to upgrade if necessary.

jdemel:
Well. This is a milestone for this PR. After months this is the first time all checks pass again. And MacOS builds use SSE/AVX. Something that wasn't the case before. I'd like to say thank you to everyone who helped to get to this point!

There are still some points on my list before I consider this PR ready.

jdemel:
For ARM, we don't check for `hard|softfp`. 
```cpp
static int i_can_has_softfp (void) {
    return 1;
}

static int i_can_has_hardfp (void) {
    return 1;
}
```
The [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html) state that this is OK because we can only compile for one or the other. An online check wouldn't give us any more info because we only compile one or the other machine.

The same argument applies for 32bit vs 64bit. We must compile for one or the other.
```cpp
static int i_can_has_32 (void) {
    return 1;
}

static int i_can_has_64 (void) {
    return 1;
}
```
Thus, I assume it is correct to just `return 1` and only have x32 or x64 machines available.

[MSVC docs](https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=vs-2019) State that on x64 only `AVX|AVX2|AVX512` are available. For x32 it is `IA32|SSE|SSE2|AVX|AVX2|AVX512`. 
Currently, we run our Windows CI on x64 and remove `mmx` and `sse` from our available archs. That causes all further archs to be removed as well because we rely on previous archs to be present to select available machines.

We do not distinguish between x32 or x64 for our arch checks. Thus, it is currently difficult to reliably choose one or the other architecture. What we do is, we choose `AVX` for SSSE3 etc. kernels for MSVC. It would be great to be able to compile all kernels with MSVC. I wonder if we should set the minimum compile flag to `/arch:AVX` and be done with it. 
`AVX` is already around for quite a while but some really small CPUs didn't support it until very recently. Maybe this is irrelevant because Windows doesn't support them either and we can rely on AVX but we can't really enable optimized kernels with MSVC builds as long as we don't know. This doesn't make the current situation any worse than before. But it would be great to make all kernels work.

jdemel:
I assume this PR is ready. Thus, I removed the wip and Don't Merge labels.

I just checked ABI compliance and couldn't find any changes. Thus, it is 100% ABI-compliant. That was my last bullet point to work on.

@michaelld I'd like to make this PR ready to be merged but I'd like to see another VOLK release before it gets merged. I'd like this PR to be around for a bit on `master` and wait if more people try and test it.

michaelld:
> I assume this PR is ready. Thus, I removed the wip and Don't Merge labels.
> 
> I just checked ABI compliance and couldn't find any changes. Thus, it is 100% ABI-compliant. That was my last bullet point to work on.
> 
> @michaelld I'd like to make this PR ready to be merged but I'd like to see another VOLK release before it gets merged. I'd like this PR to be around for a bit on `master` and wait if more people try and test it.

ACK ... I'll start reviewing / testing. Thx!

ValZapod:
> I assume this PR is ready

Windows now works? Can you compare the assembly? Is there any difference in speed in AVX or AVX2, i.e. it will mean that check now work correctly??

lib/CMakeLists.txt
michaelld: nice!

michaelld (approved):
LGTM

michaelld:
This all reads well to me & it passes all CI & the CI is now more robust. This may or not have a bug or 2 -- I don't see any but who knows? -- but it's the right way to be going & hence let's get it into the wild for folks to test / comment on!

michaelld:
My wondering now is how well this change will function on the new Arm Macs ... I'm looking forward to testing it out when I get time ... I have one of the beta ARM64 units for UHD / Volk / GR development ... making slow but regular progress!

PR #394, opened 2020-07-10
ci: Remove GCC 4.8 test

jdemel:
GCC 4.8 is too old to compile the latest Linux kernel by now. It was the
default in Ubuntu 14.04 which is obsolete as well. We shouldn't test
build chains that we really don't target anymore because they are
outdated and unsupported.


jdemel:
Since the original issues I saw with GCC 4.8 are resolved. I close this PR without merging it.

PR #396, opened 2020-07-31
Lower require CMake to 3.7.2

hjf:
Raspbian only has CMake 3.7.2 in the official repos, but this project requires 3.8.

I tried lowering the required CMake version by hand, building, and testing. It passed all tests. Seems to be working fine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@
 ########################################################################
 # Project setup
 ########################################################################
-cmake_minimum_required(VERSION 3.8)
+cmake_minimum_required(VERSION 3.7.2)
 set(CMAKE_BUILD_TYPE ${CMAKE_BUILD_TYPE} CACHE STRING "Choose build type: None Debug Release RelWithDebInfo MinSizeRel")
 project(volk)
 


jdemel:
I just checked and found a cmake 3.16 Debian package in the Raspbian Repos for Buster. Thus, I assume there's a way to update to cmake 3.16.
Besides, I found VOLK 2.3 in this repo too. Thus, I'd recommend to use this unless you want to develop new kernels or add more NEON versions. That would be great.

We moved to CMake 3.8 minimum because of several reasons. First, we wanted to be in sync with GNU Radio. Second, we used some cmake functionality that would only work with CMake 3.8.

jdemel (commented):
LGTM. Though, we had reasons to move to CMake 3.8 because we wanted to be able to use more modern CMake features. Thus, I'm hesitant.

hjf:
I wasn't aware of the release of Buster for Raspberry Pi. I'm using Stretch. The latest version of CMake there is 3.7.2, and the latest version of volk is 1.3-1+rpi1.

Maybe I should contact the maintainer of the volk package for Raspbian instead, since this only affects that distro.

jdemel:
Debian Stretch ran out of support more or less. At least everyone is encouraged to move to Buster. Also, newer versions of Raspbian are Buster-based. So I assume it is a good time to move to a new release. 
I saw VOLK 2.3 in the Raspbian Buster repos. (as opposed to 1.3). Thus, VOLK 2.3 may be in new Raspbian releases already.

jdemel:
@hjf are your issues resolved here? If yes, I'd like to close this. Thanks for having this discussion here.

jdemel:
I close this PR now. If you still see a need to lower the required CMake version, please feel free to reopen this issue.

PR #399, opened 2020-08-04
Added new kernel 4ic_deinterleave_8i_x2 with generic

### LINKED ISSUE #398: Add 4ic deinterleave to 8i x2
I think this can be done pretty efficiently, if bytewise arithmetic shift operators exist. Otherwise its probably just loop unrolling? The `volk_8ic_deinterleave_16i_x2` kernels are much more complicated than I expected though so I'm probably not aware of a lot of nuances of available SIMD operations.

```
uint8_t input[size];
uint8_t out_1[size];
uint8_t out_2[size];

for (int i = 0; i < size; i++) {
    out_1[i] = input[i] << 4;
    out_1[i] = out_1[i] >> 4;
    out_2[i] = input[i] >> 4;
}
```

dkozel:
Here's the generic implementation of a new 4bit signed integer deinterleave kernel. I'll be adding an SSE2 kernel shortly.

I'm interested in feedback on the code arrangement and implementation style, this is my first VOLK modification.


dkozel:
Clearly needs formatting. There appears to also be an issue with the input or output types which doesn't surprise me. Any feedback/advice welcome. I'll test for correctness when I can, preferably after sorting out the types issue.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2020 Derek Kozel
+ *
```
jdemel: This license header needs adjustment.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
...
+ * \page volk_4ic_deinterleave_8i_x2
+ *
+ * \b Overview
+ *
+ * Deinterleaves the complex 4-bit integer vector into I & Q vector data
+ * and converts them to 8-bit integers.
+ *
+ * <b> Dispatcher Prototype</b>
+ * \code
+ * void volk_4ic_deinterleave_8i_x2(int8_t* iBuffer, int8t_t* qBuffer, const lv_4sc_t complexVector, unsigned int num_points) \endcode
```
jdemel: I assume the input vector is passed by pointer as well.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
...
+ * \li qBuffer: The Q buffer output data.
+ *
+ * \b Example
+ * \code
+ * int N = 10000;
+ *
+ * volk_4ic_deinterleave_8i_x2();
+ *
+ * volk_free(x);
+ * \endcode
```
jdemel, line 51: An example with your use case in mind that describes input and output would be great.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
...
+                                                       unsigned int num_points)
+{
+    const int8_t* complexVectorPtr = (const int8_t*)complexVector;
+    int8_t* iBufferPtr = iBuffer;
+    int8_t* qBufferPtr = qBuffer;
+    for (unsigned int number = 0; number < num_points; number++) {
+       *iBufferPtr = (*complexVectorPtr) >> 4;
+       iBufferPtr++;
+       *qBufferPtr = ((*complexVectorPtr) << 4) >> 4;
+       qBufferPtr++;
```
jdemel: I assume you can just post-increment your output pointers in one line.
But it looks like you don't increment your input pointer at all. 
The generic kernels serves as a reference. All other kernel versions are validated against this one. That's how we run tests.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
...
+		// Store the results
+		_mm_storeu_si128(((__m128i*)(iBuffer)) + number, input_even);
+		_mm_storeu_si128(((__m128i*)(qBuffer)) + number, input_odd);
+	}
+
+    number = sixteenthPoints * 16;
+    for (; number < num_points; number++) {
+        iBuffer[number] = complexVector[number] >> 4;
+        qBuffer[number] = (complexVector[number] << 4) >> 4;
+    }
```
jdemel: I suggest to call the generic kernel for the tail to avoid copypasta.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
...
+		// Combine sign bits with deinterleaved input
+		input_even = _mm_or_si128(input_even, _mm_andnot_si128(mask, sign_even));
+		input_odd  = _mm_or_si128(input_odd, _mm_andnot_si128(mask, sign_odd));
+
+		// Store the results
+		_mm_storeu_si128(((__m128i*)(iBuffer)) + number, input_even);
+		_mm_storeu_si128(((__m128i*)(qBuffer)) + number, input_odd);
+	}
+
+    number = sixteenthPoints * 16;
```
jdemel: I suggest to integrate this into the for loop instead of reusing the variable `number`. It makes it easier to understand the code.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
...
+    // https://stackoverflow.com/questions/63200053/deinterleve-vector-of-nibbles-using-simd
+	const __m128i mask = _mm_set1_epi32(0x0F0F0F0F);
+	const __m128i signed_max = _mm_set1_epi32(0x07070707);
+
+    unsigned int number = 0;
+    const unsigned int sixteenthPoints = num_points / 16;
+	for (; number < sixteenthPoints; number++)
+	{
+		// Load and deinterleave input half-bytes
+		__m128i input_even = _mm_loadu_si128(((const __m128i*)(complexVector)) + number);
```
jdemel: I suggest to use pointer arithmetic and just increment by `16` after every load.
Also, this kernel is marked as `_a_sse2` (aligned). Therefore, I'd expect aligned loads and stores.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,109 @@
...
+		__m128i sign_even = _mm_cmpgt_epi8(input_even, signed_max);
+		__m128i sign_odd  = _mm_cmpgt_epi8(input_odd, signed_max);
+
+		// Combine sign bits with deinterleaved input
+		input_even = _mm_or_si128(input_even, _mm_andnot_si128(mask, sign_even));
+		input_odd  = _mm_or_si128(input_odd, _mm_andnot_si128(mask, sign_odd));
+
+		// Store the results
+		_mm_storeu_si128(((__m128i*)(iBuffer)) + number, input_even);
+		_mm_storeu_si128(((__m128i*)(qBuffer)) + number, input_odd);
```
jdemel: cf. `load`. I suggest pointer increments and aligned store calls.
You may put your intrinsics into their own function in `include/sse2_intrinsics.h` and include this header here. That would make it easier to implement aligned/unaligned kernels because they only differ in their load/store calls.

jdemel (commented):
Thanks for your PR. I put in a lot of comments. I hope that'll help.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
dkozel: Updated to the standard Copyright 2020 FSF.. Part of GNU Radio.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
dkozel: Signature updated to the correct `int8_t *`

kernels/volk/volk_4ic_deinterleave_8i_x2.h
dkozel: I've reworked it to use pointer arithmetic.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
dkozel: Reworked all the indexes to be more readable.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
dkozel: Changed to do so.

dkozel:
There's some const char * error at runtime. I'm not sure what the cause is. I'll have time to debug on the weekend. I changed the current SSE2 function to unaligned, I'll work on an aligned one with the "instrinsic intrinsic" setup once I have the first two kernels working.

odrisci:
The problem is that the qa subsystem does not handle sub-byte types: it sees the size of the input type as zero bytes and allocates a zero byte array to store it.

I think it might be easiest to rename the function to something like `volk_8i_deinterleave_nibbles_8i_x2` with signature:

```c++
void volk_8i_deinterleave_nibbles_8i_x2( int8_t * lsn, int8_t * msn, const int8_t *in, unsigned int num_samples);
```

Where `lsn` and `msn` are the least and most significant nibbles respectively. I think this interface would be better as it avoids making implicit assumptions about which nibble is I and which is Q

jdemel:
@odrisci thanks for pointing this out! Unfortunately, our system to deduce IO is fragile.

kernels/volk/volk_4ic_deinterleave_8i_x2.h
```diff
@@ -0,0 +1,127 @@
...
+                                                       unsigned int num_points)
+{
+    const int8_t* complexVectorPtr = complexVector;
+    int8_t* iBufferPtr = iBuffer;
+    int8_t* qBufferPtr = qBuffer;
+    for (unsigned int i = 0; i < num_points; i++) {
+        *iBufferPtr++ = (int8_t)(*complexVectorPtr) >> 4;
+        *qBufferPtr++ = ((int8_t)(*complexVectorPtr++) << 4) >> 4;
+    }
+}
```
jdemel, line 75: I don't know if a bitmask `0x0f` or bit shifts should be preferred but intuitively I'd prefer bitmasks because they seem to be more explicit.

The output is a signed type but the current implementation would never return a negative value. Is that intended? I assume no because the SSE impl takes explicit case of the sign bit.

odrisci:
@jdemel No problem. I don't think its too fragile - just if you want to handle packed bits you will probably need to think about it and plan for that. For example, here we are looking at a datatype `4ic` which is two 4-bit integers (a real and a complex) packed into a signed byte. But it is not obvious: a) which is the real and which is the imaginary; b) what is the encoding of the 4 bits. 

We have done some work in the GNSS community on a metadata format for GNSS IF data logs. Most GNSS front-ends use only 1 to 4 bits per sample and pack multiple samples from multiple frequency bands into words of 1, 2 or 4 bytes. Essentially we have seen almost every possible combination of bit and byte ordering (big endian bytes in the word with little endian arrangement of samples within a byte for example!) 

The metadata standard was recently approved by the Institute of Navigation in the US and can be found here if you want to have a look: https://sdr.ion.org/

Anyway, that's why I recommend using the `deinterleave_nibbles` name - you get the functionality you need without opening the sub-byte samples can of worms

kernels/volk/volk_4ic_deinterleave_8i_x2.h
odrisci: Unfortunately bitshifting negative numbers is undefined behaviour (or implementation defined, I can't remember) though most compilers will do what you expect.

The other problem here is that C/C++ only define integer arithmetic over `int` types, so smaller types are up-cast. So technically the result of

```c++
(int8_t)(*complexVectorPtr++) <<4
```

is of type `int`.

I would rewrite this as either:

```c
for (unsigned int i = 0; i < num_points; i++) {
        *iBufferPtr++ = (*complexVectorPtr >> 4);
        *qBufferPtr++ = (int8_t)((*complexVectorPtr++)  << 4) >> 4;
}
```
if you are happy with the undefined behaviour, or:

```c
for (unsigned int i = 0; i < num_points; i++) {
        *iBufferPtr++ = (*complexVectorPtr) /16;
        *qBufferPtr++ = (int8_t)((*complexVectorPtr++) *16) /16;
   }
```
if you are not. The latter might not optimise as well as the former, but is guaranteed to be correct. Note that I've left the last cast to `int8_t` as implicit in each case - I'm not sure if this agrees with your coding style.

ghostop14: Just saw the use case @dkozel was working on, and both nibbles would have a sign bit.  I wonder if this may be the correct way.  That would put the signed bits where they need to be then scale the values down appropriately?

for (unsigned int i = 0; i < num_points; i++) {
        *iBufferPtr++ = (*complexVectorPtr & 0xF0)/16;
        *qBufferPtr++ = (int8_t)((*complexVectorPtr++)  << 4)/16;
}

dkozel:
I wasn't able to figure out the best path forward here, and the immediate use has been worked around separately. I'm going to close and will open a new PR if I ever loop back around to this. Thanks all for the feedback and comments.

Issue #387, opened 2020-06-14
Inconsistent project name in CMake


skoslowski:
I getting the following warning from CMake when consuming VOLK (e.g. in GNU Radio):
```
CMake Warning (dev) at /usr/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:272 (message):
  The package name passed to `find_package_handle_standard_args` (VOLK) does
  not match the name of the calling package (Volk).  This can lead to
  problems in calling code that expects `find_package` result variables
  (e.g., `_FOUND`) to follow a certain pattern.
```
The (obvious) solution would be to rename those variables using `CMAKE_FIND_PACKAGE_NAME` instead of a hard-coded casing or, alternatively, use consistent casing. For backwards compatibly, I'd keep the current variables around until some future version TDB. 

First, however, the "right" casing of VOLK should be decided. Currently the cmake project name (and main target is "volk"), the export namespace and config module is "Volk" and the config variable prefix is "VOLK_". Ideally, these should all use the same casing.

jdemel:
I agree. 

Do you have a more comprehensive list of CMake lines that would need to be adjusted?
How did you trigger this warning? 

The preferred solution would probably be to rename all CMake variables to `VOLK` and `VOLK_`. Most of those `VOLK_` variables are not hardcoded but CMake generated. Thus, following this style would probably be the right way forward.
Though, the target name, for `target_link_libraries` is `Volk::volk` which is an inconsistency in itself. It should be `volk::volk`.

Do you have a suggestion how to fix this issue?

skoslowski:
I'd actually be willing to work on this myself, so I'll provide a list of changes in form of a PR =)
To trigger the warnings I simply configured a GR build on a fedora32 machine.

I strongly suggest to use the same casing for both targets and variables in the config module, see e.g. https://cmake.org/cmake/help/git-stage/module/FindPython.html

So its either VOLK, Volk or volk. A mix isn't really improving the current state.

In any case, the question of backwards compat needs to be decided. Break now or keep copy with the current names around? There isn't really a way to "warn on use", so all you can do is hope people read the release notes or have these the aliases default off.

drmpeg:
On Ubuntu, these issues show up if you use CMake version 3.17 or later.

jdemel:
@skoslowski did #403 already fix this issue or are there more changes required?

skoslowski:
[...]

PR #403, opened 2020-09-25
cmake: Remove "find_package_handle_standard_args" warning.

drmpeg:


drmpeg:
Fixes https://github.com/gnuradio/volk/issues/387

Goes with https://github.com/gnuradio/gnuradio/pull/3781

jdemel:
Thanks for posting this PR. From the discussion in #387 I conclude, we need to decide how we want to spell the project in terms of case. Candidates are:

- `volk`: current project name and library name.
- `Volk`: current CMake package name
- `VOLK`: Oldish package name?

As far as I can tell, we move away from `VOLK`, thus I consider this candidate to be out. Since our C prefix is `volk` it might be more consistent to go with `volk` everywhere. But currently, we go with `Volk` for CMake packages.

I'd really like to merge this PR but I assume we need to decide on that matter first.
Is there a way to introduce multiple package names? That would provide backwards compatibility.

drmpeg:
I'm not sure if any additional changes would actually be useful (given that CMake does all kinds of conversions to all upper case and all lower case).

BTW, the purpose of this fix is to match the case of the installed file ```VolkConfig.cmake```. Another way to accomplish the same thing would have been to rename ```VolkConfig.cmake``` to ```VOLKConfig.cmake```.

jdemel (approved):
I'd say it improves the current situation and doesn't break things. We can merge it.

I wouldn't close #387 yet because it tackles a broader issue.

I assume we can agree that the correct spelling should be `volk` or `Volk` everywhere in the project.

michaelld (approved):
LGTM

michaelld:
I agree with `volk` and `Volk` .... not `VOLK` ("screaming" as Marcus puts it so aptly)!

PR #404, opened 2020-10-05
build: Enable to not install volk_modtool

doronbehar:
Sometimes, for instance when building volk for usage with gnuradio, these files reference python, and you may wish to not have them on the machine.


jdemel (approved):
LGTM. 
The default keeps the current behavior. 
It enables users to not install an optional tool that requires Python.

michaelld (approved):
LGTM

PR #405, opened 2020-10-08
Update README Install Instructions for Successful `volk_profile` Upon Build/Install Completion 

awthomp:
The current VOLK build instructions do not explicitly instruct the user to run a `sudo ldconfig` command post VOLK build/installation. Without the ldconfig command, the `volk_profile` scrip will say that it can't find the libvolk shared object, causing the user to think something went wrong in the build process. This simple README change will clarify the install instructions and decrease user frustration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,8 +33,12 @@ $ make
 $ make test
 $ sudo make install
 
-# Link and cache shared library
+# Perform post-installation steps
+
+# Linux OS: Link and cache shared library
 $ sudo ldconfig
+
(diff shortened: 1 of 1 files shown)


michaelld (changes requested):
This is fine, but it is for Linux only, so please update the comment portion accordingly. You might also add that on Windows and macOS one needs to set the PATH environment variable correctly to access the installed executables (and, on Windows, the libraries too).

michaelld (approved):
LGTM; thanks!

jdemel (approved):
LGTM. Thanks!

PR #409, opened 2020-10-23
docs: Add hints how to handle required submodules

jdemel:
Now that we use `cpu_features`, we need to do a recursive `git
checkout`. That might be a stumbling block for some users. This commit
should help with some hints.


michaelld (approved):
LGTM thx!

PR #416, opened 2020-11-23
Include git submodules (i.e. cpu_features) in release tarball

ryanvolz:
Currently with cpu_features included as a submodule, using the release tarballs directly to build volk fails because the contents of the submodule are not included in the tarball. This is an issue for the conda-forge packages I maintain, because by policy those must be built from a tarball and not a git repository.

This PR adds to the release script so that the contents of the git submodules are added to the tarballs that are generated. The github-generated archives attached to every tagged version will still lack the contents of any submodules; this is apparently a known limitation with the only remedy being to attach your own release tarballs, which is already standard practice here.

As for the just-released version 2.4.0, it would be great to get new tarballs uploaded (with a "post" version?) that include the contents of cpu_features, but I can probably work around it for conda-forge if that's too messy.

### DIFF
--- a/scripts/tools/release.sh
+++ b/scripts/tools/release.sh
@@ -83,6 +83,12 @@ git tag --annotate --cleanup=verbatim -F "${annotationfile}" "${releaseprefix}${
 tarprefix="${project}-${version}"
 outfile="${tempdir}/${tarprefix}.tar"
 git archive "--output=${outfile}" "--prefix=${tarprefix}/" HEAD
+# Append submodule archives
+git submodule foreach --recursive "git archive --output ${tempdir}/${tarprefix}-sub-\${sha1}.tar --prefix=${tarprefix}/\${sm_path}/ HEAD"
+if [[ $(ls ${tempdir}/${tarprefix}-sub*.tar | wc -l) != 0  ]]; then
+  # combine all archives into one tar
+  tar --concatenate --file ${outfile} ${tempdir}/${tarprefix}-sub*.tar
+fi
 echo "Created tape archive '${outfile}' of size $(du -h ${outfile})."
 
 # 6. compress


michaelld (commented):
LGTM on basic read-through. I agree that we should be providing the submodule in the tarball, since that creates a whole archive that the user can use to build Volk offline otherwise. Since @jdemel is the one doing releases, I'll defer to him here as to the best way to go about doing this. Once we get the tarball dealt with, I'll look into the CMake scripting to make sure one can use an external install of the submodule, since that will be a desirable feature for some package managers (such as me in MacPorts under macOS) ...

ryanvolz:
> I'll look into the CMake scripting to make sure one can use an external install of the submodule, since that will be a desirable feature for some package managers (such as me in MacPorts under macOS) ...

This would also work for conda-forge (although the most recent cpu_features there is a bit old and I'd probably have to push through an update) if it is preferable to go that route instead of this and not just in addition to it.

michaelld:
> This would also work for conda-forge (although the most recent cpu_features there is a bit old and I'd probably have to push through an update) if it is preferable to go that route instead of this and not just in addition to it.

I think this is a "both and" situation :)

jdemel:
Apparently, we need a few updates to our release workflow.

- Add `cpu_features` as a submodule or rely on distro versions if they are available. I thought I'd add it as a submodule because we don't expose that API.
- Add [GPG signed release tarballs](https://wiki.debian.org/Creating%20signed%20GitHub%20releases) with
`gpg --detach-sign --armor volk-2.4.0.tar.xz`

I'll test your PR with the release workflow. My other comments just add to the things I'd need to figure out.

I assume, this calls for a bugfix release.

jdemel:
Thanks @ryanvolz it seems to work perfectly fine.

PR #417, opened 2020-11-23
Check for lib64 versus lib and set LIB_SUFFIX accordingly.

balister:
OE passes CMAKE_INSTALL_LIBDIR to the build. Use this variable to set
LIB_SUFFIX. It is very possible we could rewrite cmake to use this directly.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -209,6 +209,13 @@ if(DOXYGEN_FOUND)
 endif(DOXYGEN_FOUND)
 
 ########################################################################
+# Detect /lib versus /lib64
+########################################################################
+if (${CMAKE_INSTALL_LIBDIR} MATCHES lib64)
+    set(LIB_SUFFIX 64)
+endif()
+
+########################################################################
 # Setup the package config file
 ########################################################################
 #set variables found in the pc.in file


balister:
I created this for an OE build that used lib64 a long time ago. Looks like it does solve the lib/lib64 issue on Fedora. It will need checking on other distros.

CMakeLists.txt
```diff
@@ -208,6 +208,13 @@ if(DOXYGEN_FOUND)
     )
 endif(DOXYGEN_FOUND)
 
+########################################################################
+# Detect /lib versus /lib64
+########################################################################
+if (CMAKE_INSTALL_LIBDIR MATCHES lib64)
```
michaelld: CMake is confusing as to whether the left side can be `CMAKE_INSTALL_LIBDIR` or `${CMAKE_INSTALL_LIBDIR}`. I'd prefer the latter, since that makes it clear that we mean the contents of the variable and not the variable name itself.

CMakeLists.txt
```diff
@@ -208,6 +208,13 @@ if(DOXYGEN_FOUND)
...
+    set(LIB_SUFFIX 64)
```
michaelld, line 215: Do we provide an option for `LIB_SUFFIX` to be set elsewhere before this? If so, please check for whether the variable is set & if so then if it is -not- "64" then print a warning & then overwrite the variable as noted here.

michaelld (changes requested):
good start ... some wonderings on how to make this change more user-friendly

CMakeLists.txt
balister:
```
[balister@thuvia build (fix-lib64)]$ grep LIB_SUFFIX ../CMakeLists.txt 
    set(LIB_SUFFIX 64)
set(libdir "\${exec_prefix}/lib${LIB_SUFFIX}")
    DESTINATION lib${LIB_SUFFIX}/pkgconfig
set(VOLK_LIBRARY_DIR   lib${LIB_SUFFIX})
    set(CMAKE_MODULES_DIR lib${LIB_SUFFIX}/cmake)
[balister@thuvia build (fix-lib64)]$
```

```
[balister@thuvia build (fix-lib64)]$ grep LIB_SUFFIX -r ../cmake/*
../cmake/Modules/FindORC.cmake:	  PATHS ${ORC_ROOT}/lib${LIB_SUFFIX} ${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX})
../cmake/Modules/FindORC.cmake:	     PATHS ${ORC_ROOT}/lib${LIB_SUFFIX} ${CMAKE_INSTALL_PREFIX}/lib${LIB_SUFFIX})
[balister@thuvia build (fix-lib64)]$ 
```

I can't find any signs of anyone else setting LIB_SUFFIX

CMakeLists.txt
balister: Let me look. This patch is 4 years old and worked for a powerpc64 OE build and seems to work on my F3x machine. IS this what you are referring to? 
'''
The if command was written very early in CMake’s history, predating the ${} variable evaluation syntax, and for convenience evaluates variables named by its arguments as shown in the above signatures.
'''

balister: Updated.

CMakeLists.txt
balister: So the question is, how will this interact if you use this volk as a submodule for gnuradio? Will we bump the git has of volk on  gnuradio-3.8.2?

michaelld: Pretty sure we will never use this Volk as a submodule to GR. That would have to be GR 3.8, which is coming into bug-fix-mode only soon isn't it? I am so out of the GR dev loop at the moment; too saturated with work!

michaelld: I'd say let's leave the code as-is now. It's a good starting point & we can always fix it up down the road if for some reason GR does update the submodule to this commit / release.

michaelld (approved):

jdemel (approved):
LGTM. I was looking for a canonical CMake way to detect the libsuffix. Would your approach be the one I was looking for?

jdemel:
2 approvals and a very small change. Merging...

PR #418, opened 2020-11-28
Update release script

jdemel:
I added a bit more info that's printed to the shell. Also, all the GPG signature generation is added.

Most importantly:
`gpg --detach-sign --armor ${tag}.tar.gz`

This does follow [Debian Github release instructions](https://wiki.debian.org/Creating%20signed%20GitHub%20releases).

### DIFF
--- a/scripts/tools/release.sh
+++ b/scripts/tools/release.sh
@@ -52,7 +52,7 @@ $(git shortlog -e ${last_release}..HEAD)
 "
 echo "${shortlog}"
 
-#echo "${shortlog}" > ${deltafile}
+echo "${shortlog}" > ${deltafile}
 
 ${EDITOR} ${deltafile}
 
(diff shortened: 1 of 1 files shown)


scripts/tools/release.sh
```diff
@@ -113,14 +116,24 @@ signify-openbsd -C -p "${pubkey}" -x "${filelist}.sig"
 popd
 echo "checked."
 
+# 7.2 with GPG
+echo "signing tarballs with GPG ..."
+gpg --armor --detach-sign "${outfile}".gz
+gpg --armor --detach-sign "${outfile}".xz
+
```
michaelld, line 123: Do you want to add a note about (roughly) "Remember to attach the detached signature as binary to the release."? This comes from https://wiki.debian.org/Creating%20signed%20GitHub%20releases part 7 ... which seems desirable and very doable. I'm good either way, really.

michaelld (approved):
LGTM ... with a question

scripts/tools/release.sh
jdemel: All the signatures and tarballs (`.gz` and `.xz`) end up in the `archive` folder together with all the signature stuff. The last step I have to do is to manually upload all those files to a github release. I made sure the GPG `.asc` files end up in the archive folder as well. Thanks for the hint.

PR #420, opened 2020-11-30
cpu_features CMake option

maitbot:
Make use of cpu_features a CMake option with sensible defaults per arch.
(avoiding CMake failures of cpu_features for unknown architectures)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -114,14 +114,25 @@ endif(MSVC)
 # Dependencies setup
 ########################################################################
 
-# cpu_features
-set(BUILD_PIC ON CACHE BOOL
+# cpu_features - sensible defaults, user settable option
+if(CMAKE_SYSTEM_PROCESSOR MATCHES
+    "(^mips)|(^arm)|(^aarch64)|(x86_64)|(AMD64|amd64)|(^i.86$)|(^powerpc)|(^ppc)")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -114,14 +114,24 @@ endif(MSVC)
...
 set(BUILD_PIC ON CACHE BOOL
     "Build cpu_features with Position Independent Code (PIC)."
     FORCE)
 set(BUILD_SHARED_LIBS_SAVED "${BUILD_SHARED_LIBS}")
 set(BUILD_SHARED_LIBS OFF)
 add_subdirectory(cpu_features)
 set(BUILD_SHARED_LIBS "${BUILD_SHARED_LIBS_SAVED}")
+else()
+message(STATUS "Building Volk without cpu_features")
+endif()
```
jdemel: Could you add indentation in this section? This would make it easier to read and understand in the future.

jdemel (changes requested):
I have one comment, hence my request for changes but besides this looks really well. I checked with all our CI and it looks like it still builds and runs with AVX/NEON on all platforms if they are available.

jdemel (approved):
LGTM. Thanks!

michaelld (approved):
LGTM; thx!

PR #429, opened 2020-12-22
Update README.md

JamMarHer:
Advanced SIMD (aka NEON) is mandatory for AArch64.
There is no need for adding extra options to the compiler for 64 bit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ git submmodule update --init --recursive
 that'll pull in missing submodules.
 
 
-### Building on Raspberry Pi and other ARM boards
+### Building on Raspberry Pi and other ARM boards (32 bit)
 
 To build for these boards you need specify the correct cmake toolchain file for best performance.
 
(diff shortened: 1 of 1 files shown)


jdemel (commented):
Thanks for your contribution!

Do you mind to add your 64bit comment in the README as well? This might be helpful for new users.

michaelld (changes requested):
The README file needs some work ; I see a typo (likely more than 1) and some errors with respect to macOS. We need a section for how to handle Apple ARM64 / Aarch64. This change is fine, but I agree with @jdemel that I'd like to see your comment herein noted in the README file itself. @JamMarHer can you make this change? If not I'll look into a new PR that does this & more. Thx!

jdemel (approved):
LGTM.

Thanks for your contribution!

jdemel:
@michaelld do we need these commits to be signed off?

michaelld:
> @michaelld do we need these commits to be signed off?

I think these are good to go as-is. They are small and not to any code. Not sure if that's what you were going for.

michaelld (approved):
LGTM; thx!

michaelld:
I'm going to go ahead and merge based on my prior comment.

PR #432, opened 2020-12-28
cmake: Check that 'distutils' is available

rear1019:
See commit message for details.

### DIFF
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -100,6 +100,11 @@ endmacro(VOLK_PYTHON_CHECK_MODULE)
 # Sets the python installation directory VOLK_PYTHON_DIR
 ########################################################################
 if(NOT DEFINED VOLK_PYTHON_DIR)
+VOLK_PYTHON_CHECK_MODULE("distutils" "distutils" "True" HAVE_DISTUTILS)
+if(NOT HAVE_DISTUTILS)
+    message(FATAL_ERROR "Python module 'distutils' from Python’s Standard Library not found. On Debian install 'python3-distutils'.")
+endif()
+
 execute_process(COMMAND ${PYTHON_EXECUTABLE} -c "
 from distutils import sysconfig
 print(sysconfig.get_python_lib(plat_specific=True, prefix=''))


cmake/Modules/VolkPython.cmake
```diff
@@ -100,6 +100,11 @@ endmacro(VOLK_PYTHON_CHECK_MODULE)
 # Sets the python installation directory VOLK_PYTHON_DIR
 ########################################################################
 if(NOT DEFINED VOLK_PYTHON_DIR)
+VOLK_PYTHON_CHECK_MODULE("distutils" "distutils" "True" HAVE_DISTUTILS)
+if(NOT HAVE_DISTUTILS)
+    message(FATAL_ERROR "Debian does not include the 'distutils' module from Python’s Standard Library in its Python package; install 'python-disutils' or 'python3-distutils'")
```
nickoe: Minor comment: I am a bit on the edge with the wording. This is also true for ubuntu and probably also their derivatives.

Maybe reword to start focus on the module missing and add the help text afterward?: 
``` 
Python module 'distutils' from Python’s Standard Library not found. On Debian install 'python3-distutils'.
```

cmake/Modules/VolkPython.cmake
```diff
@@ -100,6 +100,11 @@ endmacro(VOLK_PYTHON_CHECK_MODULE)
 # Sets the python installation directory VOLK_PYTHON_DIR
 ########################################################################
 if(NOT DEFINED VOLK_PYTHON_DIR)
+VOLK_PYTHON_CHECK_MODULE("distutils" "distutils" "True" HAVE_DISTUTILS)
+if(NOT HAVE_DISTUTILS)
+    message(FATAL_ERROR "Python module 'distutils' from Python’s Standard Library not found. On Debian install 'python3-distutils'.")
+endif()
+
```
jdemel, line 107: Thanks for your fix.

I'd consider this whole `distutils` issue to be problematic because cmake seems to rely on it more or less:
https://gitlab.kitware.com/cmake/cmake/-/blob/master/Modules/FindPython/Support.cmake see line 470ff.

Also, I'd like to add that we require Py3k and thus https://cmake.org/cmake/help/git-stage/module/FindPython3.html would be great, though it is only available with cmake 3.12 and we require 3.8 only. If we want to continue to support Ubuntu 18.04 out of the box, we are pretty much restricted to cmake 3.10.

Besides, this looks good!

jdemel (approved):
LGTM.

I added a "wishlist" how to fix this issue at some point in the future with a newer CMake version.

michaelld (approved):
LGTM; good catch!

PR #437, opened 2021-01-17
travis: Reduce TravisCI usage

jdemel:
TravisCI introduced caps on how much CI we can run for free. Thus, we
need to reduce TravisCI usage. I removed the tests that are just "nice
to have" and are allowed to fail. In other words the `s390x` and
`ppc64le` architecture tests.

Also, I added a CI line to print the most important aspects of the used VOLK configuration.


michaelld (approved):
LGTM ; good idea for reducing CI time to be within our allocation!

PR #443, opened 2021-01-31
Fix segmentation fault when using GCC 8 (see #441)

### LINKED ISSUE #441: volk_profile segmentation fault (master branch)
I ran volk_profile (using a custom install directory) and manually setting my LD_LIBRARY_PATH on a ubuntu 20.04 and gcc 8
```
mkdir build
cd build
cmake .. -DCMAKE_INSTALL_PREFIX=../opt
make -j64
make test
make install

```
After running for a while it eventually segfaults, which is shown below and the back trace coming out of gdb:
I did try to rerun with the --path option to see if it would work, but it failed in the same manner. 
```
RUN_VOLK_TESTS: volk_32fc_s32f_power_spectral_densitypuppet_32f(131071,1987)
a_avx completed in 1169.55 ms
a_sse3 completed in 1170.15 ms
generic completed in 1096.39 ms
Best aligned arch: generic
Best unaligned arch: generic

Program received signal SIGSEGV, Segmentation fault.
0x000055e6edc938bb in std::filesystem::__cxx11::path::~path() ()
(gdb) bt
#0  0x000055e6edc938bb in std::filesystem::__cxx11::path::~path() ()
#1  0x000055e6edc938dc in std::filesystem::__cxx11::path::~path() ()
#2  0x000055e6edc938dc in std::filesystem::__cxx11::path::~path() ()
#3  0x000055e6edc9205c in write_results(std::vector<volk_test_results_t, std::allocator<volk_test_results_t> > const*, bool, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >) ()
#4  0x000055e6edc927ac in write_results(std::vector<volk_test_results_t, std::allocator<volk_test_results_t> > const*, bool) ()
#5  0x000055e6edc8ab00 in main ()

```

carlesfernandez:
This PR tries to fix the usage of the filesystem library with GCC 8 (which requires linking against -lstdc++fs), as reported at #441.


jdemel (approved):
Thanks for your fix! LGTM.

michaelld (approved):
LGTM ; interesting compiler idiosyncrasy to have to work around! Given how old GCC8 is, at least we don't have to worry about it being fixed :)

PR #444, opened 2021-01-31
ci: Move tests from TravisCI to Actions

jdemel:
We see issues with TravisCI because of their open source policy change. Here, I start to move tests to Github Actions. Particularly the default builds with clang and GCC are now on GH Actions.

What's available: GCC/Clang 7, 8, 9, 10
That's an improvement over the previous situation where we only had GCC 7 on GH Actions because TravisCI is unavailable.

Apparently, Actions does not yet support Clang 11. Also, old compilers are left out. Anything before GCC/Clang 7 is excluded. For GCC, CMake reports issues and some packages seem to be missing. For Clang, packages for these compilers are missing.
    
Actions does not support ARM etc. Thus, only x86 is supported. We might be able to add QEMU + cross-compiles. Also, ARM support seems to be on their roadmap for Q1'21.
    
Another step is included in the CI chain that prints the current configuration and hardware info as found by `list_cpu_features` and `volk-config-info`.

Missing features:
- QEMU cross compiles for ARM
- Intel SDE tests. Though, Actions seems to have AVX512 available and we might get away without this test.
- Native ARM builds. This might change in the near future. Support for ARM is on their roadmap.
- Older GCC/Clang before version 7. GCC 7.1 was released 2017-05-02. Clang 7 was released 2018-09-19. Is it worth the effort to add them back?

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,6 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build-ubuntu:
+    name: Build on ubuntu-latest ${{ matrix.compiler.name }}
 
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


jdemel:
This [run-on-arch action](https://github.com/uraimo/run-on-arch-action) might be a useful addition to our CI pipeline.

jdemel:
I added some tests for `arch64`. I hope we can still move to native ARM builds as soon as Actions offers that option.

michaelld (approved):
LGTM ; makes sense to move to GH Actions given all else going on with Travis ; checks look good, so I'm moving ahead here. Hopefully we can find a way to get ARM32 support working; also adding in Apple ARM64 support once it's available.

drmpeg:
For armv7, you have to have "-mfpu=neon" in the compiler flags. Maybe you could try the following.

cmake -DCMAKE_CXX_FLAGS:STRING="-mfpu=neon" -DCMAKE_C_FLAGS:STRING="-mfpu=neon" -DCMAKE_ASM_FLAGS:STRING="-mfpu=neon -g" ../

I've never been able to get clang to work on armv7. It always fails with a syntax error on some in-line assembly.

jdemel:
@drmpeg Thanks for the hint. If it compiles with GCC, we're already way better off. I'd assume that this is a native compilation in the container. Thus, I'd expect the flags to be present. Nevertheless, I don't know for sure and I didn't check yet. Would you be willing to have a look at this issue?

drmpeg:
Sure, I'll give it a try. BTW, here's the clang error.
```
/home/ubuntu/xfer/volktest/kernels/volk/volk_32fc_convert_16ic.h:216:9: error: couldn't allocate output register for constraint 'w'
        VCVTRQ_S32_F32(toint_a, ret1);
        ^
/home/ubuntu/xfer/volktest/kernels/volk/volk_32fc_convert_16ic.h:167:16: note: expanded from macro 'VCVTRQ_S32_F32'
    __VOLK_ASM("VCVTR.S32.F32 %[r0], %[v0]\n\t" \
```

drmpeg:
I couldn't get around the compiler identification issue. It looks like a bug in CMake.

https://gitlab.kitware.com/cmake/cmake/-/issues/20568

jdemel:
That's what I saw too. Thanks for figuring out the issue. I assume, we'd need to do a cross-compile and then run the test in such a Container.

PR #447, opened 2021-02-14
kernel: Fix typo in implementation name

jdemel:
The AVX2 kernels in `volk_16ic_x2_dot_prod_16ic` were named incorrectly.
Instead of `_avx2` the function names ended with `_axv2`. This commit
fixes the typo.

However, this does change two function names in files that we install.
This might require us to add a wrapper with the old name that just calls
the new, and correct, name.


michaelld (approved):
Good catch!

michaelld:
Thinking since this was caught so quickly that we just need to get it fixed in the code here since it will never have been "in the wild" to need further changes ... not 100% sure of course ...

michaelld:
I'm going ahead with the merge on the hopes that this will take care of any issues with those typos ... if not then we'll deal with it in another PR :)

PR #451, opened 2021-02-15
Fix python version detection

### LINKED ISSUE #450: volk fails to build with Python 3.10: string comparison in cmake '3.10.0' < '3.4'
Downstream bugzilla:
https://bugzilla.redhat.com/show_bug.cgi?id=1928144

The problem is in line #160 in CMakeLists.txt
VOLK_PYTHON_CHECK_MODULE("python >= 3.4" sys "sys.version.split()[0] >= '3.4'" PYTHON_MIN_VER_FOUND)
https://github.com/gnuradio/volk/blob/master/CMakeLists.txt#L160
```
>>> sys.version.split()[0]
'3.10.0a5'
>>> sys.version.split()[0] >= '3.4'
False
```
Credit for analysis: 
Tomáš Hrnčiar <thrnciar@redhat.com>

yarda:
Fixes #450

Signed-off-by: Jaroslav Škarvada <jskarvad@redhat.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,7 @@ endif()
 
 # Python
 include(VolkPython) #sets PYTHON_EXECUTABLE and PYTHON_DASH_B
-VOLK_PYTHON_CHECK_MODULE("python >= 3.4" sys "sys.version.split()[0] >= '3.4'" PYTHON_MIN_VER_FOUND)
+VOLK_PYTHON_CHECK_MODULE("python >= 3.4" sys "sys.version_info >= (3, 4)" PYTHON_MIN_VER_FOUND)
 VOLK_PYTHON_CHECK_MODULE("mako >= 0.4.2" mako "mako.__version__ >= '0.4.2'" MAKO_FOUND)
 
 if(NOT PYTHON_MIN_VER_FOUND)


yarda:
I guess the '>' is OK, because there are usually more data in the tuple (like e.g. 'stable'), so the '>' should much anything newer than '3.4', e.g. '3.4.0' stable version including.

CMakeLists.txt
```diff
@@ -157,7 +157,7 @@ endif()
 
 # Python
 include(VolkPython) #sets PYTHON_EXECUTABLE and PYTHON_DASH_B
-VOLK_PYTHON_CHECK_MODULE("python >= 3.4" sys "sys.version.split()[0] >= '3.4'" PYTHON_MIN_VER_FOUND)
+VOLK_PYTHON_CHECK_MODULE("python >= 3.4" sys "sys.version_info > (3, 4)" PYTHON_MIN_VER_FOUND)
```
michaelld: To be fair, should be ">=" since in theory we could use Python 3.4.0 & this change would prevent that version from being accepted.

michaelld (changes requested):
Very minor nitpick change. After that, I agree that we do need this change exactly for the reason in the issue noted! I did some playing around in Python and verified that this method works while the current method won't with Python 3.10 or even 3.9.10 ... basically any double-digit version portion.

michaelld (approved):
LGTM; thx!

PR #452, opened 2021-02-17
setup version and soversion for the lib

dernasherbrezon:
version and soversion are special properties of the library. they define
build version and API version respectively
(https://cmake.org/cmake/help/latest/prop_tgt/VERSION.html)

Additionally, refactor libver -> soversion to make it clear (googlable)

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -556,7 +556,8 @@ endif()
 if(NOT MSVC)
   target_link_libraries(volk PUBLIC m)
 endif()
-set_target_properties(volk PROPERTIES SOVERSION ${LIBVER})
+set_target_properties(volk PROPERTIES VERSION ${VERSION})
+set_target_properties(volk PROPERTIES SOVERSION ${SOVERSION})
 set_target_properties(volk PROPERTIES DEFINE_SYMBOL "volk_EXPORTS")
 
 #Install locations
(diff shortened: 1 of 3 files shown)


lib/CMakeLists.txt
```diff
@@ -556,7 +556,8 @@ endif()
 if(NOT MSVC)
   target_link_libraries(volk PUBLIC m)
 endif()
-set_target_properties(volk PROPERTIES SOVERSION ${LIBVER})
+set_target_properties(volk PROPERTIES VERSION ${VERSION})
+set_target_properties(volk PROPERTIES SOVERSION ${SOVERSION})
```
jdemel, line 560: We already set `SOVERSION` but not directly. Also, it might be good to set this in `VolkVersion.cmake`.

jdemel (approved):
LGTM.
It took me a while to get to the bottom of this. `LIBVER` is the name we currently use. We should've used `VERSION`/`SOVERSION` instead. Since Volk is a shared library, `SOVERSION` seems to be appropriate. Further, as long as `SOVERSION` is set, `VERSION` will assume the same value. This would be in line with our SemVer strategy,

lib/CMakeLists.txt
jdemel: The PR fixes this. I mark it as resolved.

jdemel:
@michaelld Do you have any further comments? I assume this PR can be merged.

michaelld (approved):
LGTM; thx!

Issue #430, opened 2020-12-27
error: ‘isinf’ was not declared in this scope


cherchais:
When I installed gnuradio3.8 using pybombs, I encountered the following error in compiling libvolk:
```bash
 Install tree:
|
\- gnuradio
   |
   \- libvolk
[INFO] Phase 1 complete: All binary dependencies installed.
[INFO] Phase 2: Recursively installing source packages to prefix:
[INFO] Installing package: libvolk
[WARNING] Build dir already exists: /home/lsh/pybombs/bladeRF/src/libvolk/build
Building:    (100%) [=========================================================================================================]
[  5%] Built target utils
[  8%] Built target cpu_features
[ 10%] Built target unix_based_hardware_detection
[ 13%] Built target list_cpu_features
[ 76%] Built target volk_obj
[ 77%] Built target volk
[ 79%] Building CXX object lib/CMakeFiles/volk_test_all.dir/testqa.cc.o
In file included from /home/lsh/pybombs/bladeRF/src/libvolk/build/include/volk/volk.h:31:0,
                 from /home/lsh/pybombs/bladeRF/src/libvolk/lib/qa_utils.h:5,
                 from /home/lsh/pybombs/bladeRF/src/libvolk/lib/kernel_tests.h:1,
                 from /home/lsh/pybombs/bladeRF/src/libvolk/lib/testqa.cc:31:
/home/lsh/pybombs/bladeRF/src/libvolk/include/volk/volk_common.h: In function ‘float log2f_non_ieee(float)’:
/home/lsh/pybombs/bladeRF/src/libvolk/include/volk/volk_common.h:153:24: error: ‘isinf’ was not declared in this scope
     return isinf(result) ? copysignf(127.0f, result) : result;
                        ^
/home/lsh/pybombs/bladeRF/src/libvolk/include/volk/volk_common.h:153:24: note: suggested alternative:
In file included from /usr/include/c++/5/complex:44:0,
                 from /home/lsh/pybombs/bladeRF/src/libvolk/include/volk/volk_complex.h:23,
                 from /home/lsh/pybombs/bladeRF/src/libvolk/build/include/volk/volk_typedefs.h:30,
                 from /home/lsh/pybombs/bladeRF/src/libvolk/build/include/volk/volk.h:29,
                 from /home/lsh/pybombs/bladeRF/src/libvolk/lib/qa_utils.h:5,
[...]

PR #378, opened 2020-05-02
ci: Remove Ubuntu 16.04 GCC5 test on TravisCI

jdemel:
This particular combination of Ubuntu and GCC does not compile due
to some bug in this Ubuntu version. Since this is a really old
Ubuntu version we decided to drop this test.

Users who want to use VOLK with this distro are adviced to stick with
older versions. It would be too much of a burden to work around a bug in
outdated software.

Issue #375, opened 2020-04-10
CI fails on g++5 Ubuntu 16.04 missing isinf


jdemel:
I observed this for all PRs since the updated Travis CI was merged. When `log2f_non_ieee` was added, we introduced a bug for Ubuntu 16.04 with G++ 5. Apparently it builds for gcc-5 for the C part but things fail on the C++ side.

On the one hand this a bug specific to Ubuntu 16.04/g++5 on the other hand, this is an LTS release. Arguably Ubuntu 16.04 is very old and you shouldn't use old compilers/distros with new libraries but we need to make a decision for this one how to handle this error.

1. Just drop the Ubuntu 16.04/g++5 test
2. Change CI to use Ubuntu 18.04/g++5 OR Ubuntu 14.04/g++5 (previous mode)
3. Fix this as a conditional bug with all kinds of possible side-effects.

[A similar bug report](https://github.com/facebookarchive/caffe2/issues/1633)

```
Scanning dependencies of target volk_test_all
[ 89%] Building CXX object lib/CMakeFiles/volk_test_all.dir/testqa.cc.o
In file included from /home/travis/build/gnuradio/volk/build/include/volk/volk.h:31:0,
                 from /home/travis/build/gnuradio/volk/lib/qa_utils.h:5,
                 from /home/travis/build/gnuradio/volk/lib/kernel_tests.h:1,
                 from /home/travis/build/gnuradio/volk/lib/testqa.cc:31:

/home/travis/build/gnuradio/volk/include/volk/volk_common.h: In function ‘float log2f_non_ieee(float)’:
/home/travis/build/gnuradio/volk/include/volk/volk_common.h:153:24: error: ‘isinf’ was not declared in this scope
     return isinf(result) ? copysignf(127.0f, result) : result;
                        ^

/home/travis/build/gnuradio/volk/include/volk/volk_common.h:153:24: note: suggested alternative:
In file included from /usr/include/c++/5/complex:44:0,
                 from /home/travis/build/gnuradio/volk/include/volk/volk_complex.h:23,
                 from /home/travis/build/gnuradio/volk/build/include/volk/volk_typedefs.h:30,
                 from /home/travis/build/gnuradio/volk/build/include/volk/volk.h:29,
                 from /home/travis/build/gnuradio/volk/lib/qa_utils.h:5,
                 from /home/travis/build/gnuradio/volk/lib/kernel_tests.h:1,
[...]

PR #455, opened 2021-03-08
gnuradio/volk/#430 move log2f_non_ieee back to kernel header

jbruno:
This minor change allows VOLK to build with older versions of GCC. Currently GCC6 is the minimum, and some developers are stuck using older tools.

Project fully builds for x86-64 as well as armv7. Tested by running volk_profile on both targets.

### DIFF
--- a/include/volk/volk_common.h
+++ b/include/volk/volk_common.h
@@ -145,13 +145,23 @@ union bit256 {
 ////////////////////////////////////////////////////////////////////////
 // log2f
 ////////////////////////////////////////////////////////////////////////
+#ifdef __cplusplus
+#include <cmath>
+#else
 #include <math.h>
+#endif
 // +-Inf -> +-127.0f in order to match the behaviour of the SIMD kernels
 static inline float log2f_non_ieee(float f)
 {
+#ifdef __cplusplus
+    float const result = std::log2f(f);
+    return std::isinf(result) ? std::copysignf(127.0f, result) : result;
+#else
     float const result = log2f(f);
     return isinf(result) ? copysignf(127.0f, result) : result;
+#endif
 }
+
 
 ////////////////////////////////////////////////////////////////////////
 // Constant used to do log10 calculations as faster log2
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ git submodule update --init --recursive
 that'll pull in missing submodules.
 
 
-### Building on Raspberry Pi and other ARM boards
+### Building on Raspberry Pi and other ARM boards (32 bit)
 
 To build for these boards you need specify the correct cmake toolchain file for best performance.
 
--- a/README.md
+++ b/README.md
@@ -64,6 +64,8 @@ that'll pull in missing submodules.
 
 To build for these boards you need specify the correct cmake toolchain file for best performance.
 
+_Note: There is no need for adding extra options to the compiler for 64 bit._
+
 * Raspberry Pi 4 `arm_cortex_a72_hardfp_native.cmake`
 * Raspberry Pi 3 `arm_cortex_a53_hardfp_native.cmake`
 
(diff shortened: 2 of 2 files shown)


michaelld (commented):
Thank you for the submission. This is quite fascinating! Can you explain how this change does what you're noting? I truly don't get how just moving the static function helps here, since it will be compiled one way or the other. Can you post build logs showing the error without this change?

jbruno:
Older versions of G++ (pre v6) don't automatically sub in "std::" when building C code. A user in #430 linked this:

https://developers.redhat.com/blog/2016/02/29/why-cstdlib-is-more-complicated-than-you-might-think/

which describes it well.

The fix comes about because the kernels are all C code, and like the syntax. The C++ code which includes volk_common.h does not like the syntax, however. The C++ function would require "std::isinf".

I got away with moving it to the kernel header, because none of the C++ application code actually uses that function. Only a couple of the kernels do.

jbruno:
[make_master_ububtu1604.log](https://github.com/gnuradio/volk/files/6104219/make_master_ububtu1604.log)
[cmake_master_ububtu1604.log](https://github.com/gnuradio/volk/files/6104220/cmake_master_ububtu1604.log)

Build logs attached. This is on Ubuntu 16.04 with GCC 5.4.0

michaelld:
> I got away with moving it to the kernel header, because none of the C++ application code actually uses that function. Only a couple of the kernels do.

Gotcha! Thank you for the detailed explanation. What you bring up is indeed a critical oversight in mixing C & C++ code. Can you test wrapping the original function with `extern "C" {}`? That might do the trick, and if so then it's a better and safer solution. I will boot up an older computer of mine that uses GCC5 optionally to see if I can replicate the issue & verify the fix independently.

jbruno:
It does not build, with the same errors. 

However, wrapping the entire function with a `#ifndef __cplusplus` macro does the trick. 

Clearly, this disables the function for C++ code, but it does have the added safety of not relying on the kernel headers being included in the right order.

michaelld:
hmm ... interesting. Thanks for testing! I think a point in having that code where it is is to make it available outside the kernel. Which means it needs to be usable in C or C++. Hence my hoping that `extern` would do the trick. I need to look at the code more to decide if I'm OK with just moving that function into the kernel.

jbruno:
Understood. It's a simple enough function that I can likely macro wrap 2 versions. Let me give that a shot and update the branch.

include/volk/volk_common.h
```diff
@@ -145,14 +145,24 @@ union bit256 {
...
 static inline float log2f_non_ieee(float f)
 {
+#ifdef __cplusplus
+    float const result = std::log2f(f);
+    return std::isinf(result) ? std::copysignf(127.0f, result) : result;
+#else
     float const result = log2f(f);
     return isinf(result) ? copysignf(127.0f, result) : result;
+#endif
 }
```
jdemel, line 163: This file should be in our C domain. If we can avoid C++ here, that'd make life easier. 
The old function signature looked like this:
```c
#include <math.h>
static inline float log2f_non_ieee(float f)
{
    float const result = log2f(f);
    return isinf(result) ? copysignf(127.0f, result) : result;
}
```
As fas as I recall, this builds with GCC 4.8 in our TravisCI system.
And it does [build with GCC 4/5 on godbolt](https://godbolt.org/z/9vcfGj)
We dropped our CI specifically for Ubuntu 16.04 with GCC 5 in #378 because this particular distribution would cause issues. There's a discussion in #375 with some further pointers. Ubuntu 16.04 has another ~1.5 months. EOL is 30.4.2021 (2021-04-30 if you prefer this format).

The error report indicates [`isinf`](https://en.cppreference.com/w/c/numeric/math/isinf) is missing.

Would it be possible to check for old GCC versions instead and put the whole `#ifdef` thing in there and leave a default version that corresponds to the current implementation? Right now, the code looks like this is a C/C++ differences issue. But I'd argue it is an old compiler issue instead. If the code reflects that, that would be great.

jbruno: After spinning up a few Docker containers with older toolchains (e.g. CentOS 7), I am able to build the C and C++ code as it exists in master.

It does appear that you are correct in that this is isolated to the GCC 5.4.0 implementation in Ubuntu 16.04, and perhaps some older and even less relevant versions of Ubutnu (which I did not test). Even the GCC 5.4 of your godbolt link compiles fine.

jbruno: Based on this, I'd say that checking compiler version is a less correct solution.

As the maintainers, I leave it up to you to reject the merge request at this point, if that's your preference based on how long you care about VOLK building on Ubuntu 16.04. If you want to keep going with the knowledge that this is likely isolated to Ubuntu 16.04, I'm happy to keep working with you.

jdemel:
Now that Ubuntu 16.04 is EOL, I'd say we can avoid this work around (and thus added complexity).

Still, I'd be happy to merge your README updates. Would you mind to reduce your PR to these commits?

jdemel:
Closing this PR now. There's reason to hope that this issue doesn't get raised again on later distros.

PR #456, opened 2021-03-09
Fix MSVC builds

Aang23:
Earlier today I attempted building VOLK in my Windows VM, which failed since MSVCC doesn't seem to support -fpermissive, so I fixed what was causing issues.

Thought I'd PR.


jdemel:
Thanks for your PR.

Casting everything to the correct type seems cumbersome. I hope we could go for a more straightforward approach and do:
`lv_cmake(.3f, 0.95393f)` instead of `(lv_32fc_t)lv_cmake(.3, 0.95393)`.

Would that work for you too?

Aang23:
That would totally work, I went with casting as a quick fix more than anything else. I'll edit my PR later today, thanks!

Aang23:
Done!

michaelld (approved):
LGTM ; thx!

jdemel:
Thanks for your contribution!

PR #458, opened 2021-03-14
conv_k7_r2: Disable broken AVX2 code

rear1019:
This has burned too many people. A point release should be considered.

### DIFF
--- a/kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
+++ b/kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
@@ -188,87 +188,87 @@ static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms,
 #endif /*LV_HAVE_SSE3*/
 
 
-#if LV_HAVE_AVX2
-
-#include <immintrin.h>
-#include <stdio.h>
-
-static inline void volk_8u_conv_k7_r2puppet_8u_avx2(unsigned char* syms,
-                                                    unsigned char* dec,
-                                                    unsigned int framebits)
-{
-
-
-    static int once = 1;
-    int d_numstates = (1 << 6);
-    int rate = 2;
-    static unsigned char* D;
-    static unsigned char* Y;
-    static unsigned char* X;
-    static unsigned int excess = 6;
-    static unsigned char* Branchtab;
-    static unsigned char Partab[256];
-
-    int d_polys[2] = { 79, 109 };
-
-
-    if (once) {
-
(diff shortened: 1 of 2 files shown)


kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
```diff
@@ -188,87 +188,87 @@ static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms,
...
-#include <immintrin.h>
-#include <stdio.h>
-
-static inline void volk_8u_conv_k7_r2puppet_8u_avx2(unsigned char* syms,
-                                                    unsigned char* dec,
-                                                    unsigned int framebits)
-{
-
-
-    static int once = 1;
```
jdemel, line 202: Might that be a reason it works in our CI but not in applications?

kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
```diff
@@ -188,87 +188,87 @@ static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms,
...
-#endif /*LV_HAVE_AVX2*/
+//#if LV_HAVE_AVX2
+//
+//#include <immintrin.h>
+//#include <stdio.h>
+//
+// static inline void volk_8u_conv_k7_r2puppet_8u_avx2(unsigned char* syms,
+//                                                    unsigned char* dec,
+//                                                    unsigned int framebits)
+//{
```
jdemel, line 199: I was really hoping someone would step up to fix this issue. But at this point I doubt it.

jdemel (approved):
Sad but LGTM.

I'd really like to have this kernel fixed. But unless we find someone with rather intimate knowledge of the algorithm and the code, this might be a really hard problem. Just the length and complexity of the function tells us that we should probably refactor it into multiple functions.

Given that, most FEC algorithms are just too complex to be squashed into a single kernel. They are better off in their distinct module. In order to ensure good performance they need quite a lot of state together with initialization etc.

kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
rear1019: If I remember correctly the puppet function are only used in testing.

michaelld (approved):
I'm OK with this as-is, though my preference -- mostly being pragmatic -- is to use `#if 0` and `#endif /* 0 */` around the code to be disabled rather than commenting out every single line, because it comes in such easy chunks to do this disabling over. If this were a bunch of smaller pieces of code, then I'd certainly use comments.

PR #462, opened 2021-04-11
ci: Update compiler names in CI

jdemel:
We use GH Actions with ubuntu-latest which defaults to 20.04 now.
Apparently, this does not install g++ but only gcc with the gcc package.
Thus, we install the appropriate g++ package which depends on gcc.


jdemel:
A small update to our CI. I just go ahed and merge it.

PR #463, opened 2021-04-11
travis: Remove Ubuntu 14.04 test

jdemel:
14.04 is EOL for some time now. This test breaks with some CMake issue.
Just remove it. Maybe, we can still use some TravisCI. At the moment it
looks like this is not going to happen any time soon.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,12 +26,6 @@ matrix:
       before_script:
         - cd ${TRAVIS_BUILD_DIR} && ./scripts/ci/download_intel_sde.sh
 
-    # Job 2 ... gcc-4.8
-    - name: Linux Ubuntu 14.04 (trusty) x86 GCC 4.8
-      dist: trusty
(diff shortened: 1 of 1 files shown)


michaelld (approved):
I'm good with this ... 14.04 is pretty old & I think it's gonna be quite rare that someone is still using it. 16.04 is still in use sometimes so we should support it for a while longer, but 18.04 is much more common.

jdemel:
Well, 16.04 will be EOL in about 2 weeks 30th April '21. I suggest to drop support for systems that are themselves not supported anymore.

michaelld:
Interesting; I didn't know that 16.04 was EOL so soon. Volk builds out of the box though, yes? I'd say we keep up CI for it for a while longer; we do have customers using 16.04 still, for better or worse. We encourage them to update to 18.04 if not 20.04, but some just can't yet.

PR #464, opened 2021-04-11
c11: Remove pre-C11 preprocessor instructions

jdemel:
We check for pre-C99 in `volk_complex.h`. But we require C11. Just remove this outdated check.


michaelld (approved):
LGTM ; good catch!

PR #465, opened 2021-04-11
ci: Add more debug output to tests

jdemel:
Print system config for more tests. Also, make sure fail-fast is disabled.


michaelld (approved):
LGTM

PR #466, opened 2021-04-14
Add kernel volk_32fc_accumulator_s32fc

Ka-zam:
New PR. Fixed some bugs in stdddev_and_mean that somehow slipped through as well.


jdemel:
This PR is a resubmission of #318 . There, we already have 2 approvals.
Though, this PR includes some fixes for stddev_and_mean as well.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
```diff
@@ -175,8 +175,8 @@ static inline void volk_32f_stddev_and_mean_32f_x2_neon(float* stddev,
 
     const float* in_ptr = inputBuffer;
 
-    __VOLK_ATTR_ALIGNED(16) float SumLocal[8] = { 0.f };
-    __VOLK_ATTR_ALIGNED(16) float SquareSumLocal[8] = { 0.f };
+    __VOLK_ATTR_ALIGNED(32) float SumLocal[8] = { 0.f };
+    __VOLK_ATTR_ALIGNED(32) float SquareSumLocal[8] = { 0.f };
```
jdemel, line 179: Why does the alignment need to be twice as much as expected? You doubled the other values as well.

jdemel (commented):
I have one remark. But besides, this PR LGTM.

kernels/volk/volk_32f_stddev_and_mean_32f_x2.h
Ka-zam: Because of temporary confusion on my part. But I fixed it and added some alignment directives for NEON as well. I think this is correct now.

jdemel: I'd assume we need 16byte alignment for NEON according to the [docs](https://developer.arm.com/documentation/ddi0344/f/unaligned-data-and-mixed-endian-data-support/unaligned-data-access-support/neon-data-alignment).

jdemel:
We have some random CI failures. I assume, we need to investigate those. `sum_of_poly` and `stddev_` are the two that fail randomly. I hope there aren't any other.
The failing CI test here is due to this random behavior.

Ka-zam:
I set the relative inaccuracy threshold to 1e-4 as opposed to the default 1e-2 without too much investigation. 
The new kernel is more accurate as it should be but perhaps the threshold is a bit too tight. I will make some test runs and see what I can find.

Ka-zam:
The reason for random failures are AFAICT that the generic implementation may fail on long vectors as the accumulator overflows. So the fix would be to implement the dual accumulator algorithm for the generic case.

jdemel:
> The reason for random failures are AFAICT that the generic implementation may fail on long vectors as the accumulator overflows. So the fix would be to implement the dual accumulator algorithm for the generic case.

So the generic implementation uses a different algorithm? For the sake of more reliable CI, it would be better to use the same algorithm in the generic impl.
BUT: I didn't see any `accumulator` failures. For now, I'd assume this kernel, that we should discuss here, seems to be fine. Unfortunately, other kernels, e.g. `sum_of_poly` and `stddev` seem to fail randomly. I suggest to move this discussion to a new thread.

Ka-zam:
Agreed, this kernel appears to be fine. It does use the same algorithm.

The random failures of stddev I think can be solved by updating the generic algorithm so that it doesn't overflow.

jdemel (approved):
This code received approval in #318 . It still looks good to me. Thanks!

michaelld (approved):
LGTM still ; @Ka-zam would you please squash everything into a single commit? There are some changes that are reverted within the commits; it will be much cleaner to just have a single commit with all of the changes. Simple squash keeping the commit messages as is, is fine with me. Thx!

michaelld:
@Ka-zam looks like the squash got messed up ... can you please try again?

jdemel:
This PR has 2 approvals and could be merged. I'm a bit hesitant though. 
@Ka-zam would it be correct to merge this PR, then move on to #469 to improve upon this PR? In that case, I'd like to have both PRs ready to be merged and then merge them both.

Ka-zam:
Yes, let's try to make this a bit more clean. I'll take a look as soon as #469 is merged.

PR #469, opened 2021-04-28
Fixed order of operations and updated the generic algorithm 

### LINKED ISSUE #471: accumulate_square_sum does not compile on ARMv7 CI
I just stumbled across this CI failure:
https://travis-ci.com/github/gnuradio/volk/jobs/502050291
This is something that needs to be investigated.

Ka-zam:
Fixed a subtle bug.


Ka-zam:
I tried to replicate the error but I can't. As far as I can tell the test vector is a float array of random numbers with mean 0.0 and variance 1.0, correct?

I get the following:
```
vlen  : 131071
iters : 100
mean  : 0
stddev: 1

NAME              TIME        MEAN    RELERR (dB)   STDDEV     REL ERR (dB)
---------------------------------------------------------------------------
2-pass double    648.34    0.00054404     N/A     1.00181375       N/A
naive  float     319.34    0.00054403    -48.0    1.00180852      -52.8
VOLK generic     116.24    0.00054403    -54.3    1.00180876      -53.0
VOLK u_sse        31.06    0.00054403    -53.7    1.00181317      -62.4
VOLK a_sse        30.97    0.00054403    -53.7    1.00181317      -62.4
VOLK u_avx        18.16    0.00054403    -54.8    1.00181389      -68.6
VOLK a_avx        17.87    0.00054403    -54.8    1.00181389      -68.6

```

For the problematic case where the variance is much small than the mean I get the following:
```
vlen  : 131071
iters : 100
mean  : 100
stddev: 0.1

NAME              TIME        MEAN    RELERR (dB)   STDDEV     REL ERR (dB)
---------------------------------------------------------------------------
2-pass double    638.49  100.00025287     N/A     0.09954573       N/A
naive  float     320.86   99.99999237    -55.8          -nan        nan
VOLK generic     118.29   99.99986267    -54.1    0.09954573      -71.7
VOLK u_sse        31.03  100.00028229    -65.3    0.09954570      -66.3
VOLK a_sse        31.28  100.00028229    -65.3    0.09954570      -66.3
VOLK u_avx        17.95  100.00025940    -71.9    0.09954610      -54.3
VOLK a_avx        17.86  100.00025940    -71.9    0.09954610      -54.3
```

jdemel:
Yes the test vector is a random vector. Though, I don't know the exact parameters. The tests provide the same input vectors to all kernels and compare the output against the generic kernel output. Thus, the generic kernel is the reference. The working assumption is that all kernels implement the same function and thus yield the same result. Obviously, there'd be some floating point error that we allow.
One issue we might see here, is that the CI system compares absolute errors not relative errors. That's a shortcoming with our tests. The relative errors look really good. My idea would be to allow a higher absolute error for now and rework our tests to be better readable and to improve them.

Ka-zam:
I swapped out the FMA for an explicit multiply and add and now it builds.

Is there a way of knowing if armv7 or armv8 is available? In that case the FMA instruction can be used instead. I'm actually not sure why it doesn't build on armv7 because according to the documentation this instruction is available:
[vfmaq_f32 instruction](https://developer.arm.com/architectures/instruction-sets/simd-isas/neon/intrinsics?search=vfmaq_f32)

jdemel:
Thanks for the fix! You just need to sign your commit.

The compiler error says

```
In file included from /home/travis/build/gnuradio/volk/kernels/volk/volk_16i_32fc_dot_prod_32fc.h:100:0,
                 from /home/travis/build/gnuradio/volk/build/lib/volk_machine_neon.c:40:

/usr/lib/gcc-cross/arm-linux-gnueabihf/7/include/arm_neon.h:1712:1: error: inlining failed in call to always_inline ‘vfmaq_f32’: target specific option mismatch

 vfmaq_f32 (float32x4_t __a, float32x4_t __b, float32x4_t __c)

 ^~~~~~~~~
```

the gist is `error: inlining failed in call to always_inline ‘vfmaq_f32’: target specific option mismatch`. The function is available but apparently it can't compile on ARMv7.

jdemel:
Thanks for working on this! If you want, you can merge in #474 which might be a further small improvement.

jdemel (approved):
LGTM. Thanks for working on this!

Ka-zam:
Thanks @drmpeg for the LV_HAVE_NEONV8 trick! Seems to be pretty complete now except for some build failure on MINGW but that seems unrelated to this PR.

michaelld (approved):
LGTM ; thx!

michaelld:
@jdemel if we merge in #474 then I think that conflicts with this PR, but I'd prefer to also fix the end-of-file missing returns ... and thus I'd prefer merging that other & then @Ka-zam might need to rebase this PR to be mergeable ... what do you think?

Ka-zam:
I added the newlines, can we merge now?

My thoughts are as the squash got messed up on the f32c_accumulator perhaps I will just abandon that PR and create a new fresh PR from the latest master when this is merged.

PR #474, opened 2021-05-15
Make _neon_accumulate_square_sum_f32() compatible with armv7.

drmpeg:
Signed-off-by: Ron Economos <w6rz@comcast.net>


drmpeg:
Sorry, I didn't see the other PR addressing this issue before I posted this one. However, since it passes CI, the ```#ifdef LV_HAVE_NEONV8``` trick should be useful.

include/volk/volk_neon_intrinsics.h
```diff
@@ -286,7 +286,12 @@ static inline float32x4_t _neon_accumulate_square_sum_f32(float32x4_t sq_acc,
...
     aux = vmulq_f32(aux, aux);
+#ifdef LV_HAVE_NEONV8
     return vfmaq_f32(sq_acc, aux, rec);
+#else
+    aux = vmulq_f32(aux, rec);
+    return vaddq_f32(aux, sq_acc);
+#endif
 }
 
-#endif /* INCLUDE_VOLK_VOLK_NEON_INTRINSICS_H_ */
```
jdemel, line 292: How did those newlines get lost? I suggest to keep them and make this PR even more concise.

jdemel (commented):
I suggest to remove the newline changes. Besides this PR LGTM.
We have two more PRs, #466 and #469 that deal with the same lines of code. I assume we merge this PR last.

include/volk/volk_neon_intrinsics.h
drmpeg: The newlines were lost in the previous commit. https://github.com/gnuradio/volk/commit/66a1b536bafe4d6e53ff88215231504ccc1893d2

My editor (I just use gedit on Ubuntu) will always insert a newline at the end of file. git diff hates missing a newline at end of file, so that's why I fixed the other two files.

jdemel: Ah. You re-added them. That's fine then.

michaelld (approved):
LGTM .. I'm OK with fixing the end-of-file newline here ... I'd prefer those sorts of things to be kept separate, but there aren't very many and the change is useful and otherwise non-impactful ... thus: let's just go with it unless someone else greatly disagrees!

jdemel:
I merged the other two PRs. So now, what shall we do with this one? I assume all issues are resolved now. I'm sorry this got a bit messy.

drmpeg:
I think it should just be closed. Looks like @Ka-zam covered all the bases.

PR #487, opened 2021-07-13
Update CODE_OF_CONDUCT file

michaelld:
subject says it all: GR just updated their CoC ; this change brings the Volk CoC in closer alignment with the GR CoC. Please note that I tweaked the language with respect to "regardless of" because I hate that language as it always forgets some group or class of people. I hope I got the language better & also the MD markup correct! This PR is testing for both of those. I am not committed to the language, but I do hope we can find a better way to be fully inclusive rather than via some special list.

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,80 +1,127 @@
-# Contributor Covenant Code of Conduct
+# VOLK Code of Conduct
 
-This Code of Conduct is also available [here on the GNU Radio
-wiki](https://wiki.gnuradio.org/index.php/Code_of_Conduct).
+This Code of Conduct is adapted from the GNU Radio Code of Conduct,
+available in the [gr-governance repository][GRCoC], dated 2021-07-12.
 
 ## Our Pledge
 
-In the interest of fostering an open and welcoming environment, we as
-contributors and maintainers pledge to making participation in our project and
-our community a harassment-free experience for everyone, regardless of age, body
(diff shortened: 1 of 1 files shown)


CODE_OF_CONDUCT.md
```diff
@@ -1,80 +1,127 @@
...
+We as members, contributors, and leaders strive to make participation
+in our community a harassment-free experience for
+everyone. Period. Qualities and characteristics we note specifically
+include regardless of age, body size, visible or invisible disability,
+ethnicity, sex characteristics, gender identity, gender expression,
+sexual identity, sexual orientation, level of experience, education,
+socio-economic status, nationality, personal appearance, race, caste,
+color, religion, and military status -- but lists like these cannot be
+complete and hence we note "for everyone. Period." in order to be as
+inclusive as possible.
```
jdemel: I assume this is the paragraph where you added some things?
I'd say the goal is clear. It sounds right. Only, I'd remove the "Period." parts. These make it sound grumpy.

jdemel (commented):
Thanks for taking care of this update. GR adapted its own CoC based on another source. I don't have the source handy at the moment.

CODE_OF_CONDUCT.md
michaelld: OK & That's fine with me. Here is [the original GR source](https://github.com/gnuradio/gr-governance/blob/main/CODE_OF_CONDUCT.md).

jdemel (approved):
LGTM.
If there is anything missing or incorrect, we can always improve it.

PR #491, opened 2021-09-06
LGPL resubmission

rear1019:
I, rear1019, hereby resubmit all my contributions to the VOLK project
and repository under the terms of the LGPLv3+.


rear1019:
(force push: rebased against correct branch)

michaelld (approved):
Although I prefer actual names rather than nicknames for the 2nd column entries, from what I can tell all of `rear1019`s submissions are in that nickname rather than some actual name. Hence, I believe that this PR meets the requirements we set for relicensing. If @jdemel thinks otherwise then he can decide to not approve; I leave approval and merging up to him.

jdemel (approved):
LGTM.

@rear1019 you always only use this nick in your commits etc. correct? This should be fine then.

PR #519, opened 2021-09-26
index: Fix volk_get_index

### LINKED ISSUE #516: volk_get_index broken, stuck in infinite loop
While debugging gnuradio/gnuradio#5013, I stumbled across `volk_get_index` calling itself indefinitely when using `volk_8u_x2_encodeframepolar_8u` (at least on Deb11). No matter the reason for finding the right implementation of that kernel failing, this function mustn't go into infinite recursion.

Maybe, we should just actually `return -1;` and deal with that (by `abort`ing) in the calling functions `volk_rank_archs` and `volk.tmpl.c`: `${kern.name}_manual`. That would at least make debugging easier.

```
(gdb) bt
#0  volk_get_index (impl_name=<optimized out>, n_impls=<optimized out>, impl_names=<optimized out>) at ./lib/volk_rank_archs.c:44
#1  volk_rank_archs (kern_name=kern_name@entry=0x7fd65caaf788 "volk_8u_x2_encodeframepolar_8u", impl_names=impl_names@entry=0x7fd65cb8da90 <volk_machine_avx2_64_mmx_orc+67632>, impl_deps=impl_deps@entry=0x7fd65cb8db50 <volk_machine_avx2_64_mmx_orc+67824>,
    alignment=alignment@entry=0x7fd65cb8dbb0 <volk_machine_avx2_64_mmx_orc+67920>, n_impls=n_impls@entry=4, align=align@entry=true) at ./lib/volk_rank_archs.c:68
#2  0x00007fd65c986088 in __init_volk_8u_x2_encodeframepolar_8u () at ./obj-x86_64-linux-gnu/lib/volk.c:10997
#3  __volk_8u_x2_encodeframepolar_8u (frame=0x10d7a80 "", temp=0x12c1da0 "", frame_size=16) at ./obj-x86_64-linux-gnu/lib/volk.c:11022
#4  0x00007fd65b998230 in gr::fec::code::polar_decoder_sc_systematic::generic_work(void*, void*) () from /gnuradio/build/gr-fec/lib/libgnuradio-fec.so.3.10.0git
#5  0x00007fd65b95a37f in gr::fec::decoder_impl::general_work(int, std::vector<int, std::allocator<int> >&, std::vector<void const*, std::allocator<void const*> >&, std::vector<void*, std::allocator<void*> >&) () from /gnuradio/build/gr-fec/lib/libgnuradio-fec.so.3.10.0git
#6  0x00007fd65cf2de7a in gr::block_executor::run_one_iteration() () from /gnuradio/build/gnuradio-runtime/lib/libgnuradio-runtime.so.3.10.0git
#7  0x00007fd65cfa159a in gr::tpb_thread_body::tpb_thread_body(std::shared_ptr<gr::block>, std::shared_ptr<boost::barrier>, int) () from /gnuradio/build/gnuradio-runtime/lib/libgnuradio-runtime.so.3.10.0git
#8  0x00007fd65cf92b44 in gr::thread::thread_body_wrapper<gr::tpb_container>::operator()() () from /gnuradio/build/gnuradio-runtime/lib/libgnuradio-runtime.so.3.10.0git
#9  0x00007fd65d1bf787 in boost::(anonymous namespace)::thread_proxy (param=<optimized out>) at libs/thread/src/pthread/thread.cpp:179
#10 0x00007fd65f0c3ea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
#11 0x00007fd65ee56def in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb) n

Thread 3 "fec_decoder4" hit Breakpoint 2, volk_get_index (impl_name=<optimized out>, n_impls=<optimized out>, impl_names=<optimized out>) at ./lib/volk_rank_archs.c:36
36	in ./lib/volk_rank_archs.c
(gdb) bt
#0  volk_get_index (impl_name=<optimized out>, n_impls=<optimized out>, impl_names=<optimized out>) at ./lib/volk_rank_archs.c:36
#1  volk_rank_archs (kern_name=kern_name@entry=0x7fd65caaf788 "volk_8u_x2_encodeframepolar_8u", impl_names=impl_names@entry=0x7fd65cb8da90 <volk_machine_avx2_64_mmx_orc+67632>, impl_deps=impl_deps@entry=0x7fd65cb8db50 <volk_machine_avx2_64_mmx_orc+67824>,
    alignment=alignment@entry=0x7fd65cb8dbb0 <volk_machine_avx2_64_mmx_orc+67920>, n_impls=n_impls@entry=4, align=align@entry=true) at ./lib/volk_rank_archs.c:68
#2  0x00007fd65c986088 in __init_volk_8u_x2_encodeframepolar_8u () at ./obj-x86_64-linux-gnu/lib/volk.c:10997
#3  __volk_8u_x2_encodeframepolar_8u (frame=0x10d7a80 "", temp=0x12c1da0 "", frame_size=16) at ./obj-x86_64-linux-gnu/lib/volk.c:11022
#4  0x00007fd65b998230 in gr::fec::code::polar_decoder_sc_systematic::generic_work(void*, void*) () from /gnuradio/build/gr-fec/lib/libgnuradio-fec.so.3.10.0git
#5  0x00007fd65b95a37f in gr::fec::decoder_impl::general_work(int, std::vector<int, std::allocator<int> >&, std::vector<void const*, std::allocator<void const*> >&, std::vector<void*, std::allocator<void*> >&) () from /gnuradio/build/gr-fec/lib/libgnuradio-fec.so.3.10.0git
#6  0x00007fd65cf2de7a in gr::block_executor::run_one_iteration() () from /gnuradio/build/gnuradio-runtime/lib/libgnuradio-runtime.so.3.10.0git
#7  0x00007fd65cfa159a in gr::tpb_thread_body::tpb_thread_body(std::shared_ptr<gr::block>, std::shared_ptr<boost::barrier>, int) () from /gnuradio/build/gnuradio-runtime/lib/libgnuradio-runtime.so.3.10.0git
#8  0x00007fd65cf92b44 in gr::thread::thread_body_wrapper<gr::tpb_container>::operator()() () from /gnuradio/build/gnuradio-runtime/lib/libgnuradio-runtime.so.3.10.0git
#9  0x00007fd65d1bf787 in boost::(anonymous namespace)::thread_proxy (param=<optimized out>) at libs/thread/src/pthread/thread.cpp:179
#10 0x00007fd65f0c3ea7 in start_thread (arg=<optimized out>) at pthread_create.c:477
#11 0x00007fd65ee56def in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
(gdb)
```

This is a tail recursion in `volk_get_index`. Sadly, it will never terminate (until it causes memory exhaustion in ctest, I guess). 

`volk_get_index` is a bag of mixed emotions for me:

```C
int volk_get_index(const char* impl_names[], // list of implementations by name
                   const size_t n_impls,     // number of implementations available
                   const char* impl_name     // the implementation name to find
)
{
    unsigned int i;
        if (!strncmp(impl_names[i], impl_name, 20)) {
            return i;
        }
    }
    // TODO return -1;
    // something terrible should happen here
    fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
    return volk_get_index(impl_names, n_impls, "generic"); // but we'll fake it for now
}
```

1. `strncmp(a,b,20)`: We'll happily compare equal impls that are only different after the first 20 characters.
2. It never checks whether the impl that can't be found *is* the `_generic` one. If that is the case, it just recurses to looking up the generic one, and we end up where we are.

What *is* confusing is why neither `volk_get_info` nor the `volk_profile` tool *know* of `volk_8u_x2_encodeframepolar_8u_generic`. In fact, the latter doesn't seem to want to know any implementations of that at all.

_Originally posted by @marcusmueller in https://github.com/gnuradio/gnuradio/issues/5013#issuecomment-923369047_

jdemel:
This function results in an infinite loop on Debian 11 for some impls.
This is a first step to fix it.

@mbr0wn , @marcusmueller does that fix the reported issue? If you could point me to a CI script for Debian 11 that we can integrate here, that'd be very helpful.

Fix #516

EDIT
This is a minimal Debian 11 Container to test things.
```docker
FROM debian:bullseye

RUN apt-get update \
        && apt-get install -y \
        build-essential git cmake python3-mako python3-distutils liborc-dev
```
[GNU Radio CI container for Debian 11](https://hub.docker.com/layers/gnuradio/ci/debian-11-3.10/images/sha256-d3de16fb91a50ee2472d35bd1fb27e18715e4dba69098bf3b99e0109aaf0c62c?context=explore) where `volk_profile` hangs on `volk_32f_8u_polarbutterflypuppet_32f`.
The corresponding Dockerfile: [Dockerfile](https://github.com/gnuradio/gnuradio-docker/blob/master/ci/ci-debian-11-3.10/Dockerfile).

### DIFF
--- a/lib/volk_rank_archs.c
+++ b/lib/volk_rank_archs.c
@@ -22,13 +22,16 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
 {
     unsigned int i;
     for (i = 0; i < n_impls; i++) {
-        if (!strncmp(impl_names[i], impl_name, 20)) {
+        if (!strncmp(impl_names[i], impl_name, 42)) {
             return i;
         }
     }
     // TODO return -1;
     // something terrible should happen here
     fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
+    if (strncmp(impl_name, "generic", 20)) {
+        return -1;
+    }
     return volk_get_index(impl_names, n_impls, "generic"); // but we'll fake it for now
 }
 
--- a/lib/volk_rank_archs.c
+++ b/lib/volk_rank_archs.c
@@ -22,7 +22,7 @@ int volk_search_index(const char* impl_names[],
 {
     unsigned int i;
     for (i = 0; i < n_impls; i++) {
-        if (strncmp(impl_names[i], impl_name, 42) != 0) {
+        if (!strncmp(impl_names[i], impl_name, 42)) {
             return i;
         }
     }
@@ -48,10 +48,10 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
     idx = volk_search_index(impl_names, n_impls, "generic");
     if (idx >= 0) {
         fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
-        return idx;
+    } else {
+        fprintf(stderr, "Volk ERROR: no arch found. generic impl missing!\n");
     }
-    fprintf(stderr, "Volk warning: no arch found, returning -1 aka not found!\n");
-    return -1;
+    return idx;
 }
 
 int volk_rank_archs(const char* kern_name,    // name of the kernel to rank
--- a/lib/volk_rank_archs.c
+++ b/lib/volk_rank_archs.c
@@ -1,6 +1,6 @@
 /* -*- c++ -*- */
 /*
- * Copyright 2011-2012 Free Software Foundation, Inc.
+ * Copyright 2011-2012, 2021 Free Software Foundation, Inc.
  *
  * This file is part of VOLK
  *
@@ -15,24 +15,43 @@
 #include <volk/volk_prefs.h>
 #include <volk_rank_archs.h>
 
+// This function is supposed to be private to this compilation unit.
+int volk_search_index(const char* impl_names[],
+                      const size_t n_impls,
+                      const char* impl_name)
+{
+    unsigned int i;
+    for (i = 0; i < n_impls; i++) {
+        if (strncmp(impl_names[i], impl_name, 42) != 0) {
+            return i;
+        }
+    }
+    return -1; // Indicate we couldn't find anything!
+}
+
 int volk_get_index(const char* impl_names[], // list of implementations by name
                    const size_t n_impls,     // number of implementations available
                    const char* impl_name     // the implementation name to find
 )
 {
-    unsigned int i;
-    for (i = 0; i < n_impls; i++) {
-        if (!strncmp(impl_names[i], impl_name, 42)) {
-            return i;
-        }
+    /*
+     * We follow a 3 step process.
+     * 1. Search for the requested impl. Return index if found.
+     * 2. Search for the generic impl. Return as fail-safe.
+     * 3. Return -1 to indicate an error. Caller needs to handle this case.
+     */
+    int idx = volk_search_index(impl_names, n_impls, impl_name);
+    if (idx >= 0) {
+        return idx;
     }
-    // TODO return -1;
-    // something terrible should happen here
-    fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
-    if (strncmp(impl_name, "generic", 20)) {
-        return -1;
+
+    idx = volk_search_index(impl_names, n_impls, "generic");
+    if (idx >= 0) {
+        fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
+        return idx;
     }
-    return volk_get_index(impl_names, n_impls, "generic"); // but we'll fake it for now
+    fprintf(stderr, "Volk warning: no arch found, returning -1 aka not found!\n");
+    return -1;
 }
 
 int volk_rank_archs(const char* kern_name,    // name of the kernel to rank


marcusmueller:
Don't we need to handle -1 in the calling functions to avoid calling into random memory?

michaelld:
> Don't we need to handle -1 in the calling functions to avoid calling into random memory?

+1 from me on this: looks like that's just in tmpl/volk.tmpl.c : 191

lib/volk_rank_archs.c
```diff
@@ -34,13 +34,16 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
 {
     unsigned int i;
     for (i = 0; i < n_impls; i++) {
-        if (!strncmp(impl_names[i], impl_name, 20)) {
+        if (!strncmp(impl_names[i], impl_name, 42)) {
```
michaelld, line 25: since `strncmp` returns `int` of "less than, equal to, or greater than zero" can we go with `strncmp () == 0` here? Compilers will know how to optimize the code here, and this avoids any possible issues with binary not (`!`) and negative integers. Yes this code should work fine, but it's just as easy to compare with 0.

lib/volk_rank_archs.c
```diff
@@ -34,13 +34,16 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
...
             return i;
         }
     }
     // TODO return -1;
     // something terrible should happen here
     fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
+    if (strncmp(impl_name, "generic", 20)) {
```
michaelld: can we use 7 or 8 instead of 20? "generic" is 7 bytes long, so 20 way more than necessary ...

michaelld: IDK whether we use 7 or 8 bytes ... would have to test `strncmp` to see ...

michaelld: Also: If we're adding in this check, then let's save the result of this `strncmp` and return it instead of the recursive call to this function. Once through and done.

michaelld (changes requested):

lib/volk_rank_archs.c
jdemel: If we change the recursive structure, we'd need to change the whole function quite significantly.
Right now, we follow a 3 step process:
1. Search for the user requested impl
2. If user request cannot be fulfilled: Search for generic impl.
3. New third step: Give up and return `-1`.
I'm actually not entirely sure, if the term `generic` is the only thing we ever search for. It might be smth like `vector_generic` as well. However, not during the second pass.

Right now, it feels like we actually need 2 functions. One with the loop and one to handle all the different error conditions.

lib/volk_rank_archs.c
jdemel: I really hope it is `!= 0`. Fixed it.

jdemel:
I fixed the "file local comments". Thus, we need to discuss how to deal with the `-1` return value.
Why did we avoid to deal with it until now? We ended up in an infinite loop and thus terminated useful execution at this point.
Considering this, the actual bug report should be fixed now.

`volk_get_index` is used in 2 functions:
1. `lib/volk_rank_archs.c:volk_rank_archs`: return values propagate here.
2. `tmpl/volk.tmpl.c:186`, i.e. in all `_manual` kernel calls.

`volk_rank_archs` is used in "one" place.
- `tmpl/volk.tmpl.c:151-152`: kernel init. Thus, the first call to any kernel.

Both functions in `tmpl/volk.tmpl.c` return void. We can't signal errors that way. We'd have to change the API.
We go back and forth between signed and unsigned variables in these functions. But that's a separate issue.

We could treat `-1` in both functions as: use the first impl in the list and hope it is supported on your system.
I'd really like to raise a `NotImplementedError` but that's obviously not an option either.

I know this is an unsatisfactory but we could ignore this problem for now and add it to the list of issues to fix for our next major release. Would that be a viable option?

jdemel:
I reverted the change where `!` is replaced with `!= 0`. It caused issues with most systems except Ubuntu x86. I rebased to the latest `main` as well.

michaelld (approved):
LGTM; this change will be an API change (returning -1), so do we want to just have it in a 3.0-beta branch & not related to 2.X at all?

michaelld:
> I reverted the change where `!` is replaced with `!= 0`. It caused issues with most systems except Ubuntu x86. I rebased to the latest `main` as well.

interesting ... I wonder why? Anyway, `!` is fine

michaelld:
This PR fixes the direct issue, which is important. It does change the API unfortunately by now returning -1 where it didn't before. I guess as an interim we could instead have it return the first entry (if there is one) & print a warning noting as such. Of course if there is no entry then we'd want to return -1 ... so, does this help? Hmmm ...

jdemel:
> This PR fixes the direct issue, which is important. It does change the API unfortunately by now returning -1 where it didn't before. I guess as an interim we could instead have it return the first entry (if there is one) & print a warning noting as such. Of course if there is no entry then we'd want to return -1 ... so, does this help? Hmmm ...

It sounds like every solution somehow breaks the API. However, at the moment the broken API won't work at all. After this change, it will. If we'd do another 2.x release, I'd argue this fix should go in there.
I added a message "Volk ERROR ...". If we'd `return 0;`, we'd most probably cause an "illegal instruction error". 

Besides, it might be time to add a Debian test. However, [Debian 'bullseye' ships GCC 10](https://packages.debian.org/en/bullseye/devel/gcc). We already have a test with GCC 10. The question would be: Is this distro specific?

lib/volk_rank_archs.c
```diff
@@ -27,21 +27,43 @@
...
+        return idx;
+    }
+
+    idx = volk_search_index(impl_names, n_impls, "generic");
+    if (idx >= 0) {
+        fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
+    } else {
+        fprintf(stderr, "Volk ERROR: no arch found. generic impl missing!\n");
+    }
+    return idx;
```
marcusmueller, line 54: we're just bubbling up the -1 here in case of error, but we need to handle that. Otherwise, we'll jump into some random memory location when the consumer program actually calls the kernel. That's worse than an infinite loop.

So, this is half the fix so far.

marcusmueller (changes requested):

lib/volk_rank_archs.c
marcusmueller: at the very least, `__init_${kern.name}` needs to check for -1 and `abort` in that case. Failing early instead of wandering into random memory, where we become a stability or security issue, is necessary. Otherwise, we're just making bad code worse!

marcusmueller: also note that `__init_..` casts the `int` that `volk_rank_archs` returns into a `size_t`, which is unsigned, and potentially of a different size!

marcusmueller:
By the way, how do we figure out what breaks finding that `generic` implementation? even if this PR was perfect in every way, it would *not* solve the underlying issue that leads to https://github.com/gnuradio/volk/issues/516 . And maybe fixing that would completely rewrite this part of code, anyway...

jdemel:
Updated the initial PR description.
I started some tests with:
```docker
FROM debian:bullseye

RUN apt-get update \
        && apt-get install -y \
        build-essential git cmake python3-mako python3-distutils liborc-dev
```
I couldn't find `liborc-dev` and `python3-distutils` in the GNU Radio CI Docker file.
`liborc-dev` is essentially a dependency and should be available anyways.
`python3-distutils` is only a build dependency. So it won't be installed with a binary.

Besides, I can't reproduce the issue in my minimal container with the current `main`. Debian 11 ships VOLK 2.4.1 which I didn't test yet.

marcusmueller:
> liborc-dev is essentially a dependency and should be available anyways.

liborc-**dev** should only be a build-time dependency, right?

jdemel:
I converted this PR to a draft. At the moment I doubt it really fixes the issue it is supposed to fix. We need a better understanding of the issue here.

jdemel:
This draft PR was introduced because of a bug that turned out to be outside of VOLK. I'm closing this PR now.

PR #523, opened 2021-09-30
Fix allocated size not a multiple of alignment

AlexandreRouma:
In some cases, it is required that the allocated size is a multiple of the alignment size.

When using `-fsanitize=address` to do memory debugging with libasan, it errors out complaining about the size of the allocation not being a multiple of the alignment. This makes memory debugging basically impossible. I've confirmed that this fix lets me use the address sanitizer properly.

This pull request ensured that this requirement is always met.

### DIFF
--- a/lib/volk_malloc.c
+++ b/lib/volk_malloc.c
@@ -50,6 +50,17 @@
 
 void* volk_malloc(size_t size, size_t alignment)
 {
+    if ((size == 0) || (alignment == 0)) {
+        fprintf(stderr, "VOLK: Error allocating memory: either size or alignment is 0");
+        return NULL;
+    }
+    // Tweak size to satisfy ASAN (the GCC address sanitizer).
+    // Calling 'volk_malloc' might therefor result in the allocation of more memory than
+    // requested for correct alignment. Any allocation size change here will in general not
+    // impact the end result since initial size alignment is required either way.
+    if (size % alignment) {
+        size += alignment - (size % alignment);
+    }
 #if HAVE_POSIX_MEMALIGN
     // quoting posix_memalign() man page:
     // "alignment must be a power of two and a multiple of sizeof(void *)"


jdemel:
Hi @AlexandreRouma . Thanks for your contribution.

A few notes
- We are in the process to re-license VOLK under LGPL. Thus, we can only accept new contributions under the new license. I want to make sure, you're aware of that.
- We generally require signed commits. `git commit -s`. It adds a `Signed off by ...` line at the end of your commit message. The link behind our DCO check explains this.

lib/volk_malloc.c
```diff
@@ -50,6 +50,10 @@
 
 void* volk_malloc(size_t size, size_t alignment)
 {
+    // Round size to a multiple of the alignment
+    if (size % alignment) {
+        size += alignment - (size % alignment);
+    }
```
jdemel: I'd argue the code is pretty clear. And does not require this comment.

However, I suggest to write a comment that explains the reason to do this change.
```
// Make ASAN happy...
```

jdemel (commented):
I'm a bit hesitant with that change because it affects a third party tool that is out of scope of our work.

Should we add a CI test? It'd notify us if we break this fix.

AlexandreRouma:
> We are in the process to re-license VOLK under LGPL. Thus, we can only accept new contributions under the new license. I want to make sure, you're aware of that.

Yes I am aware of it.

> I'm a bit hesitant with that change because it affects a third party tool that is out of scope of our work.
> Should we add a CI test? It'd notify us if we break this fix.

Well, I'd argue it's not really third party, it's basically part of GCC.

Here is a test you can run to check if it's broken. This will error out on my system (Archlinux amd64) with the latest libvolk commit. Adding the lines successfully fixes the issue.
```c
#include <volk/volk.h>

int main() {
    // Allocate a prime number of bytes to make sure it's not a multiple of the alignment.
    float* test = volk_malloc(113, volk_get_alignment());
    volk_free(test);
    return 0;
}
```
Compiler command: `gcc asan_test.c -lvolk -g -fsanitize=address -o asan_test`
GCC version 11.1.0

Execution prints out
```
=================================================================
==1924==ERROR: AddressSanitizer: invalid alignment requested in aligned_alloc: 32, alignment must be a power of two and the requested size 0x71 must be a multiple of alignment (thread T0)
    #0 0x7f638bcb3dd9 in __interceptor_aligned_alloc /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:192
    #1 0x7f638b7a68ff in volk_malloc (/usr/lib/libvolk.so.2.5+0x618ff)
    #2 0x7f638b5a0b24 in __libc_start_main (/usr/lib/libc.so.6+0x27b24)

==1924==HINT: if you don't care about these errors you may set allocator_may_return_null=1
SUMMARY: AddressSanitizer: invalid-aligned-alloc-alignment /build/gcc/src/gcc/libsanitizer/asan/asan_malloc_linux.cpp:192 in __interceptor_aligned_alloc
==1924==ABORTING
```
return code is 1

> We generally require signed commits. git commit -s. It adds a Signed off by ... line at the end of your commit message. The link behind our DCO check explains this.

> However, I suggest to write a comment that explains the reason to do this change.

Alright, I'll change that comment and use `-s`. I wrote that PR on the bus through the github web UI since I had to leave but I have access to my PC again now.

AlexandreRouma:
Alright, I can't get -s working. Not sure what I'm doing wrong but it's just not doing anything.

jdemel:
About signed commits:
You'll need to do a force push eventually. In this case I suggest:
```
git rebase -i HEAD~3
```
Then you just squash the 3 commits in your PR into one. Next you'll be asked to edit the commit message. Here you can just clean up all 3 commit messages. In the end, It should look smth like this
```
asan: Fix volk_malloc alignment bug

ASAN etc. explain your reasons here. 
you can even add explanations and hints here. e.g. how you test with ASAN.

Signed-off-by: Name <email>
```
Then just force push these changes. Afterwards the DCO check is satisfied. The reason why it doesn't work at the moment is that all commits are required to be signed. Your first commit isn't though.

Thanks for the additional explanation.

@michaelld would it be a good addition to craft a CI test?

michaelld:
> @michaelld would it be a good addition to craft a CI test?

Yes I'd think so!

AlexandreRouma:
btw, just checked it on a clean docker instance for several distro and that code with ASAN enabled is also broken, so I can confirm it's not just my arch system.

lib/volk_malloc.c
```diff
@@ -50,6 +50,12 @@
 
 void* volk_malloc(size_t size, size_t alignment)
 {
+    // The ASAN address sanitizer requires the allocation size to be a multiple of
+    // the the alignment. Add bytes to size so that it is a multiple of alignment
+    // if it's not.
```
michaelld: Recommend something more like the following for the comment text:
```
Regardless of the changes here to make ASAN (address sanitizer) checking happy, calling `volk_malloc` might result in the allocation of more memory than requested for correct alignment. Any allocation size change here will in general not impact the end result, since initial size alignment is required either way.
```

lib/volk_malloc.c
```diff
@@ -50,6 +50,12 @@
...
+    if (size % alignment) {
```
michaelld, line 61: Please add the following code (correctly formatted) just before this section. We need this before the original code from this PR to make sure the math in the PR is valid. We didn't need it before because we passed any math issues into other functions.
```
if ((size == 0) || (alignment == 0)) {
    fprintf(stderr, "VOLK: Error allocating memory: either size or alignment is 0");
    return NULL;
}
```

michaelld (changes requested):
Minor code changes needed if we're adding in this new math section. Also recommend updating the comment

AlexandreRouma:
Changes made and the commit message seems to show properly, but the DCO seems to still complain.
Sorry if I'm making a mess out of this PR, I'm not used to the "signed off by" stuff, never used it
image

michaelld (approved):
The PR changes LGTM ... I can't see the DSO, so I can't comment on that. Let's see what CI reports, too.

jdemel (approved):
The changes look good to me.

What the DCO checker does is: Go through all commits in this PR, currently 3, and search for a "Signed-off-by: " line. The first of your 3 commits with the message "bugfix" is missing this line. Thus, I suggested to squash those 3 commits and make sure the commit message contains all the required data. We can guide you through the process. I know it can be cumbersome at first.

AlexandreRouma:
Alright I see.
If possible send the git commands I need to run to fix this

jdemel:
@AlexandreRouma I tried to explain the steps earlier. Here I add a few more comments to be more verbose.
I don't want to annoy you with this stuff. We decided to require a DCO with a sign-off to make sure all contributors confirm that they are allowed to contribute their changes. If I'd change the commit messages, I'd feel like I'd forge that. And that would be wrong.

> ```
> git rebase -i HEAD~3
> ```

A text editor will open and you need to pick one commit and set the other 2 to squash, i.e. change the word `pick` to `s`.
Next, you'll be presented with a file with all the commit messages and lots of explanations. Every line that doesn't start with `#` is included in the commit message. Thus, you just need to rewrite these parts and sort them accordingly.

> Then you just squash the 3 commits in your PR into one. Next you'll be asked to edit the commit message. Here you can just clean up all 3 commit messages. In the end, It should look smth like this
> 
> ```
> asan: Fix volk_malloc alignment bug
> 
> ASAN etc. explain your reasons here. 
> you can even add explanations and hints here. e.g. how you test with ASAN.
> 
> Signed-off-by: Name <email>
> ```

Most of the commit message is already there. You just need to sort it and remove duplicates. e.g. You'll probably be presented 2 `Signed-off-by` lines.

> Then just force push these changes. Afterwards the DCO check is satisfied. The reason why it doesn't work at the moment is that all commits are required to be signed. Your first commit isn't though.

With
```
git remote -v
```
you can figure out your remote name. Then with
```
git push -f [YOUR REMOTE NAME] main
```
you force push your changes to this remote repository. Changes will show up automatically in this PR.

AlexandreRouma:
Alright, the DCO is FINALLY happy.
Thanks for the detailed explanation btw!

jdemel:
@AlexandreRouma Thanks a lot!

jdemel:
@AlexandreRouma could you add yourself to the list in [AUTHORS_RESUBMITTING_UNDER_LGPL_LICENSE](https://github.com/gnuradio/volk/blob/main/AUTHORS_RESUBMITTING_UNDER_LGPL_LICENSE.md). This would make it easier for us to track.

AlexandreRouma:
Alright I'll have a look tomorrow, been very busy lately

Issue #174, opened 2018-03-27
purge boost dependency


jdemel:
I'd like to open this issue as a reference. In the 1.3.1/1.4 release announcement was a remark to purge boost from VOLK. I consider this issue a place to discuss the required changes.
A quick `grep -irnI boost .` reveals essentially to files which still contain `boost`.

Affected files:
- `lib/qa_utils.h`
`BOOST_AUTO_TEST_CASE` and `BOOST_CHECK_EQUAL`. Since there is no `#include <boost>`, I assume this is just a naming thing. Though, there is also some CMake magic in `lib/CMakeLists.txt`. I suggest to rename these macros and remove the boost reference in the corresponding CMake file.
- `apps/volk_profile.cc`
`boost::filesystem`. This is probably the tough one. C++11 does not support any filesystem magic. All I could find is a pointer to some experimental C++17 features. I guess C++17 is  way to new for VOLK right now. Boost is only used in the function `write_results` with `boost::filesystem::path`, `boost::filesystem::exists` and `boost::filesystem::create_directories`. I suggest to purge the defined namespace `fs` in order to make it more obvious that Boost is used here. And then we need to find an elegant way to remove these function calls. Maybe someone can come up with a good solution here.

marcusmueller:
There's plenty single-header-file libs that one could include. But you'd basically buy independence from boost with importing someone else's code that might break in some future version. Don't know if that's worth it.

noc0lour:
I'd vote for a single header file (or a implementation written by someone else) instead of including Boost. It's really just two lines for creating the configuration directory structure. That's not a reason to still have boost around. Maybe we can have something which wraps a simple implementation and `<filesystem>` or backport `<filesystem>` (necessary parts) to volk.

marcusmueller:
@noc0lour 

* https://github.com/ZhreShold/zupply/wiki/Filesystem

n-west:
Thanks @jdemel.

Removing boost is 100% worth it because it’s a huge PITA to link against and there are non-gnuradio uses of VOLK including mine which would greatly benefit from dropping boost as a dependency. And yes, the last bit of (used) boost is in creating the config directory.

I would like either conditionalized c++17 support or a simple header only solution that can be dropped in. I haven’t looked at merits of options yet. The reason this is the last bit left is because I didn’t want to deal with the nightmare of cross-platform directory trees myself.

noc0lour:
[...]

PR #548, opened 2021-12-21
CMake: Removed files used by nothing in the build system

marcusmueller:


jdemel (approved):
LGTM. Thanks for this simplification!

At first, I thought about bumping the minimum Boost version. However, Boost is only required on old systems that lack compilers with `std::filesystem` support. A modern system comes with a modern compiler and doesn't require Boost at all.

apps/CMakeLists.txt
```diff
@@ -48,10 +48,7 @@ target_include_directories(volk_profile
 )
 
 if(NOT FILESYSTEM_FOUND)
-    target_include_directories(volk_profile
-        PUBLIC ${Boost_INCLUDE_DIRS}
-    )
-    target_link_libraries(volk_profile PRIVATE ${Boost_LIBRARIES})
+    target_link_libraries(volk_profile PRIVATE Boost::filesystem)
```
michaelld, line 51: Still need to use the `Boost_INCLUDE_DIRS` for includes here

michaelld (changes requested):
1 required change before I'll accept this. Also a meta-discussion, thinking we want to update minimum Boost to something reasonably recent, like even 1.60.0. When was `Boost::filesystem` added? At issue here is that by keeping min version at 1.35, we still need the `Boost_NOGO_VERSIONS` stuff from `VolkBOOST.cmake`. So, either we leave this all alone or we bump the Boost version so-as to not need the NOGO stuff.

marcusmueller:
Re: boost bumping: yes, I'd say we should do that. 1.35 is positively ancient, pulling up with GNU Radio makes sense to me – chances are that non-GR projects that use VOLK have at least the same Boost versions as GR 3.8 could rely on

apps/CMakeLists.txt
marcusmueller: `target_link_libraries` with a config target does add the includes to the target, or am I mistaken?

For example, gnuradio/gr-uhd/lib/CMakeLists.txt: https://github.com/gnuradio/gnuradio/blob/master/gr-uhd/lib/CMakeLists.txt#L34

We `target_link_library` the `gnuradio-uhd` target against the UHD::UDH target, which publicly says "hey, include this", which leads to the target automagically gaining these include dependencies.

jdemel:
I'm OK to bump the minimum Boost version. I just don't see the need to do so. Say, we bump the minimum version to 1.65. In that case, we can assume a reasonable recent compiler to be available. This compiler will support `std::filesystem` and thus, we don't need Boost at all.

We removed the last CI case for Linux, when we dropped tests for Ubuntu 14.04.
AppVeyor CI is the only test where we still add Boost. And we can probably remove that too because these tests rely on ancient toolchains.

Long story short. Maybe it's time to have another look at #174 and go through with it.

jdemel:
If we decide to finally purge Boost, we would merge #557 and close #174 .
PR #557 supersedes this PR because it goes even further. However, we need to decide if it is time to purge Boost.

michaelld:
Boost is purged from #557  ... so, closing this PR as it has been superseded!

michaelld:
OK ... will be added ... soon. Closing this is safe :)

PR #551, opened 2021-12-22
ci: Disable msys2 windows CI

### LINKED ISSUE #477: MinGW CI test fails
We merged support for MinGW/MSYS2 recently in #470 .

Unfortunately, our CI fails now:
https://github.com/gnuradio/volk/runs/2686775368

@Alesha72003 would you be willing to help here?

jdemel:
This test run fails consistently for quite a while now. So far we
couldn't find anyone to fix it. Disabling it.

Fix #477

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -191,39 +191,39 @@ jobs:
     - name: test
       run: cd build && ctest -V -C Release
 
-  build-windows-msys2:
-    name: Build on windows-latest using MinGW and MSYS2
+  # build-windows-msys2:
+  #   name: Build on windows-latest using MinGW and MSYS2
     
-    runs-on: windows-latest
-    steps:
-      - uses: msys2/setup-msys2@v2
-        with:
-          update: true
-          install: >-
-            base-devel
-            git
-            mingw-w64-x86_64-boost
-            mingw-w64-x86_64-gcc-libs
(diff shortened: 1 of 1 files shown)


michaelld (approved):
Disabling this test makes sense. Discussion below for posterity & maybe someone else knows how to fix this issue based on the below?

[the error here is](https://github.com/gnuradio/volk/runs/4607213216?check_suite_focus=true), for example:
```
/D/a/_temp/msys64/usr/bin/python3: can't open file '/D/a/volk/volk/build/lib/D:/a/volk/volk/gen/volk_tmpl_utils.py': [Errno 2] No such file or directory
```
and the Py command should actually be just
```
D:/a/volk/volk/gen/volk_tmpl_utils.py
```
... so somehow some extra cruft is being prepended to the string `gen/volk_tmpl_utils.py` ([as found here, for example](https://github.com/gnuradio/volk/blob/main/lib/CMakeLists.txt#L95-L100)). Note that every time these Python scripts are called this error happens, but for some reason during configuration the error is ignored & that stage finishes.

Guessing that the CMake variable `PROJECT_SOURCE_DIR` might be being set incorrectly, but that's the correct variable to be using, and hence if this is the issue then maybe we just need a newer CMake version?

IDK how to fix this. Maybe there's updates to GH workflows that would help, but I'm not expert here so not sure what to do.

PR #552, opened 2021-12-29
docs: Update docs

jdemel:
Doxygen generated docs need an update. e.g. we don't want cpu_features
docs included in our docs. Also, the structure needs an update and we
include our awesome logo now too.


michaelld (approved):
LGTM

PR #555, opened 2022-01-07
Add deprecations for untested kernels

### LINKED ISSUE #549: deprecate untested kernels
VOLK includes kernels that are untested.

https://github.com/gnuradio/volk/blob/0e844c6115f8e2cda5646a45f6162482c32eef8d/lib/kernel_tests.h#L169-L178

- volk_16i_x5_add_quad_16i_x4 
- volk_16i_branch_4_state_8
- volk_16i_max_star_16i
- volk_16i_max_star_horizontal_16i
- volk_16i_permute_and_scalar_add
- volk_16i_x4_quad_max_star_16i

I suggest to deprecate them and remove them in a future VOLK version. However, I suggest to check if GNU Radio uses them anywhere. If these kernels are still in use, we should re-introduce them into our CI and make sure they properly describe what they're doing.

jdemel:
We have six kernels:
- `volk_16i_x5_add_quad_16i_x4`
- `volk_16i_branch_4_state_8`
- `volk_16i_max_star_16i`
- `volk_16i_max_star_horizontal_16i`
- `volk_16i_permute_and_scalar_add`
- `volk_16i_x4_quad_max_star_16i`

that we don't test. They should be marked deprecated. If nobody complains, we can remove them in a future VOLK version. Otherwise, we need proper tests for those kernels because they are currently untested.

We assume that MSVC is the compiler on Windows and thus, we don't add deprecation warnings. Other platforms receive these warnings though. Ideas on how to improve these warnings are appreciated.

Fix #549

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -476,7 +476,7 @@ list(APPEND volk_sources ${CMAKE_CURRENT_BINARY_DIR}/constants.c)
 ########################################################################
 if(NOT WIN32)
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fvisibility=hidden")
-    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fvisibility=hidden")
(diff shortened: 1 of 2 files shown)


michaelld (approved):
Agreed! LGTM

PR #557, opened 2022-01-20
boost: Remove boost dependency

### LINKED ISSUE #174: purge boost dependency

jdemel:
VOLK used Boost in the past quite heavily. We reduced Boost usage since then. At the moment, we only require `boost::filesystem` for very old systems. For all recent systems we rely on `std::filesystem` or `std::experimental::filesystem`.

The latest Ubuntu version that still required Boost was 14.04. Starting with Ubuntu 16.04, we can use `std::experimental::filesystem` and later `std::filesystem`.

Fix #174


jdemel:
So far, all CI tests pass without Boost. However, AppVeyor complains:
```
C:\projects\volk\kernels\volk/volk_32f_sin_32f.h(515): error C3861: '_mm512_floor_ps': identifier not found 
```
This is just an example. It seems like VS 2017 doesn't like to compile AVX512 because it doesn't support it although CMake reports its availability. Thus, we try to move to VS 2019 on AppVeyor as well.

marcusmueller:
> So far, all CI tests pass without Boost. However, AppVeyor complains:
> 
> ```
> C:\projects\volk\kernels\volk/volk_32f_sin_32f.h(515): error C3861: '_mm512_floor_ps': identifier not found 
> ```

> This is just an example. It seems like VS 2017 doesn't like to compile AVX512 because it doesn't support it although CMake reports its availability. Thus, we try to move to VS 2019 on AppVeyor as well.

This seems to be a separate issue and should be tracked separately, I guess?

jdemel:
> > So far, all CI tests pass without Boost. However, AppVeyor complains:
> > ```
> > C:\projects\volk\kernels\volk/volk_32f_sin_32f.h(515): error C3861: '_mm512_floor_ps': identifier not found 
> > ```
> 
> > This is just an example. It seems like VS 2017 doesn't like to compile AVX512 because it doesn't support it although CMake reports its availability. Thus, we try to move to VS 2019 on AppVeyor as well.
> 
> This seems to be a separate issue and should be tracked separately, I guess?

Yes, this is a separate issue.

michaelld (changes requested):
Top-level CMakeLists.txt:29 ... let's just change the CXX_STANDARD to 17 ... no reason to bother with 11 any longer!

CMakeLists.txt
```diff
@@ -195,13 +177,8 @@ if(${FILESYSTEM_FOUND})
     set(CMAKE_CXX_STANDARD 17)
```
michaelld: Per my Request: Let's just move these up to the STANDARD definitions ; that simplifies the current `if/else` to just `if(NOT FILESYSTEM_FOUND` ... no need to `${}` around it either, BTW

jdemel: Agreed. Fixed it.

michaelld (approved):
LGTM

PR #558, opened 2022-01-22
cpu_features: Make cpu_features submodule optional

### LINKED ISSUE #556: Please allow using external cpu_features package
Bundling and installing ```cpu_features``` causes conflicts with pre-installed package.

Version: 2.5.0

jdemel:
Since cpu_features is available as a package in recent distributions, we'd like to use the distro provided package.

We try to find the CMake CpuFeatures package and build against it, if we can't find one, we try to use the git submodule. If both options fail, we emit an error message.

Fix #556


michaelld (approved):
LGTM on reading, no testing. If something isn't quite correct I'm sure someone will let us know ;) I like the idea of allowing cpu_features to be installed separately ... always my preference when possible!

PR #561, opened 2022-02-09
zenodo: Re-organize .zenodo.json

jdemel:
Instead of a solely alphabetical order, we introduce three categories.
The names of the current maintainers go first.
Next, all contributors where full names (first and last) are available.
Finally, all contributors where we only have GitHub handles are listed.


michaelld (approved):
LGTM

PR #565, opened 2022-04-09
Fix out-of-bounds reads

argilo:
Several kernels pre-load data for an upcoming loop iteration. This leads to out-of-bounds reads on their final iterations. I've fixed this by moving the reads to the beginning of their loops.

Out-of-bounds reads can be detected with valgrind, but only after the `vlen_twiddle` value is changed from 5 to 0 to prevent extra space from being allocated in test buffers:

https://github.com/gnuradio/volk/blob/ea0ca360e5e22fe4c9ae07a14f7f63407ec761cc/lib/qa_utils.cc#L531-L534

After updating the kernel, I verified that no out-of-bounds reads occur, and that the kernels still behave correctly, by running the following command:

```
for i in {128..144}; do valgrind apps/volk_profile -n -v $i -i 1 | grep -E 'fail|offset'; done
```

I did not see any noticeable performance difference between the old & new versions.

Signed-off-by: Clayton Smith <argilo@gmail.com>


jdemel:
Thanks for fixing these. Did you run the updated kernels with different vector lengths?

The dot_prod kernel is a nightmare indeed.

argilo:
Yes. I tested all kernels with vector lengths 128-144 (using valgrind + `vlen_twiddle = 0`).

michaelld (approved):
LGTM. The non-dot_prod changes I can follow & those make sense. The dot_prod ones are pretty non-obvious so I'll trust your testing for code functionality.

michaelld:
Question: At least for the non-dot_prod changes, do we really get much benefit from the preloading commands? The whole point of preloading is to do it while other stuff is going on, a little before the data is required. While I agree with these changes so-as to not be accessing out of bounds in arrays, it also means that there is almost no code separating the preload and actual use. I'd be curious the timing comparison between this version & one where the code just directly uses the arguments -- no preloading.

argilo:
> Question: At least for the non-dot_prod changes, do we really get much benefit from the preloading commands?

No idea. I know very little about SIMD or low-level optimization. My goal was just to get rid of the out-of-bounds reads.

argilo:
It looks like it's not possible to operate directly on memory, so the `load` operations are required.

jdemel:
The `load` and `store` operations are necessary. It'd be great to have a tool to benchmark and visualize all kernels to give users a feeling for how their systems might perform.
At least for GCC with `-O3`, I observed that GCC reorders the assembly instructions with respect to the intrinsics if this is beneficial. I wouldn't argue with the compiler.
I doubt that it is beneficial to preload data. Essentially, we'd try to be smarter than all the CPU optimizations which would imply we know what a specific CPU does. At least CPPCon presentations suggest to assume otherwise.

jdemel (approved):
LGTM.

PR #567, opened 2022-04-14
ci: Update recursive git clone

jdemel:
The GH Action to clone `checkout` supports submodule clones now. We should use it
instead of a separate GH action that just broke.


jdemel:
Merging this change to make sure our CI works.

PR #568, opened 2022-04-23
Fix broken neon kernels

argilo:
Running `volk_profile` with various vector lengths (e.g. `for i in {16..31}; do echo $i; for j in {1..1000}; do apps/volk_profile -n -v $i -i 1 | grep -E 'fail|offset|^  '; done; done` on a Raspberry Pi shows that some Neon kernels are broken:

* `volk_32f_x3_sum_of_poly_32f_neonvert` ignores trailing values completely.
* `volk_32fc_index_min_32u_neon` writes the wrong value to the running minimum while handling trailing values, forgetting to compute the squared magnitude.
* `volk_32fc_index_max_32u_neon` writes the wrong starting value to the running maximum, and writes the wrong value to the running maximum while handling trailing values, forgetting to compute the squared magnitude.

I've fixed those problems here.

Signed-off-by: Clayton Smith <argilo@gmail.com>

### DIFF
--- a/kernels/volk/volk_32f_x3_sum_of_poly_32f.h
+++ b/kernels/volk/volk_32f_x3_sum_of_poly_32f.h
@@ -493,8 +493,8 @@ volk_32f_x3_sum_of_poly_32f_neonvert(float* __restrict target,
     float thrd = 0.0;
     float frth = 0.0;
 
-    for (i = 4 * num_points / 4; i < num_points; ++i) {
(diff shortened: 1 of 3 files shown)


jdemel (approved):
LGTM.

The whole `if (num_points) ...` business looks sketchy but this should be part of another PR.

michaelld (approved):
LGTM ; thx!

PR #573, opened 2022-05-26
Added MIPS support to volk_cpu.tmpl.c.

michael-roe:


jdemel:
If we merge this PR, we enable VOLK on another architecture. We don't have any CI for that arch. Thus, we need to be clear that this is experimental.

michaelld (commented):
Is the intent here to just allow building VOLK on MIPS via generic kernels? If so, can you provide the output of running `make test` (from the cmake build directory) as well as `volk_profile` run after installing (actually the file `~/.volk/volk_config`)?

michaelld:
I'd say we should enable PPC in the same way & also tag it experimental.

jdemel:
How and where should we communicate that MIPS support is experimental?

jdemel:
I suggest to add a comment to the [top-level `README.md` file under Hardware architectures](https://github.com/gnuradio/volk#hardware-architectures).

michael-roe:
Even without this patch, we can build on MIPS.

Some kernels don't work on MIPS, because of the ABI issue with complex numbers.

The intent of this patch is to get in place most of the machinery you would need in order to have kernels that make use of MIPS Application Specific Extensions. (I wanted this for an experiment, and this part of the experiment seemed of sufficiently general usefulness as to be worth upstreaming).

Documenting MIPS support as experimental seems appropriate.

michaelld:
> top-level `README.md` file under Hardware architectures

Agreed! Maybe something like: ```Currently VOLK aims to run with optimized kernels on x86 with SSE/AVX and ARM with NEON. VOLK has experimental support for PPC and MIPS; we are happy to accept contributions to make these architectures less experimental.```

michael-roe:
That seems ok to me. Do you want me to include that text in a revised version of this PR, or can it just go in as a separate commit?

jdemel:
It'd be great if you made these changes. A separate commit for these changes seems appropriate.

balister:
I'm pretty sure volk QA passes on risc-v. I've been thinking about adding qemu riscv builds to the sausage machine.

michael-roe:
It would be good if we ran QA on risc-v.

I think it will fail because of the issue to do with passing a C++ complex as an argument to a function that expects a C complex.

jdemel:
Merging this PR now, since #609 contains the documentation change that should make it clear that MIPS support is experimental.

PR #575, opened 2022-05-29
Fix float to int conversion.

michael-roe:
This fixes the bug in volk_32f_s32f_convert_32i in which input values greater than or equal to (float) INT_MAX led to undefined behavior.

This is my second attempt at a PR for this bug; please ignore the previous one.

### DIFF
--- a/kernels/volk/volk_32f_s32f_convert_32i.h
+++ b/kernels/volk/volk_32f_s32f_convert_32i.h
@@ -237,18 +237,19 @@ static inline void volk_32f_s32f_convert_32i_generic(int32_t* outputVector,
 {
     int32_t* outputVectorPtr = outputVector;
     const float* inputVectorPtr = inputVector;
-    unsigned int number = 0;
-    float min_val = INT_MIN;
-    float max_val = INT_MAX;
-    float r;
+    const float min_val = (float)INT_MIN;
+    const float max_val = (float)INT_MAX;
 
-    for (number = 0; number < num_points; number++) {
-        r = *inputVectorPtr++ * scalar;
-        if (r > max_val)
-            r = max_val;
+    for (unsigned int number = 0; number < num_points; number++) {
+        const float r = *inputVectorPtr++ * scalar;
+        int s;
+        if (r >= max_val)
+            s = INT_MAX;
         else if (r < min_val)
-            r = min_val;
-        *outputVectorPtr++ = (int32_t)rintf(r);
+            s = INT_MIN;
+        else
+            s = (int32_t)rintf(r);
+        *outputVectorPtr++ = s;
     }
 }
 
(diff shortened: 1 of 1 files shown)


jdemel:
Thanks for your PR!

Should #574 be closed because this PR supersedes it?

kernels/volk/volk_32f_s32f_convert_32i.h
```diff
@@ -431,14 +434,17 @@ static inline void volk_32f_s32f_convert_32i_a_generic(int32_t* outputVector,
     float min_val = INT_MIN;
     float max_val = INT_MAX;
     float r;
+    int s;
```
jdemel: The generic version doesn't benefit from an aligned version. We don't do anything fancy with aligned vectors besides choosing an implementation that expects aligned vectors.

I suggest to replace the function body with a call to the generic version. I'd support to remove this implementation entirely. However, we install all kernel headers and thus some users might use this function directly.

kernels/volk/volk_32f_s32f_convert_32i.h
```diff
@@ -241,14 +241,17 @@ static inline void volk_32f_s32f_convert_32i_generic(int32_t* outputVector,
     float min_val = INT_MIN;
     float max_val = INT_MAX;
     float r;
```
jdemel: This implementation would benefit from quite a bit of clean-up. Would you be willing to include this in your PR?

I suggest to move the loop variable declaration `number` into the for loop.
The `float r` declaration could go into the loop as well. `min_val` and `max_val` should be `const`. 
The `int s;` declaration is only necessary inside the loop. Thus, I suspect it would be better to move it into this scope.

jdemel (commented):
Thanks for your PR. It looks good. Though, I have a few suggestions to improve it.

kernels/volk/volk_32f_s32f_convert_32i.h
michael-roe: Ok, I can do that.

I agree, having two copies of the same code is a bad idea (which caused me to miss fixing one of the copies in the previous attempt); one should call the other.

michael-roe:
Ok, I've made those changes. Does it look ok now?

kernels/volk/volk_32f_s32f_convert_32i.h
```diff
@@ -237,18 +237,20 @@ static inline void volk_32f_s32f_convert_32i_generic(int32_t* outputVector,
...
-    float max_val = INT_MAX;
-    float r;
+    const float min_val = (float)INT_MIN;
+    const float max_val = (float)INT_MAX;
 
-    for (number = 0; number < num_points; number++) {
+    for (unsigned int number = 0; number < num_points; number++) {
+        float r;
+        int s;
         r = *inputVectorPtr++ * scalar;
```
jdemel: One last comment. I suggest to merge line 244 with 246.

```suggestion
        const float r = *inputVectorPtr++ * scalar;
```
I'd declare it `const` as well. This helps others understand the code better.

jdemel (commented):
Thanks a lot for your work! One last comment and things are ready.

jdemel:
Our C++ formatting CI complains about one line. That should be trivial to fix.

michael-roe:
Ok, done.

jdemel (approved):
LGTM. Thanks for working on all my suggestions too.

PR #578, opened 2022-07-01
Fix CMake Python path calculation

### LINKED ISSUE #579: CMake warnings bulding
After recursively cloning volk into /tmp, I ran this command:
```bash
cd /tmp/volk ; mkdir build ; cd build
cmake -DCMAKE_COLOR_MAKEFILE=NO \
   -DCMAKE_PREFIX_PATH=/opt/foss \
   -DCMAKE_INSTALL_PREFIX=/opt/foss \
   -DCMAKE_BUILD_TYPE=Release \
    -S ..
CMake Warning at CMakeLists.txt:140 (find_package):
  By not providing "FindCpuFeatures.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "CpuFeatures", but CMake did not find one.

  Could not find a package configuration file provided by "CpuFeatures" with
  any of the following names:

    CpuFeaturesConfig.cmake
    cpufeatures-config.cmake

  Add the installation prefix of "CpuFeatures" to CMAKE_PREFIX_PATH or set
  "CpuFeatures_DIR" to a directory containing one of the above files.  If
  "CpuFeatures" provides a separate development package or SDK, be sure it
  has been installed.
```

It doesn't seem to affect building, but the warnings should be silenced.

jdemel:
Apparently, the Python install path calculation is broken on some new systems, e.g. Fedora 36 and Ubuntu 22.04.

These are two approaches to fix it.
https://github.com/gnuradio/gnuradio/pull/5979
https://github.com/gnuradio/gnuradio/pull/5981

I went with the approach in https://github.com/gnuradio/gnuradio/pull/5981 because it fits better into how we handle things. e.g. the result is a relative path. We use this approach in other places as well.

### DIFF
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -112,17 +112,26 @@ execute_process(
     COMMAND ${PYTHON_EXECUTABLE} -c "import os
 import sysconfig
 import site
+
 install_dir = None
 prefix = '${CMAKE_INSTALL_PREFIX}'
-#use sites when the prefix is already recognized
+
+#use `site` when the prefix is already recognized
 try:
   paths = [p for p in site.getsitepackages() if p.startswith(prefix)]
   if len(paths) == 1: install_dir = paths[0]
 except AttributeError: pass
+
 if not install_dir:
-    #find where to install the python module
-    install_dir = sysconfig.get_path('platlib')
-    prefix = sysconfig.get_config_var('prefix')
+    # find where to install the python module
+    # for Python 3.11+, we could use the 'venv' scheme for all platforms
+    if os.name == 'nt':
+        scheme = 'nt'
+    else:
+        scheme = 'posix_prefix'
+    install_dir = sysconfig.get_path('platlib', scheme)
+    prefix = sysconfig.get_path('data', scheme)
+
 #strip the prefix to return a relative path
 print(os.path.relpath(install_dir, prefix))"
     OUTPUT_STRIP_TRAILING_WHITESPACE
--- a/cmake/Modules/VolkPython.cmake
+++ b/cmake/Modules/VolkPython.cmake
@@ -114,17 +114,23 @@ import sysconfig
 import site
 
 install_dir = None
+# The next line passes a CMake variable into our script.
 prefix = '${CMAKE_INSTALL_PREFIX}'
 
-#use `site` when the prefix is already recognized
+# We use `site` to identify if our chosen prefix is a default one.
+# https://docs.python.org/3/library/site.html
 try:
-  paths = [p for p in site.getsitepackages() if p.startswith(prefix)]
-  if len(paths) == 1: install_dir = paths[0]
+    # https://docs.python.org/3/library/site.html#site.getsitepackages
+    paths = [p for p in site.getsitepackages() if p.startswith(prefix)]
+    if len(paths) == 1: install_dir = paths[0]
 except AttributeError: pass
 
+# If we found a default install path, `install_dir` is set.
 if not install_dir:
-    # find where to install the python module
-    # for Python 3.11+, we could use the 'venv' scheme for all platforms
+    # We use a custom install prefix!
+    # Determine the correct install path in that prefix on the current platform.
+    # For Python 3.11+, we could use the 'venv' scheme for all platforms
+    # https://docs.python.org/3.11/library/sysconfig.html#installation-paths
     if os.name == 'nt':
         scheme = 'nt'
     else:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,7 +137,7 @@ else()
 endif()
 
 if (VOLK_CPU_FEATURES)
-  find_package(CpuFeatures)
+  find_package(CpuFeatures QUIET)
   if(NOT CpuFeatures_FOUND)
     message(STATUS "cpu_features package not found. Requiring cpu_features submodule ...")
     if(NOT EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/cpu_features/CMakeLists.txt" )


cmake/Modules/VolkPython.cmake
```diff
@@ -112,17 +112,26 @@ execute_process(
...
-    install_dir = sysconfig.get_path('platlib')
-    prefix = sysconfig.get_config_var('prefix')
+    # find where to install the python module
+    # for Python 3.11+, we could use the 'venv' scheme for all platforms
+    if os.name == 'nt':
+        scheme = 'nt'
+    else:
+        scheme = 'posix_prefix'
+    install_dir = sysconfig.get_path('platlib', scheme)
+    prefix = sysconfig.get_path('data', scheme)
```
michaelld, line 139: Thinking this should be `prefix`, not `data`, per the prior code ... yes?

michaelld (changes requested):
Minor tweak otherwise LGTM

cmake/Modules/VolkPython.cmake
jdemel: On Ubuntu 22.04 the options for the first parameter are:
```
>>> sysconfig.get_path_names()
('stdlib', 'platstdlib', 'purelib', 'platlib', 'include', 'scripts', 'data')
```
The results are:
```
path=stdlib              /usr/lib/python3.10
path=platstdlib          /usr/lib/python3.10
path=purelib             /usr/lib/python3.10/site-packages
path=platlib             /usr/lib/python3.10/site-packages
path=include             /usr/include/python3.10
path=scripts             /usr/bin
path=data                /usr
```
with scheme `posix_prefix`. Further, with `data` and `platlib` the resulting relative paths are:
```
scheme=deb_system          lib/python3/dist-packages
scheme=nt                  Lib/site-packages
scheme=nt_user             Python/site-packages
scheme=osx_framework_user  lib/python/site-packages
scheme=posix_home          lib/python
scheme=posix_local         lib/python3.10/dist-packages
scheme=posix_prefix        lib/python3.10/site-packages
scheme=posix_user          lib/python3.10/site-packages
```
Thus, `scheme='posix_prefix'` seems to be fitting. However, it seems like we might want to opt for `scheme='posix_local'` because on Ubuntu 22.04:
```
>>> import sys
>>> sys.path
[...,
'/usr/lib/python310.zip', 
'/usr/lib/python3.10', 
'/usr/lib/python3.10/lib-dynload', 
'/usr/local/lib/python3.10/dist-packages', 
'/usr/lib/python3/dist-packages']
```
On Ubuntu 20.04:
```
[...
'/usr/lib/python38.zip', 
'/usr/lib/python3.8', 
'/usr/lib/python3.8/lib-dynload', 
'/usr/local/lib/python3.8/dist-packages', 
'/usr/lib/python3/dist-packages']
```
Then again, in a `fedora:36` Docker container:
```
[...
'/usr/lib64/python310.zip', 
'/usr/lib64/python3.10', 
'/usr/lib64/python3.10/lib-dynload', 
'/usr/lib64/python3.10/site-packages', 
'/usr/lib/python3.10/site-packages']
```
Basically, we want that Python files get installed into the correct folder for the default `CMAKE_INSTALL_PREFIX` which is `/usr/local`. The installation should work without modifying the `PYTHONPATH` variable.
The variables are: 
- `lib` vs `lib64`
- `dist-packages` vs `site-packages`
- `python` vs `python3` vs `python3.x`.
It seems like our variable choice works for Fedora 36+ at least:
https://docs.fedoraproject.org/en-US/fedora/latest/release-notes/developers/Development_Python/

ryanvolz:
I'll try to synthesize everything I've come to know about this.

There are two primary cases that this needs to handle:

1. Installation to the default `CMAKE_INSTALL_PREFIX`, and have the Python modules found by the default system Python without any modification of the `PYTHONPATH`.
2. Installation to a custom `CMAKE_INSTALL_PREFIX` (e.g. PYBOMBS, conda), where the user knows what they are doing or is using some other mechanism to add to the `PYTHONPATH`.

Because Debian/Ubuntu, Fedora, and perhaps other distributions do funny things with where they want Python modules to be installed by default (e.g. `dist-packages` vs. `site-packages`, `lib` vs. `lib64`, `python` vs. `python3` vs. `python3.x`), we've concluded that both of those cases require separate handling.

Currently, the first case is handled by the
```
try:
  paths = [p for p in site.getsitepackages() if p.startswith(prefix)]
  if len(paths) == 1: install_dir = paths[0]
```
portion of the code. So if you're trying to install to a prefix like `/usr/local` that is already searched for Python modules, it will go where the system wants it.

The second case is what's handled by the following portion of the code, with the calls to `sysconfig.get_path('platlib', scheme)`. And it's because that code is handling this second case that we explicitly *need* to use the `posix_prefix` or `nt` scheme. The reason is that we want this to be wherever a plain Python installation would put its modules in a custom prefix, and we don't want to get the distribution-specific modifications to `dist-packages` or whatever. This is why the `venv` scheme has been implemented for Python 3.11+, and why we really should switch to that if it is available, because it ensures a plain scheme that has not been modified by the distribution.

So in conclusion, this PR should work for those cases as it currently stands.

***

Were I to rewrite this from scratch knowing all of that now, I would do something like this (with lots of comments to explain everything that I just detailed above):

1. Call `sysconfig.get_path('platlib')` to get the default module installation location for the Python that is currently being used.
2. Check to see if that path is in the desired prefix. If it is, use that path. (This handles the first case above).
3. If the default path is not in the desired prefix, ask Python to give us an appropriate path while providing the desired prefix. This would involve setting `scheme='venv'` if it exists or falling back to `scheme='nt'` or `scheme='posix_prefix'` as appropriate for the platform. Then you can call `sysconfig.get_path('platlib', scheme=scheme, vars={'base': prefix, 'platbase': prefix})` to get the path within the desired prefix. (This handles the second case above).
4. Finally, you can transform the `install_dir` from an absolute path to a relative path with `os.path.relpath(install_dir, prefix)` before returning.

jdemel:
@ryanvolz Thanks for your detailed answer!

I updated my PR with more comments. I hope this is a good step towards a more long term solution to this recurring problem.

I hope it is sufficient to not use `vars=...` because this option is difficult to handle as well and needs more documentation.

cmake/Modules/VolkPython.cmake
michaelld: Wow this is a lot more complicated than I thought it would be! Myu vote would be `posix_local` as that seems better overall. Each OS will have some specific default, which is fine so long as the OS is consistent ... which is what we want here, after all.

michaelld: So long as we provide a way to override the install location that's the bottom line for me. Having a consistent default is good, of course!

michaelld:
@jdemel do you think this PR is ready for a final review? Let me know & I'll get it done.

FYI please add that this closes https://github.com/gnuradio/volk/issues/579

cmake/Modules/VolkPython.cmake
jdemel: I suggest to use `posix_prefix` and be consistent with GR. Thus, a non-standard install path uses `site-packages`. That would also be the case with Python 3.11+ and `venv`.

michaelld: @jdemel OK ... so the code in the PR as is, no further changes? yes?

jdemel: Yes. I'd leave this PR as is.

michaelld (approved):
LGTM ; let's get this live to see how well it works!

michaelld:
Good discussion everyone! This PR passes QA/CI & it's simple enough to fix if there are issues down the road for some specific OS. Merging.

PR #580, opened 2022-07-23
files: Re-arrange files to clean-up top-level folder

jdemel:
The top-level folder has collected a lot of files. This is an effort to move some into subfolder. We aim to continue this effort with more commits.

I'm sure this new structure is worth discussing. Maybe some files ended up int the wrong folder. Maybe some should stay in the top-level folder.

PR #582, opened 2022-07-23
doxygen: Move doxygen files into docs

jdemel:
This is a first stab at re-arranging doxygen files to have fewer files in our top-level directory.

This includes all `*.md` files as well. We might want to exclude some of them.


CMakeLists.txt
```diff
@@ -209,16 +209,7 @@ endif(ENABLE_ORC)
...
-        ${CMAKE_SOURCE_DIR}/Doxyfile.in
-        ${CMAKE_BINARY_DIR}/Doxyfile
-    @ONLY)
-
-    add_custom_target(volk_doc
-        ${DOXYGEN_EXECUTABLE} ${CMAKE_BINARY_DIR}/Doxyfile
-        WORKING_DIRECTORY ${CMAKE_BINARY_DIR}
-        COMMENT "Generating documentation with Doxygen" VERBATIM
-    )
+  add_subdirectory(docs)
```
michaelld: Let's move this whole section, from `find_package(Doxygen)` to `endif(DOXYGEN_FOUND)` into the subdirectory. That's better CMake programming: only check in the current CMakeLists.txt file for stuff that's required within it, and don't rely on previous checks that may or not have taken place.

michaelld (changes requested):
Minor tweak & this will be good to go. I need to learn how to edit PRs ... not tonight though

CMakeLists.txt
jdemel: I agree. Are there actually any resources that teach you "good CMake"?

jdemel:
Should some of the `*.md` files be excluded?

CMakeLists.txt
michaelld: Hmmm ... good question ... IDK .. mine is via experience, having used CMake now since GR switched, ages ago; and, working on tons or projects that use CMake

michaelld:
> Should some of the `*.md` files be excluded?

Not sure I follow this comment for this PR? Maybe you meant #580 ?

jdemel:
If we'd merge #580 (and I assume we do now), then my comment is more relevant. I'll wait for the other things to happen and update this PR accordingly. I assume I mixed up some files.

michaelld:
OK #580 is merged ... once this PR's code is tweaked per my request then I'm good with it. @jdemel if there are other changes needed post-580 then please go ahead & push commits for them for us to review.

CMakeLists.txt
jdemel: I moved it.

jdemel:
Some `.md` files that I consider to be clutter are included in the docs at the moment.

I feel like these should not be included:
- top-level `README.md`
- docs/AUTHORS_RESUBMITTING_UNDER_LGPL_LICENSE.md`

e.g. `docs/versioning.md` were not part of our docs build before. But I assume these fit nicely in the docs as well.

michaelld (approved):
LGTM now; thx!

michaelld:
The top-level `README.md` is required for best GitHub presentation, so that needs to stay there. I'm OK with `AUTHORS_RESUBMITTING_UNDER_LGPL_LICENSE.md` and `versioning.md` being in `docs` since they fit there. Once this PR is merged, I think we're at a minimum number of extraneous files at the top-level ... do you agree?

michaelld:
I mean those extra `.md` files in `docs` are just along for the ride; we don't build them into the actual Doxygen docs beyond links, do we?

michaelld:
If you feel some of those files need to be moved & it makes sense to put them into this PR to go for it. Otherwise, I'd say let's merge this one & get the next round going.

jdemel:
I want these files to stay where they are. I just don't want to integrate them into the Doxygen build. I'll exclude them.

jdemel:
Now, this PR should be ready to be merged. @michaelld I leave this to you since it is my PR.

michaelld (approved):
LGTM ; merging

PR #584, opened 2022-07-23
spdx: Update cmake for with SPDX license identifier

jdemel:
All appropriate files in the `cmake` folder hold an identifier now. The `msvc` and `Toolchains` folders are skipped.

This commit includes new license headers for some files because they were missing. Copyright data is derived from the git log in these cases.


michaelld (approved):
LGTM

PR #585, opened 2022-07-23
spdx: Update gen and include folder with SPDX license identifier

jdemel:
All Python files in the `gen` folder hold an SPDX license identifier now. The XML files are omitted.

In the `include` folder only the VOLK GPL identifier is updated. Additional licenses in some files are unchanged. These files might need more work in their license identifiers.


michaelld (approved):
LGTM

PR #587, opened 2022-07-23
spdx: Update tmpl and top-level folder with SPDX identifiers

jdemel:
Now, all files in the tmpl and top-level folder hold SPDX license identifiers. Essentially, only the top-level CMakeLists.txt needs this identifier.


michaelld (approved):
LGTM

Issue #589, opened 2022-08-01
Building Volk on Android with NEON Support. volk_32fc_convert_16ic_neon does not compile


Aang23:
In the past, while attempting to build Volk with the Android NDK, NEON detection would not work properly, so I had to patch the CMake manually for it to work, and usually compromise with no NEON support.

Thankfully, this all appears to be fully working now (may also be due to newer NDKs not being as weird...). I was able to build Volk with full SIMD support for all 4 archs, so that's awesome news!

Though, volk_32fc_convert_16ic_neon _VOLK_ASM calls fail at compile-time, so I had to disable the NEON variant for things to work. This is the only kernel that does not build properly with the NDK.

![Screenshot from 2022-08-01 14-16-15](https://user-images.githubusercontent.com/23421917/182145551-73f21200-84c8-42da-be64-66e8d07d991b.png)

Edit : This appears to be an Android-specific issue. Building on another armeabi-v7a machine running a standard GCC-based toolchain works as expected.
Maybe just disable building this kernel if `__ANDROID__` is set. Personally I'd prefer that over having to patch it.

jdemel:
This is a special macro defined here:
https://github.com/gnuradio/volk/blob/af69c60e908cd64b3bacd4ee6b4ce4a52b711865/kernels/volk/volk_32fc_convert_16ic.h#L166-L179

Furthermore, it is only used here:

https://github.com/gnuradio/volk/blob/af69c60e908cd64b3bacd4ee6b4ce4a52b711865/kernels/volk/volk_32fc_convert_16ic.h#L216-L217

Afterwards an
https://github.com/gnuradio/volk/blob/af69c60e908cd64b3bacd4ee6b4ce4a52b711865/kernels/volk/volk_32fc_convert_16ic.h#L237
follows.

At the moment, we don't have Android CI. Thus, it'd be difficult to maintain a change. 

It seems like this change got introduced in https://github.com/gnuradio/volk/commit/d5deb3a5ec2f3744ff48a9fbceb464ccf165f1ef . Before that, this macro didn't exist.

I'd like to narrow down the issue. Could you provide a bit more info?
- Which Android version do you target? Since this is ARMv7 (or should be at least)
- Which hardware do you target?
- Which compiler do you use? Maybe we can improve this section with smth better that wasn't (widely) available 3 years ago.

Aang23:
If you want, I could PR an Android CI configuration. It's something I already have for other projects so it would not be too hard.

[...]

PR #590, opened 2022-08-03
Add Android CI

Aang23:
This PR adds CI compilation checks on Android NDK, for all 4 supported and required architectures.

Of course, armeabi-v7a currently fails, as I didn't think addressing this issue was to be done in this PR.

Initially mentioned in https://github.com/gnuradio/volk/issues/589


jdemel:
Thanks for the PR. It's nice to see that things compile in most cases. I'm still curious. Would it be possible to run our tests as well? I assume that'd require smth like our `aarch64` test approach.

Aang23:
It should be possible with a bit of work - gonna see if I can figure it out. But at the same time, clang and the same architectures are already being tested on Linux, so it is highly unlikely it would pass there and break on Android.

jdemel:
Well, #589 came up because of an Android specific issue.
Also, out of these for builds `armeabi-v7a`, `arm64-v8a`, `x86`, `x86_64`, we only test `arm64-v8a` and `x86_64` thoroughly. The other two are neglected.

.github/workflows/android_build.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  build:
+    name: Build on Android NDK ${{ matrix.arch.name }}
+
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: 
+          - { name: armeabi-v7a }
```
jdemel: As long as this test fails, I suggest to add smth like `allow-failure`. Maybe `continue-on-error` is the GH Action equivalent of the TravisCI `allow-failure` option.

Aang23:
Good point, though it was a compile-time issue.
I've done some (relatively) quick research and testing, and it does not appear to be easy at all to run tests without running some sort of Android emulator / device. Maybe someone else will have an idea at some point. You're not really supposed to run actual executable on Android after all...

I fixed up another another issue I found when building static libs.

jdemel (approved):
LGTM. This PR would be a nice addition to test VOLK on more platforms.

michaelld (approved):
LGTM ; thx!

PR #591, opened 2022-08-07
spdx: Update SPDX identifiers in missed files

jdemel:
Some files didn't carry a SPDX identifier. Also, some files needed an update. Again, this should make it easier to prepare the VOLK 3.0 release.

This should complete all the necessary files to make a VOLK 3.0 release.


michaelld (approved):
LGTM ; good catches!

PR #592, opened 2022-08-10
sse2neon on volk_8u_x4_conv_k7_r2_8u

Aang23:
This PR adds a NEON-compatible version of the volk_8u_x4_conv_k7_r2_8u kernel, leading to a very noticeable boost in performance on ARM.

Due to the way Volk is done, I had to copy-paste the entire SSE kernel despite only the headers being swapped out. For good measures, the puppet was also copy-pasted, even though it may not be really necessary. (I can change it back depending on your preference)

First mentioned in https://github.com/gnuradio/volk/issues/243.

I am submitting this as a PR not especially to merge straight away, as sse2neon and everything this method may/will imply may need to be discussed further first.


Aang23:
Not a real machine, but here's what CI had to say about performances.
Screenshot from 2022-08-10 17-57-33

The fail on gcc-8 is unrelated to this PR (at least it seems like it), but I do not know how formatting sse2neon should be dealt with.

jdemel:
I'd suggest to leave the puppet in. Due to the way our CI and volk_profile work, this is sometimes necessary. Otherwise, we'd miss this kernel in tests.

Could you add a comment section that explains how things are supposed to work?

I suggest to re-format the kernel.

Aang23:
Ok, I will leave it in.

The kernel is not the main formatting issue, sse2neon.h is. Should it just be ignored or reformatted as well?

Comments-wise, I assume you mean about sse2neon and so on?

I will also test on real hardware in a day or two. Don't have anything on hand for that right now.

jdemel:
We might want to update `sse2neon.h`. In that case, it'd be a shame if we re-format it and we can't diff anymore.

We can go 2 routes. And I feel like a bit of redundancy might be good here.
First, we can ignore `sse2neon.h` in our formatting checks:
https://github.com/gnuradio/volk/blob/0371bf8c50d44db49a86e5c9dc993fd345a51645/.github/workflows/check-pr-formatting.yml#L4-L9
Second, use [clang-format off](https://releases.llvm.org/3.6.0/tools/clang/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code) in the whole `sse2neon.h` file. Even better with a comment that provides further info why we do this.

Aang23:
That seems to work. I've added both methods.

include/volk/sse2neon.h
```diff
@@ -0,0 +1,8853 @@
+// Turn off Clang formatting, as 
+// this would make diffs a lot more
```
jdemel, line 13: That's great. One more request. Can you add a link to the source of this file? I'd like to make sure that this info doesn't get lost over time.

kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
```diff
@@ -175,6 +175,88 @@ static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms,
 #endif /*LV_HAVE_SSE3*/
 
 
+#if LV_HAVE_NEON
+
+#include "volk/sse2neon.h"
+
+static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms,
```
jdemel: Everytime I looked at the test output, I was confused. The name `spiral` is so tightly coupled with x86. I suggest to rename this implementation to `spiralneon` or `neonspiral`. Same with the puppet. It makes the intent and the way it works more obvious.

jdemel (commented):
I suggest to revert the file name change `run-tests.yml` should be without a `.txt` suffix.

Thanks for your work. I hope I just left 3 minor comments and then, this PR should be ready to be merged.

kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
Aang23: Done that as well. Though Spiral actually comes from https://spiral.ece.cmu.edu/vitgen/. Only supports x86, but it's not something to do especially with x86 otherwise, outside of the Spiral software focusing mainly on that arch.

Aang23:
Done! About run-tests.yml.txt, that wa just an oversight. I ran actions on my local machines and wanted to avoid running all tests to simply check formatting. Unfortunately, I forgot to revert before commiting.

Thanks for accepting this change as well! It made a pretty huge difference in usability in some parts of my software on ARM. This should also give GNU Radio (well, the CC Decoder, as it's the only one calling Volk's SIMD) a nice boost as well, and hopefully for a few other projects using this kernel.
As a note, I performed the same tests I used back when it was about fixing the generic implementation of this kernel. BER results of this PR match the SSE implementation very closely.

Also, for reference, qt-dab also supports using a Spiral-generated SSE kernel (though, a different code) with sse2neon, so it's nothing that new either.

jdemel (approved):
LGTM. Thanks!

include/volk/sse2neon.h
```diff
@@ -0,0 +1,8854 @@
+// Turn off Clang formatting, as
```
michaelld, line 12: What is the original copyright on this file? Hopefully it's GPL/ LGPL 3+ ... assuming it's compatible, please add the SPDX line at the top

michaelld (changes requested):
LGTM beyond my query about the license for the new header & getting SPDX in place

include/volk/sse2neon.h
Aang23: It is MIT (https://github.com/DLTcollab/sse2neon/blob/master/LICENSE), so as far as I'm aware, it is compatible.

michaelld: Yes MIT is forward compatible to GPL ... so just add in the SPDX lines & we're good to go

Aang23: Done

include/volk/sse2neon.h
michaelld: Link has been added. What do you think of the SPDX @jdemel ? should it be MIT or can we use GPL here? how do we handle copied code such as this that comes from a different license?

jdemel: Since this is just a copy from another project and we don't want to modify it, I suggest to stick with [SPDX-License-Identifier: MIT](https://spdx.org/licenses/MIT).

michaelld: works for me, since MIT license < GPL license

michaelld (approved):
LGTM

Aang23:
One last update, as I tested this on real HW (and not Android).
```
RUN_VOLK_TESTS: volk_8u_conv_k7_r2puppet_8u(131071,1987)
neonspiral completed in 16217.7 ms
generic completed in 118974 ms
```
So that's ~7.5x faster on arm64-v8a. (This was tested on an RK3318 @ 1.0Ghz)

PR #593, opened 2022-08-17
Fix armeabi-v7a on Android (volk_32fc_convert_16ic_neon)

Aang23:
See https://github.com/gnuradio/volk/issues/589.

I have tested this on both an armeabi-v7a Linux box and Android, tests passed and build suceeds.

### DIFF
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         arch: 
-          - { name: armeabi-v7a, allow_fail: true }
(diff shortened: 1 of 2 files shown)


jdemel (approved):
LGTM. All CI passes. Thus, we can hope it'll work for everyone.
I'd really like to see some armv7 CI but I assume that disappeared with TravisCI.

jdemel:
I added some armv7 tests in #594 . Maybe these changes help to verify that your PR works correctly because these tests include a test run.

michaelld (approved):
LGTM per our discussion; thx!

PR #599, opened 2022-09-05
omit build path

### LINKED ISSUE #597: `lib/constants.c.in` may contain build path
For reproducible builds, the following functions are problematic:

https://github.com/gnuradio/volk/blob/07c1952cd642897af70b6f8bbfeee3a14f54f525/lib/constants.c.in#L31-L41

Under some conditions they may contain build paths that are compiled in. @maitbot pointed out that Debian considers this to be a bug now.

maitbot:
Closes #597 

Have CMake filter out build path from COMPILER_INFO
before using the string in constants.c

Signed-off-by: A. Maitland Bottoms <bottoms@debian.org>

### DIFF
--- a/lib/CMakeLists.txt
+++ b/lib/CMakeLists.txt
@@ -454,6 +454,7 @@ message(STATUS "Loading version ${VERSION} into constants...")
 
 #double escape for windows backslash path separators
 string(REPLACE "\\" "\\\\" prefix "${prefix}")
+string(REPLACE "${CMAKE_SOURCE_DIR}" "$BUILD_DIR" COMPILER_INFO "${COMPILER_INFO}")
 
 configure_file(
     ${CMAKE_CURRENT_SOURCE_DIR}/constants.c.in


jdemel:
This is the [CMake string replace](https://cmake.org/cmake/help/latest/command/string.html#replace) command. The `COMPILER_INFO` variable may contain a build path. With this change, it shouldn't anymore. Instead every occurence is replace with `$BUILD_DIR`. Seems like a good solution to me.
@maitbot Thanks!

jdemel (approved):
LGTM

lib/CMakeLists.txt
```diff
@@ -454,6 +454,7 @@ message(STATUS "Loading version ${VERSION} into constants...")
 
 #double escape for windows backslash path separators
 string(REPLACE "\\" "\\\\" prefix "${prefix}")
+string(REPLACE "${CMAKE_SOURCE_DIR}" "$BUILD_DIR" COMPILER_INFO "${COMPILER_INFO}")
```
michaelld, line 457: is the 2nd argument intentionally not `${}` encapsulated? I don't think this matters for CMake, but it's worth asking as I'm not sure what the end result of this command will be for various SOURCE and BUILD options.

michaelld (commented):
See my comment on the issue as well as here ... I guess I'm confused as to what this change actually accomplishes

lib/CMakeLists.txt
maitbot: yes.
That argument is the literal string constant that is compiled into volk-config-info.
The end result is that compiled output becomes the same no matter what the actual build path is,
supporting reproducible builds.
The irony is that the build path replaced here is the one used in the -fdebug-prefix-map compiler option that the build demons use to tell the compiler to omit the build path from the ouput.
I have been applying this patch for the Debian packages for a while, merging this into volk will avoid a reproducible build bug for other builders of volk too.

More info under
https://tests.reproducible-builds.org/debian/issues/unstable/records_build_flags_issue.html

jdemel:
I assume all questions are answered. Merging now.

PR #602, opened 2022-09-24
Fix build for 32 bit arm with neon.

### LINKED ISSUE #601: build problem 2.5.2 32 bit
I can't bulid versiion 2.5.2 volk on a 32 bit armv7l cpu.

[build.log](https://github.com/gnuradio/volk/files/9543607/build.log)
[compile.log](https://github.com/gnuradio/volk/files/9543605/compile.log)

balister:

### DIFF
--- a/include/volk/sse2neon.h
+++ b/include/volk/sse2neon.h
@@ -444,7 +444,7 @@ FORCE_INLINE uint32_t _mm_crc32_u8(uint32_t, uint8_t);
 
 // Older gcc does not define vld1q_u8_x4 type
 #if defined(__GNUC__) && !defined(__clang__) &&                        \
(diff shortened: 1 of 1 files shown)


michaelld (approved):
This looks fine ... wondering why we limit GNUC to 12 though ... why not just generic `if(defined(__GNUC__) && ... && (defined(__arm__)) || ...` or something like that so-as to not have to keep updating this number (here from `11` to `12`)

balister:
The file comes from this repository: https://github.com/DLTcollab/sse2neon And yes, until gcc supports the type for arm, they will need to keep changing that number.

jdemel:
Thanks for the fix!

PR #603, opened 2022-09-27
Use CpuFeatures target

jsallay:
The build process is more flexible if we use cmake targets rather than directly linking to libraries. The target is used in other places in the CMakeLists.txt, but was not used in the target_link_libraries command.


marcusmueller:
"Build static on ubuntu-latest" fails because:
```
  Target "volk_static" links to:
    CpuFeatures::cpu_features
  but the target was not found. 
```

don't know why that happens, but not on the older ubuntus.

jsallay:
I don't understand either. When I get a break, I was going to try to
recreate the error in a docker container.

marcusmueller:
hm maybe the others use a directly downloaded cpufeatures and the most current one relies on a ubuntu package?

jdemel (dismissed):
LGTM. I was under the impression we would do this everywhere. Also, I assumed `cpu_features` is an alias for `CpuFeatures::cpu_features`. Do we use the same approach for the dynamic library?

lib/CMakeLists.txt
```diff
@@ -577,7 +577,7 @@ if(ENABLE_STATIC_LIBS)
   add_library(volk_static STATIC $<TARGET_OBJECTS:volk_obj>)
   target_link_libraries(volk_static PUBLIC ${volk_libraries})
   if(VOLK_CPU_FEATURES)
-    target_link_libraries(volk_static PRIVATE cpu_features)
+    target_link_libraries(volk_static PRIVATE CpuFeatures::cpu_features)
```
michaelld: To be consistent, please do the same for line 537 `target_link_libraries(volk PRIVATE cpu_features)`

michaelld (changes requested):
LGTM after this change

ashley-b:
Any chance of this getting merged in ?

jdemel:
@jsallay I guess @ashley-b 's question is mostly direct at you.

I'd like to merge this as soon as all issues are fixed. That would be 
1. the DCO issue which should be trivial to fix
2. the `ubuntu-latest` issue. Unfortunately, the error message is gone by now. It should be simple to trigger our CI though with a simple push with the DCO fix.

ashley-b:
@jsallay Thanks for the quick follow up, I just tried building it in a ubuntu container and it built with out error

lsb_release -a
```
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.5 LTS
Release:	20.04
Codename:	focal
```

Also I noticed not all the target_link_librarys for cpu_feature have been changed to use modern CMake targets.

https://github.com/jsallay/volk/blob/823e88aec459011ebd519f50ec2caa2ccc9bcb3b/lib/CMakeLists.txt#L537

and

This block does not make sense, since both lines of the if/else seem to do the same thing
https://github.com/jsallay/volk/blob/823e88aec459011ebd519f50ec2caa2ccc9bcb3b/lib/CMakeLists.txt#L510C1-L517C6

lib/CMakeLists.txt
jsallay: Done

jsallay:
Looks like `CpuFeature::cpu_features` is the consistently defined target.

jdemel:
@jsallay thanks for taking care of this PR. It looks good to me now. All the comments are addressed. Merging.

PR #606, opened 2022-10-26
Use a different instruction for armv8 neon loads.

jsallay:
The instruction used for load/store assumes that the data is interleaved and produces incorrect results for the 64-bit byteswap.

Use a different instruction that doesn't assume interleaving.

Signed-off-by: John Sallay <jasallay@gmail.com>

### DIFF
--- a/kernels/volk/volk_64u_byteswap.h
+++ b/kernels/volk/volk_64u_byteswap.h
@@ -242,10 +242,10 @@ static inline void volk_64u_byteswap_neonv8(uint64_t* intsToSwap, unsigned int n
     unsigned int number = 0;
     for (number = 0; number < n4points; ++number) {
         __VOLK_PREFETCH(inputPtr + 8);
-        input = vld2q_u8((uint8_t*)inputPtr);
+        input = vld1q_u8_x2((uint8_t*)inputPtr);
         input.val[0] = vqtbl1q_u8(input.val[0], idx);
         input.val[1] = vqtbl1q_u8(input.val[1], idx);
-        vst2q_u8((uint8_t*)inputPtr, input);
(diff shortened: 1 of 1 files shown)


jsallay:
I'm happy to add a test, but I'm not exactly sure where to do it.

kernels/volk/volk_64u_byteswap.h
```diff
@@ -242,10 +242,10 @@ static inline void volk_64u_byteswap_neonv8(uint64_t* intsToSwap, unsigned int n
     unsigned int number = 0;
     for (number = 0; number < n4points; ++number) {
         __VOLK_PREFETCH(inputPtr + 8);
-        input = vld2q_u8((uint8_t*)inputPtr);
+        input = vld1q_u8_x2((uint8_t*)inputPtr);
```
jdemel, line 245: Our GCC 8 test fails with this:
`/usr/bin/ld: libvolk.so.2.5.2: undefined reference to `vst1q_u8_x2'`

Since newer GCC versions seem to work, We might require smth like an `#ifdef` to switch between those functions.

jsallay: I can do that, but I'll have to figure out what is defined when the `vld1q_u8_x2` instruction exists.

jdemel: That, or require GCC 9 and up. But we need to find some documentation that clearly explains what's going on.

argilo:
#680 proposes to simply remove the buggy kernel instead, since it is slower than the generic kernel.

jdemel:
Thanks for the PR. Since we just removed these kernels, I'm closing this PR now.

PR #609, opened 2022-12-21
Support for MIPS and RISC-V is experimental

michael-roe:


michael-roe:
This PR adds a note on README.md that support for MIPS and RISC-V is experimental. (Any actual fixes for these architectures will be in separate PRs.)

jdemel:
I assume this PR should be merged after #573 . Thanks for this addition!

The failing tests indicate that we should switch from `ubuntu-latest` to `ubuntu-20.04`. Or, even better, use Ubuntu 20.04 only for older compilers.

michael-roe:
Yes, the discussion of #573 was that we should update README.md in a separate commit.

PR #610, opened 2022-12-25
ci: Fix CI to use later distros

jdemel:
This PR fixes CI that started to fail due to some changes in GitHub Actions. Mostly the `ubuntu-latest` label points to `ubuntu-22.04` now.

We should remove CI that relies on GCC < 9.0 and Clang < 10.0 after April '23. These are the default compilers in Ubuntu 20.04. Anything older than this should be considered obsolete.

In the process, I added a test for `armv7`, `riscv64`, `ppc64le`, and `s390x`. If the employed GH Action supports MIPS in the future, we should add that too. However, the test are configured to allow failure in case of non-ARM and non-x86 architectures.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -4,26 +4,30 @@ on: [push, pull_request]
 
 jobs:
   build-ubuntu:
-    name: Build on ubuntu-latest ${{ matrix.compiler.name }}
+    name: Build on ${{ matrix.compiler.distro }} ${{ matrix.compiler.name }}
 
     strategy:
       fail-fast: false
       matrix:
         compiler: 
-          - { name: g++-7, cc: gcc-7, cxx: g++-7 }
-          - { name: g++-8, cc: gcc-8, cxx: g++-8 }
-          - { name: g++-9, cc: gcc-9, cxx: g++-9 }
-          - { name: g++-10, cc: gcc-10, cxx: g++-10 }
-          - { name: clang-7, cc: clang-7, cxx: clang++-7 }
-          - { name: clang-8, cc: clang-8, cxx: clang++-8 }
-          - { name: clang-9, cc: clang-9, cxx: clang++-9 }
-          - { name: clang-10, cc: clang-10, cxx: clang++-10 }
-          # - { name: clang-11, cc: clang-11, cxx: clang++-11 }
+          - { name: g++-7, cc: gcc-7, cxx: g++-7, distro: ubuntu-20.04 }
+          - { name: g++-8, cc: gcc-8, cxx: g++-8, distro: ubuntu-20.04 }
+          - { name: g++-9, cc: gcc-9, cxx: g++-9, distro: ubuntu-latest }
+          - { name: g++-10, cc: gcc-10, cxx: g++-10, distro: ubuntu-latest }
+          - { name: clang-7, cc: clang-7, cxx: clang++-7, distro: ubuntu-20.04 }
+          - { name: clang-8, cc: clang-8, cxx: clang++-8, distro: ubuntu-20.04 }
+          - { name: clang-9, cc: clang-9, cxx: clang++-9, distro: ubuntu-20.04 }
+          - { name: clang-10, cc: clang-10, cxx: clang++-10, distro: ubuntu-20.04 }
+          - { name: clang-11, cc: clang-11, cxx: clang++-11, distro: ubuntu-20.04 }
+          - { name: clang-12, cc: clang-12, cxx: clang++-12, distro: ubuntu-latest }
+          - { name: clang-13, cc: clang-13, cxx: clang++-13, distro: ubuntu-latest }
+          - { name: clang-14, cc: clang-14, cxx: clang++-14, distro: ubuntu-latest }
+          # - { name: clang-15, cc: clang-15, cxx: clang++-15, distro: ubuntu-latest }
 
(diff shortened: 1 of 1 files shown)


Aang23:
Are you sure ditching tests on older compiler is a good idea? While most distros using older GCC versions can be considered EOL, there is still a pretty massive userbase on those.

I currently already need to support and patch things around to support Volk 1, but if Volk drops support this will make supporting older systems quite a bit more complicated than it already is.

I'd perhaps suggest keeping this around and dropping support whenever meaningful changes require doing so, which I wouldn't expect to happen so soon.

jdemel:
Just some thoughts on your comment. Maybe we can find a good solution or better understanding of the situation.

For now: No tests were dropped. Only new tests were added and changes to the CI system were made to make sure tests run again.

> Are you sure ditching tests on older compiler is a good idea? While most distros using older GCC versions can be considered EOL, there is still a pretty massive userbase on those.

This is a difficult topic. Generally, I feel like we should make sure VOLK runs on supported open source systems. Thus, we try to support e.g. Ubuntu 18.04 for now.
Are there any statistics on that topic? e.g. which compiler version on which distro?

I feel differently towards older systems that receive paid support though. The same feeling holds for proprietary systems. The VOLK project or GNU Radio does not receive any of that money while users are willing to send money to other parties.

> I currently already need to support and patch things around to support Volk 1, but if Volk drops support this will make supporting older systems quite a bit more complicated than it already is.

This is a very interesting topic. Are you able to share more info on that? Since we don't gather any user statistics, it is hard to know which versions are still in use. Also, how are they used? If you prefer an email or similar, feel free to get in touch.

> I'd perhaps suggest keeping this around and dropping support whenever meaningful changes require doing so, which I wouldn't expect to happen so soon.

I assume we can do that. Though, at some point I feel like we use a lot of resources for tests and need to be modest with these resources.

One more thought on that topic: a larger test matrix translates to a larger cognitive burden. This is something we need to be aware of.

Aang23:
> This is a difficult topic. Generally, I feel like we should make sure VOLK runs on supported open source systems. Thus, we try to support e.g. Ubuntu 18.04 for now.
Are there any statistics on that topic? e.g. which compiler version on which distro?
I feel differently towards older systems that receive paid support though. The same feeling holds for proprietary systems. The VOLK project or GNU Radio does not receive any of that money while users are willing to send money to other parties.

I don't know of any usage statistics, but for Ubuntu / Debian compiler options short of using backports / building it yourself are pretty limited. 16.04  had gcc5 and 18.04 gcc7. Personally I've cut the line at gcc7 due to C++17, but that doesn't mean users don't go their way to use gcc7 on 16.04 leading to the usual challenges anyway.

As for paid support, I agree. First of all no "normal" user is likely to be using those, and if they are it's such an edge case it's not an issue. Plus as you said, the FOSS projects don't get any of that money. 

> This is a very interesting topic. Are you able to share more info on that? Since we don't gather any user statistics, it is hard to know which versions are still in use. Also, how are they used? If you prefer an email or similar, feel free to get in touch.

I don't collect any statistics, but I can give you what I do know :
1) Out of all people I know using SatDump on Linux distributions, easily 1/3rd are still using distributions shipping libvolk1 and gcc7 on at least one machine. Not a massive dataset, but probably a "dataset" of around 30-40 people.
2) Every time I break something with libvolk1 support, I easily get a few issues / messages about it in the hour.  
3) Volk1 is still widespread enough for other projects such as SDR++ to also be working around it in the same way.

Not proper statics, but it all sure does make me feel like libvolk1 is still very widespread among the whole userbase. Perhaps it'd be a better idea to run a poll about it though.

The issue with Volk compared to another library actually is GNU Radio. If it was volk alone, I could just tell everyone who's using an older distribution to build Volk themselves, but in most cases GNU Radio is also installed on the same system requiring this specific volk version. At that point, it's required to both re-build Volk and GNU Radio, which is a pretty horrible user experience :-)

I currently work around, for older Volks : 
1) Lacking volk_alloc in volk1. I ship them in my repo and include then if the system doesn't.
2) Broken convolutional kernels in most current versions. While it's fixed here, many distributions still ship much older builds...
3) Missing kernels. Only one of them, but it means I have to provide the implementation myself when this happens.

Though clearly, this isn't an issue with Volk, just how things are.

> I assume we can do that. Though, at some point I feel like we use a lot of resources for tests and need to be modest with these resources.

Coming back to why I left that comment initially : currently when someone reaches out and their volk is simply too old, I can redirect them to building it from this repository, or backports shipping the latest (eg, debian). Dropping support there would instantly kill this option and lead to even more version fragmentation in my opinion.

> One more thought on that topic: a larger test matrix translates to a larger cognitive burden. This is something we need to be aware of.
Very good point too... There isn't a huge team working on Volk after all, so at some point compromises have to be made. After all, supporting outdated systems should not impact new features / improvements which are clearly more important in the long run. 

I would perhaps suggest dropping GCC versions under 7 instead, as 16.04 can clearly be considered obsolete and the burden to be on the users to setup a newer compiler etc. But still common Ubuntu, Debian, CentOS, etc releases still ship with gcc7 so I'd keep supporting it until it becomes too much of a burden.

Aang23:
Oh and - as a note, if you wish to discuss this further, I'm on Volk's Matrix server as @aang23:altillimity.com

jdemel:
@Aang23 thanks for the discussion! I just merged this PR. I will keep our talking points here in mind.

PR #615, opened 2023-01-15
cmake: Apply cmake-format

### LINKED ISSUE #423: auto format cmake files
I suggest to use [cmake_format](https://github.com/cheshirekow/cmake_format) to auto-format all our CMake files. cmake is often hard to read and having it at least formatted would ease this task. Of course, we'd need to decide on any particular style guide. etc.

### LINKED ISSUE #600: Use CMake format like GR
GNU Radio did a cmake-format. This is smth VOLK would benefit from as well.
https://github.com/gnuradio/gnuradio/pull/6100

We'd only need to pull one the [`.cmake-format.py` file](https://github.com/gnuradio/gnuradio/blob/main/.cmake-format.py) and use it in VOLK.

Luckily, this file is LGPL'd
https://github.com/gnuradio/gnuradio/blob/67ef0469a0dabb3e1d876b4ac7d79b505a26873d/.cmake-format.py#L2

jdemel:
This commit adds `.cmake-format.py`. It is mostly inspired by the GNU Radio version of it.

Steps to reproduce these results:
- `pip3 install cmakelang`
- `cmake-format --dump-config python > .cmake-format.py`
- Modify .cmake-format.py to be close to the GNU Radio version.
- `cmake-format -c .cmake-format.py -i $(git ls-files | grep CMakeLists.txt) $(git ls-files | grep .cmake)`

Fixes #600
Fixes #423


cmake/Toolchains/aarch64-linux-gnu.cmake
```diff
@@ -9,7 +9,10 @@
 set(CMAKE_SYSTEM_NAME Linux)
 set(CMAKE_SYSTEM_PROCESSOR aarch64)
 
-if(MINGW OR CYGWIN OR WIN32)
+if(
+  MINGW
+  OR CYGWIN
+  OR WIN32)
```
marcusmueller: this is extremely ugly.

cmake/Toolchains/arm_cortex_a72_hardfp_native.cmake
```diff
@@ -11,7 +11,16 @@
...
-set(CMAKE_ASM_FLAGS "${CMAKE_CXX_FLAGS} -mthumb -g" CACHE STRING "" FORCE) #same flags for asm sources
+set(CMAKE_C_COMPILER gcc)
+set(
+  CMAKE_CXX_FLAGS
+  "-march=armv8-a -mtune=cortex-a72 -mfpu=neon-fp-armv8 -mfloat-abi=hard"
+  CACHE STRING "" FORCE)
+set(
+  CMAKE_C_FLAGS
+  ${CMAKE_CXX_FLAGS}
+  CACHE STRING "" FORCE) #same flags for C sources
```
marcusmueller: Not sure, but this looks pretty different than GR's formatting. You're closing an issue that says "format like GNU Radio" with this!

marcusmueller (changes requested):

marcusmueller:
If you want to be close to the GNU Radio version, so go and copy the GNU Radio version, I'd say.

jdemel:
> If you want to be close to the GNU Radio version, so go and copy the GNU Radio version, I'd say.

I'm pretty sure, I did. I saw you @marcusmueller made some further adjustments to the GR CMake formatting. I should probably get an updated version of this.

cmake/Toolchains/aarch64-linux-gnu.cmake
jdemel: I re-added the corresponding file from gnuradio and the result is almost the same but without the new line after `if(`. I'd assume this is intended.

jdemel:
I copied the `.clang-format.py` file from GNU Radio again. I expect that the result looks like what GR does now. I diff'd the files and the VOLK version of `.clang-format.py` is the same + some comments that drop out of the default. I'd argue this is nice for future reference.

marcusmueller (approved):
looks good; thanks!

PR #616, opened 2023-01-18
Use cpu_features on RISC-V platforms

michael-roe:
This patch updates VOLK to use cpu_features on RISC-V platforms.
You will need to update the submodule pointer cpu_features first, because RISC-V support has ony recently been added to cpu_features.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -114,7 +114,7 @@ jobs:
           submodules: 'recursive'
       - uses: uraimo/run-on-arch-action@v2.5.0
         name: Build in non-x86 container
-        continue-on-error: ${{ contains(fromJson('["ppc64le", "s390x", "riscv64"]'), matrix.arch) || contains(fromJson('["clang-14"]'), matrix.compiler.name) }}
+        continue-on-error: ${{ contains(fromJson('["ppc64le", "s390x"]'), matrix.arch) || contains(fromJson('["clang-14"]'), matrix.compiler.name) }}
         id: build
         with:
           arch: ${{ matrix.arch }}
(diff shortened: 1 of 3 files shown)


jdemel:
That's great. I was already looking into RISC-V support. Thanks for the PR!

Though, there's smth wrong with our test:
https://github.com/gnuradio/volk/actions/runs/3951765765/jobs/6776828937
It passes because we don't require it to go through yet. Would be great to change that.

jdemel:
I had another look at the error message.

```
CMake Error at cpu_features/CMakeLists.txt:82 (message):
    Unsupported architectures riscv64
    cpu_features/CMakeLists.txt:126 (add_cpu_features_headers_and_sources)
```
It looks like the submodule pointer needs an update. 

Also, could you update our CI in this line:
https://github.com/gnuradio/volk/blob/8a226321b1e1892c67c79c03e4b836d6f8553711/.github/workflows/run-tests.yml#L117

Just remove the `riscv64` string from the list. That'd make it a requirement to pass this test successfully. We can still deactivate it if other tests fail.

michaelld (commented):
This LGTM ... relatively simple changes that should work .. how do we update the submodule pointer?

michael-roe:
Ok, I've updates the PR to include an update of the cpu_features submodule and the change to the CI.

michaelld:
Great! I think this PR is complete now. Let's see how the CI/QA/checks do ...

michaelld (approved):
LGTM ... the CI / QA build error has nothing to do with this PR's changes ... it's a kernel that fails sometimes (`volk_32fc_32f_dot_prod_32fc`)

jdemel (approved):
LGTM! I'm really happy to add RISC-V support. 
Also, the CI indicates that this is a platform where our C vs C++ interface issue doesn't pop up.

jdemel:
Thanks @michael-roe !

PR #617, opened 2023-01-19
add volk_32f_s32f_x2_convert_8u kernel

### LINKED ISSUE #6433: FEC Async Decoder overflows when converting from float to uint8_t
### What happened?

I have noticed that the FEC Async Decoder block can overflow when doing a conversion from `float` to `uint8_t` if the amplitude of the input is too large. This conversion happens, for instance, if the block is used with a CC Decoder Definition to implement a Viterbi decoder.

These are the problematic conversions:
https://github.com/gnuradio/gnuradio/blob/6efdc9d84e6149e3f4049ee7c558dc01106dfd06/gr-fec/lib/async_decoder_impl.cc#L132

https://github.com/gnuradio/gnuradio/blob/6efdc9d84e6149e3f4049ee7c558dc01106dfd06/gr-fec/lib/async_decoder_impl.cc#L185

This is a potential problem when using this block to perform soft-input Viterbi decoding. If the amplitude of the input is too large for any reason, all the conversions to `uint8_t` will overflow, and the effect is that we will pass the inverted symbol stream to the Viterbi decoder. With a transparent convolutional code (most common codes), the Viterbi decoder will happily output the inverted output. This is how I first noticed the problem.

There are some commented lines in the code with calls to `volk_32f_s32f_convert_8i`. These date back to fbbde8d01ab43b7a3ffd5ca9d65cf50a84e2a7be when Tom Rondeau first implemented the block in 2014. The volk kernels for conversion to integers saturate instead of overflowing, but I don't think that the kernels can be used here because there are no kernels that convert to `uint8` (unless we want to jump through some hoops by using the `int8` kernel and adding the bias to convert to `uint8` later, which is not straightforward, because the bias is specified by `d_decoder->get_shift()` as a float, even though that shift would be `128.0` in most cases).

What I think we should do is to use the same code as in the Float to UChar block, since this is exactly the same problem. This block uses [float_array_to_uchar.cc](https://github.com/gnuradio/gnuradio/blob/main/gr-blocks/lib/float_array_to_uchar.cc), which unfortunately I think it is not accessible from gr-fec.

Perhaps we should add a `volk_32f_s32f_s32f_convert_8u` kernel to Volk that includes scale and bias parameters, even if at the moment this kernel would only contain a generic implementation (similar to float_array_to_uchar.cc, but including the multiplication by the scale and addition of the bias).

Thoughts?

### System Information

All systems are affected.

### GNU Radio Version

4.0-git (dev-4.0)

### Specific Version

0.0.0.0

### Steps to Reproduce the Problem

This can be tested by feeding an input with too large amplitude (amplitude 4 is already large enough) to a Viterbi decoder implemented using FEC Async Decoder and CC Decoder Defintion, and then looking at the output of the decoder.

### Relevant log output

_No response_

daniestevez:
This adds a kernel that performs conversion from float to uint8_t using a scale and a bias, according to `out = in * scale + bias`. The output is clamped to the interval [0, 255] and rounded to the nearest integer.

The kernels are implemented mirroring those for `volk_32f_s32f_convert_8i`, but there is an additional `avx2_fma` kernel, since it makes sense to perform the scale and bias conversion using FMA if it is available.

The main intended use case for this kernel is the conversion from `float` soft symbols to `uint8` for the Viterbi decoder in GNU Radio, as a way forward to fix https://github.com/gnuradio/gnuradio/issues/6433. The kernel is also applicable in other contexts, such as for example in the Float to UChar block, which currently doesn't use Volk (although with this kernel we would use a `bias` of zero, thus doing needless operations).

I think that having the bias as part of the kernel is good, since generally when converting from `float` to unsigned integers we will need to add a bias. Having it as part of the kernel enables us to use FMA. Perhaps a bias parameter could be added to Float to UChar in the future.

### DIFF
--- a/kernels/volk/volk_32f_s32f_x2_convert_8u.h
+++ b/kernels/volk/volk_32f_s32f_x2_convert_8u.h
@@ -1,0 +1,616 @@
+/* -*- c++ -*- */
+/*
+ * Copyright 2023 Daniel Estevez <daniel@destevez.net>
+ * Copyright 2012, 2014 Free Software Foundation, Inc.
+ *
+ * This file is part of VOLK
+ *
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+/*!
+ * \page volk_32f_s32f_x2_convert_8u
+ *
+ * \b Overview
+ *
+ * Converts a floating point number to an 8-bit unsigned int after applying a
+ * multiplicative scaling factor and an additive bias.
+ *
+ * <b>Dispatcher Prototype</b>
+ * \code
+ * void volk_32f_s32f_x2_convert_8u(uint8_t* outputVector, const float* inputVector,
+ const float scale, const float bias, unsigned int num_points)
+ * \endcode
+ *
+ * \b Inputs
+ * \li inputVector: the input vector of floats.
+ * \li scale: The value multiplied against each point in the input buffer.
+ * \li bias: The value added to each multiplication by the scale.
+ * \li num_points: The number of data points.
+ *
+ * \b Outputs
+ * \li outputVector: The output vector.
+ *
+ * \b Example
+ * Convert floats from [-1,1] to 8-bit unsigend integers with a scale of 128 and a bias of
+ 128
+ *  int N = 10;
+ *   unsigned int alignment = volk_get_alignment();
+ *   float* increasing = (float*)volk_malloc(sizeof(float)*N, alignment);
+ *   uint8_t* out = (uint8_t*)volk_malloc(sizeof(uint8_t)*N, alignment);
+ *
+ *   for(unsigned int ii = 0; ii < N; ++ii){
+ *       increasing[ii] = 2.f * ((float)ii / (float)N) - 1.f;
+ *   }
+ *
+ *   float scale = 128.0f;
+ *   float bias = 128.0f;
+ *
+ *   volk_32f_s32f_x2_convert_8u(out, increasing, scale, bias, N);
+ *
+ *   for(unsigned int ii = 0; ii < N; ++ii){
+ *       printf("out[%u] = %i\n", ii, out[ii]);
+ *   }
+ *
+ *   volk_free(increasing);
+ *   volk_free(out);
+ * \endcode
+ */
+
+#ifndef INCLUDED_volk_32f_s32f_x2_convert_8u_u_H
+#define INCLUDED_volk_32f_s32f_x2_convert_8u_u_H
+
+#include <inttypes.h>
+
+static inline void volk_32f_s32f_x2_convert_8u_single(uint8_t* out, const float in)
+{
+    const float min_val = 0.0f;
+    const float max_val = UINT8_MAX;
+    if (in > max_val) {
+        *out = (uint8_t)(max_val);
+    } else if (in < min_val) {
+        *out = (uint8_t)(min_val);
+    } else {
+        *out = (uint8_t)(rintf(in));
+    }
+}
+
+
+#ifdef LV_HAVE_GENERIC
+
+static inline void volk_32f_s32f_x2_convert_8u_generic(uint8_t* outputVector,
+                                                       const float* inputVector,
+                                                       const float scale,
+                                                       const float bias,
+                                                       unsigned int num_points)
+{
+    const float* inputVectorPtr = inputVector;
+
+    for (unsigned int number = 0; number < num_points; number++) {
+        const float r = *inputVectorPtr++ * scale + bias;
+        volk_32f_s32f_x2_convert_8u_single(&outputVector[number], r);
+    }
+}
+
+#endif /* LV_HAVE_GENERIC */
+
+
+#if LV_HAVE_AVX2 && LV_HAVE_FMA
+#include <immintrin.h>
+
+static inline void volk_32f_s32f_x2_convert_8u_u_avx2_fma(uint8_t* outputVector,
+                                                          const float* inputVector,
+                                                          const float scale,
+                                                          const float bias,
+                                                          unsigned int num_points)
+{
+    const unsigned int thirtysecondPoints = num_points / 32;
+
+    const float* inputVectorPtr = (const float*)inputVector;
+    uint8_t* outputVectorPtr = outputVector;
+
+    const float min_val = 0.0f;
+    const float max_val = UINT8_MAX;
+    const __m256 vmin_val = _mm256_set1_ps(min_val);
+    const __m256 vmax_val = _mm256_set1_ps(max_val);
+
+    const __m256 vScale = _mm256_set1_ps(scale);
+    const __m256 vBias = _mm256_set1_ps(bias);
+
+    for (unsigned int number = 0; number < thirtysecondPoints; number++) {
+        __m256 inputVal1 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+        __m256 inputVal2 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+        __m256 inputVal3 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+        __m256 inputVal4 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+
+        inputVal1 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_fmadd_ps(inputVal1, vScale, vBias), vmax_val), vmin_val);
+        inputVal2 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_fmadd_ps(inputVal2, vScale, vBias), vmax_val), vmin_val);
+        inputVal3 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_fmadd_ps(inputVal3, vScale, vBias), vmax_val), vmin_val);
+        inputVal4 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_fmadd_ps(inputVal4, vScale, vBias), vmax_val), vmin_val);
+
+        __m256i intInputVal1 = _mm256_cvtps_epi32(inputVal1);
+        __m256i intInputVal2 = _mm256_cvtps_epi32(inputVal2);
+        __m256i intInputVal3 = _mm256_cvtps_epi32(inputVal3);
+        __m256i intInputVal4 = _mm256_cvtps_epi32(inputVal4);
+
+        intInputVal1 = _mm256_packs_epi32(intInputVal1, intInputVal2);
+        intInputVal1 = _mm256_permute4x64_epi64(intInputVal1, 0b11011000);
+        intInputVal3 = _mm256_packs_epi32(intInputVal3, intInputVal4);
+        intInputVal3 = _mm256_permute4x64_epi64(intInputVal3, 0b11011000);
+
+        intInputVal1 = _mm256_packus_epi16(intInputVal1, intInputVal3);
+        const __m256i intInputVal = _mm256_permute4x64_epi64(intInputVal1, 0b11011000);
+
+        _mm256_storeu_si256((__m256i*)outputVectorPtr, intInputVal);
+        outputVectorPtr += 32;
+    }
+
+    for (unsigned int number = thirtysecondPoints * 32; number < num_points; number++) {
+        const float r = inputVector[number] * scale + bias;
+        volk_32f_s32f_x2_convert_8u_single(&outputVector[number], r);
+    }
+}
+
+#endif /* LV_HAVE_AVX2 && LV_HAVE_FMA */
+
+
+#ifdef LV_HAVE_AVX2
+#include <immintrin.h>
+
+static inline void volk_32f_s32f_x2_convert_8u_u_avx2(uint8_t* outputVector,
+                                                      const float* inputVector,
+                                                      const float scale,
+                                                      const float bias,
+                                                      unsigned int num_points)
+{
+    const unsigned int thirtysecondPoints = num_points / 32;
+
+    const float* inputVectorPtr = (const float*)inputVector;
+    uint8_t* outputVectorPtr = outputVector;
+
+    const float min_val = 0.0f;
+    const float max_val = UINT8_MAX;
+    const __m256 vmin_val = _mm256_set1_ps(min_val);
+    const __m256 vmax_val = _mm256_set1_ps(max_val);
+
+    const __m256 vScale = _mm256_set1_ps(scale);
+    const __m256 vBias = _mm256_set1_ps(bias);
+
+    for (unsigned int number = 0; number < thirtysecondPoints; number++) {
+        __m256 inputVal1 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+        __m256 inputVal2 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+        __m256 inputVal3 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+        __m256 inputVal4 = _mm256_loadu_ps(inputVectorPtr);
+        inputVectorPtr += 8;
+
+        inputVal1 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_add_ps(_mm256_mul_ps(inputVal1, vScale), vBias),
+                          vmax_val),
+            vmin_val);
+        inputVal2 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_add_ps(_mm256_mul_ps(inputVal2, vScale), vBias),
+                          vmax_val),
+            vmin_val);
+        inputVal3 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_add_ps(_mm256_mul_ps(inputVal3, vScale), vBias),
+                          vmax_val),
+            vmin_val);
+        inputVal4 = _mm256_max_ps(
+            _mm256_min_ps(_mm256_add_ps(_mm256_mul_ps(inputVal4, vScale), vBias),
+                          vmax_val),
+            vmin_val);
+
+        __m256i intInputVal1 = _mm256_cvtps_epi32(inputVal1);
+        __m256i intInputVal2 = _mm256_cvtps_epi32(inputVal2);
+        __m256i intInputVal3 = _mm256_cvtps_epi32(inputVal3);
+        __m256i intInputVal4 = _mm256_cvtps_epi32(inputVal4);
+
+        intInputVal1 = _mm256_packs_epi32(intInputVal1, intInputVal2);
+        intInputVal1 = _mm256_permute4x64_epi64(intInputVal1, 0b11011000);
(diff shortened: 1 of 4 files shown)


daniestevez:
By the way, I've inadvertently linked this PR to the issue about the FEC Async Decoder overflow, so that merging this PR will close the issue. I think this is because of the wording I used on the description ("fix issue-reference"). This is unintended. After this gets merged, there is work to do on the GNU Radio side to fix the FEC Async Decoder overflow. Any idea of how to unlink it? I don't seem to have a button to do that.

daniestevez:
The test is failing on ubuntu-20.04 clang-10 and ubuntu-20.04 clang-11. The issue seems to be an error of 1 in the output of the FMA kernels.

clang-10:
```
2023-01-19T17:16:20.7409400Z test 49
2023-01-19T17:16:20.7409847Z         Start  49: qa_volk_32f_s32f_convertpuppet_8u
2023-01-19T17:16:20.7410201Z 
2023-01-19T17:16:20.7410613Z 49: Test command: /usr/bin/sh "/home/runner/work/volk/volk/build/lib/volk_32f_s32f_convertpuppet_8u_test.sh" "/home/runner/work/volk/volk/build/lib"
2023-01-19T17:16:20.7411214Z 49: Working Directory: /home/runner/work/volk/volk/build/lib
2023-01-19T17:16:20.7412465Z 49: Test timeout computed to be: 10000000
2023-01-19T17:16:20.7458680Z 49: RUN_VOLK_TESTS: volk_32f_s32f_convertpuppet_8u(131071,1)
2023-01-19T17:16:20.7621520Z 49: generic completed in 1.23702 ms
2023-01-19T17:16:20.7623264Z 49: u_avx2_fma completed in 0.248504 ms
2023-01-19T17:16:20.7625429Z 49: a_avx2_fma completed in 0.171203 ms
2023-01-19T17:16:20.7627228Z 49: u_avx2 completed in 0.204704 ms
2023-01-19T17:16:20.7628994Z 49: a_avx2 completed in 0.166603 ms
2023-01-19T17:16:20.7630859Z 49: u_sse2 completed in 0.174504 ms
2023-01-19T17:16:20.7632698Z 49: a_sse2 completed in 0.176103 ms
2023-01-19T17:16:20.7635813Z 49: u_sse completed in 0.289506 ms
2023-01-19T17:16:20.7638865Z 49: a_sse completed in 0.289006 ms
2023-01-19T17:16:20.7641298Z 49: offset 111360 in1: 14 in2: 13 tolerance was: 0
2023-01-19T17:16:20.7641612Z 49: volk_32f_s32f_convertpuppet_8u: fail on arch u_avx2_fma
2023-01-19T17:16:20.7646601Z 49: offset 111360 in1: 14 in2: 13 tolerance was: 0
2023-01-19T17:16:20.7647245Z 49: volk_32f_s32f_convertpuppet_8u: fail on arch a_avx2_fma
2023-01-19T17:16:20.7671053Z 49: Best aligned arch: a_avx2
2023-01-19T17:16:20.7671578Z 49: Best unaligned arch: u_sse2
2023-01-19T17:16:20.7679183Z  49/141 Test  #49: qa_volk_32f_s32f_convertpuppet_8u ....................***Failed    0.03 sec
```
clang-11:
```
2023-01-19T17:16:19.3379444Z test 49
2023-01-19T17:16:19.3379857Z         Start  49: qa_volk_32f_s32f_convertpuppet_8u
2023-01-19T17:16:19.3380174Z 
2023-01-19T17:16:19.3380537Z 49: Test command: /usr/bin/sh "/home/runner/work/volk/volk/build/lib/volk_32f_s32f_convertpuppet_8u_test.sh" "/home/runner/work/volk/volk/build/lib"
2023-01-19T17:16:19.3381285Z 49: Working Directory: /home/runner/work/volk/volk/build/lib
2023-01-19T17:16:19.3381934Z 49: Test timeout computed to be: 10000000
2023-01-19T17:16:19.3424431Z 49: RUN_VOLK_TESTS: volk_32f_s32f_convertpuppet_8u(131071,1)
2023-01-19T17:16:19.3592784Z 49: generic completed in 1.43449 ms
2023-01-19T17:16:19.3594519Z 49: u_avx2_fma completed in 0.211499 ms
2023-01-19T17:16:19.3596169Z 49: a_avx2_fma completed in 0.159799 ms
2023-01-19T17:16:19.3597839Z 49: u_avx2 completed in 0.159899 ms
2023-01-19T17:16:19.3599516Z 49: a_avx2 completed in 0.1585 ms
2023-01-19T17:16:19.3601493Z 49: u_sse2 completed in 0.169099 ms
2023-01-19T17:16:19.3603328Z 49: a_sse2 completed in 0.196499 ms
2023-01-19T17:16:19.3606218Z 49: u_sse completed in 0.281099 ms
2023-01-19T17:16:19.3609118Z 49: a_sse completed in 0.279399 ms
2023-01-19T17:16:19.3609359Z 49: offset 12550 in1: 38 in2: 39 tolerance was: 0
2023-01-19T17:16:19.3611892Z 49: volk_32f_s32f_convertpuppet_8u: fail on arch u_avx2_fma
2023-01-19T17:16:19.3614506Z 49: offset 12550 in1: 38 in2: 39 tolerance was: 0
2023-01-19T17:16:19.3617079Z 49: volk_32f_s32f_convertpuppet_8u: fail on arch a_avx2_fma
2023-01-19T17:16:19.3638580Z 49: Best aligned arch: a_avx2
2023-01-19T17:16:19.3638910Z 49: Best unaligned arch: u_avx2
2023-01-19T17:16:19.3649341Z  49/141 Test  #49: qa_volk_32f_s32f_convertpuppet_8u ....................***Failed    0.03 sec
```
Are the tests deterministic or do they use random inputs? The test seems to work fine on my machine every time.

I'm thinking that perhaps some of the inputs are close to a half-integer after applying the scale and bias, and FMA has perhaps less precision (or a different error) from non-FMA, so the FMA and non-FMA outputs might be rounded to integers differently.

jdemel:
Unfortunately, our current test infrastructure relies on random values. This needs an update. However, this is a work in progress.

daniestevez:
Then I guess that this failed on the clang-10 and clang-11 builds just by random chance with the inputs, and not because of these compilers doing something differently.

I think that it's quite likely that the FMA operations give slightly different results from first doing multiply and then add (which is the generic version). The [documentation for `_mm256_fmadd_ps()`](https://www.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/compiler-reference/intrinsics/intrinsics-for-avx2/intrinsics-for-fused-multiply-add-operations/mm-fmadd-ps-mm256-fmadd-ps.html) says:
> Corresponding values in two operands, a and b, are multiplied and the infinite precision intermediate results are added to corresponding values in the third operand, after which the final results are rounded to the nearest float32 values.

So it seems that the FMA result would be more accurate than multiply then add, because FMA conserves all the precision of the multiply, and then only rounds to `float`, whereas multiply and then add rounds to `float` first after the multiplication and then after the addition.

I can imagine that there are corner cases in which the FMA versus the multiply + add results sit at both sides of a half-integer boundary, and are rounded in different directions.

I think that it would be good to allow the test to have an error of `1` to account for this case. Do you agree? Can this be done with `test_params.make_absolute(1.0)`? Reading the code it seems to me that the tolerance is "inclusive", so a tolerance of `1.0` would allow an error up to and including 1.0.

PS: I wasn't able to make the unit test fail on my machine, even by running the test many times in a loop. So I'm not sure that I've identified correctly the reason why the tests fail, but it's the most likely explanation I can think of.

jdemel:
First, I'd assume this is a random hiccup. We had more tests that failed in the past. Maybe we need to compare them too. I re-ran these tests and now they pass. This is not a desirable solution. Though, we probably need to live with it until there's a better system in place.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+ *
+ * SPDX-License-Identifier: LGPL-3.0-or-later
+ */
+
+/*!
+ * \page volk_32f_s32f_x2_convert_8u
+ *
+ * \b Overview
+ *
+ * Converts a floating point number to an 8-bit unsigend int after applying a
```
jdemel: just a typo.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+ * multiplicative scaling factor and an additive bias.
+ *
+ * <b>Dispatcher Prototype</b>
+ * \code
+ * void volk_32f_s32f_x2_convert_8u(int8_t* outputVector, const float* inputVector,
```
jdemel: Should this be `...(uint8_t* outputVector, ...)`?

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+ *   volk_free(increasing);
+ *   volk_free(out);
+ * \endcode
+ */
+
+#ifndef INCLUDED_volk_32f_s32f_x2_convert_8u_u_H
+#define INCLUDED_volk_32f_s32f_x2_convert_8u_u_H
+
+#include <inttypes.h>
+#include <stdio.h>
```
jdemel: I assume this include can be removed now.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+        volk_32f_s32f_x2_convert_8u_single(&outputVector[number], r);
+    }
+}
+
+#endif /* LV_HAVE_SSE */
+
+
+#ifdef LV_HAVE_GENERIC
+
+static inline void volk_32f_s32f_x2_convert_8u_generic(uint8_t* outputVector,
```
jdemel: Could you move the generic kernel to the top, right below to the `_single` one? Makes it easier to find the reference.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+#if LV_HAVE_AVX2 && LV_HAVE_FMA
+#include <immintrin.h>
+
+static inline void volk_32f_s32f_x2_convert_8u_u_avx2_fma(uint8_t* outputVector,
+                                                          const float* inputVector,
+                                                          const float scale,
+                                                          const float bias,
+                                                          unsigned int num_points)
+{
+    unsigned int number = 0;
```
jdemel: Could you move this variable into the `for` loop declaration?

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+
+    const unsigned int thirtysecondPoints = num_points / 32;
+
+    const float* inputVectorPtr = (const float*)inputVector;
+    uint8_t* outputVectorPtr = outputVector;
+
+    float min_val = 0.0f;
+    float max_val = UINT8_MAX;
```
jdemel: Could you make those `const`? I assume they should never change.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+    float r;
```
jdemel: It looks like this variable is only needed in the tail loop. Thus, it should only be declared there.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+
+    __m256 vScale = _mm256_set1_ps(scale);
+    __m256 vBias = _mm256_set1_ps(bias);
```
jdemel: These could be `const`.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+    __m256 inputVal1, inputVal2, inputVal3, inputVal4;
+    __m256i intInputVal1, intInputVal2, intInputVal3, intInputVal4;
+    __m256 vmin_val = _mm256_set1_ps(min_val);
+    __m256 vmax_val = _mm256_set1_ps(max_val);
```
jdemel: These could be `const`. Also, I suggest to move the constant value declarations to the top. This helps to understand the code structure faster. e.g. `min_val` next to `vmin_val`.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+        intInputVal3 = _mm256_permute4x64_epi64(intInputVal3, 0b11011000);
+
+        intInputVal1 = _mm256_packus_epi16(intInputVal1, intInputVal3);
+        intInputVal = _mm256_permute4x64_epi64(intInputVal1, 0b11011000);
+
+        _mm256_storeu_si256((__m256i*)outputVectorPtr, intInputVal);
+        outputVectorPtr += 32;
+    }
+
+    number = thirtysecondPoints * 32;
```
jdemel: I suggest to use this variable in the following for loop and declare a new variable there. The compiler will do its thing anyways.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+    for (; number < num_points; number++) {
+        r = inputVector[number] * scale + bias;
```
jdemel: you can just do `float r = ...` here. The compiler will be happy to decide what to do.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+    }
+}
+
+#endif /* LV_HAVE_AVX2 && LV_HAVE_FMA */
+
+
+#ifdef LV_HAVE_AVX2
+#include <immintrin.h>
+
+static inline void volk_32f_s32f_x2_convert_8u_u_avx2(uint8_t* outputVector,
```
jdemel, line 168: The comments for the `avx2_fma` kernel apply to this kernel as well.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+    }
+}
+
+#endif /* LV_HAVE_AVX2 */
+
+
+#ifdef LV_HAVE_SSE2
+#include <emmintrin.h>
+
+static inline void volk_32f_s32f_x2_convert_8u_u_sse2(uint8_t* outputVector,
```
jdemel, line 243: cf. `avx_fma` comments.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+#ifdef LV_HAVE_GENERIC
+
+static inline void volk_32f_s32f_x2_convert_8u_generic(uint8_t* outputVector,
+                                                       const float* inputVector,
+                                                       const float scale,
+                                                       const float bias,
+                                                       unsigned int num_points)
+{
+    const float* inputVectorPtr = inputVector;
+    unsigned int number = 0;
```
jdemel: Could you move this into the loop?

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+    float r;
```
jdemel: This would probably be better located in the loop.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,692 @@
...
+#endif /* LV_HAVE_GENERIC */
+
+
+#endif /* INCLUDED_volk_32f_s32f_x2_convert_8u_u_H */
+#ifndef INCLUDED_volk_32f_s32f_x2_convert_8u_a_H
+#define INCLUDED_volk_32f_s32f_x2_convert_8u_a_H
+
+#include <inttypes.h>
+#include <stdio.h>
+#include <volk/volk_common.h>
```
jdemel: Are those still necessary?

jdemel (changes requested):
I went through the code and added suggestions to improve the structure. This applies to all kernels. Though, I didn't repeat the comments for every kernel.

I know that old kernels use a C89 function style. Though, I'd like to remove this whenever code is updated. This improves readability.

daniestevez:
Thanks. Most of your comments also apply to `volk_32f_s32f_convert_8i.h`, because I started with a copy&paste of that file. I'm applying all your suggestions now. If it seems helpful, I can also go and apply the same changes to that file, and submit as part of this or another PR.

jdemel:
That would be great. I suggest another PR though.

daniestevez:
I've force pushed a new commit following your suggestions. I've also applied other changes that seemed in-line with your thinking of declaring as `const` everything that can be `const` and moving declarations to the innermost scope possible (for instance, I moved all the `__m256 inputVal1`, etc. declarations inside the loop).

Let me know if this looks good, and if so, I'll apply the same changes to the 8i kernel on another PR.

daniestevez:
I think that it's still worth to increase the tolerance of the test of this kernel. Otherwise, from the moment that this gets merged to main, each new commit is likely to trigger this test to fail on at least one of the builds, which is potentially annoying and confusing.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
```diff
@@ -0,0 +1,635 @@
...
+#endif /* INCLUDED_volk_32f_s32f_x2_convert_8u_u_H */
+#ifndef INCLUDED_volk_32f_s32f_x2_convert_8u_a_H
+#define INCLUDED_volk_32f_s32f_x2_convert_8u_a_H
+
+#include <inttypes.h>
+#include <volk/volk_common.h>
+
+#ifdef LV_HAVE_GENERIC
+
+static inline void volk_32f_s32f_x2_convert_8u_a_generic(uint8_t* outputVector,
```
jdemel: The `_a_generic` kernel can just be deleted. That's another remainder of obsolete code.

jdemel (commented):
We could introduce a `test_convert2int_params` version around here:
https://github.com/gnuradio/volk/blob/8a226321b1e1892c67c79c03e4b836d6f8553711/lib/kernel_tests.h#L47
Then, set the tolerance to `1`. However, since the other convert kernels seem to not encounter this issue, I could imagine, we'd just wait and see. We might introduce this in another PR.

I found one last thing. Besides, it looks good to me.

kernels/volk/volk_32f_s32f_x2_convert_8u.h
daniestevez: Just deleted the kernel and force-pushed.

daniestevez:
Okay. Let's wait to see if the tests keep giving problems and will fix them in another PR if so.

I've deleted the `a_generic` kernel. I'll send another PR with the same changes for the `8i` kernel in one or two days.

jdemel (approved):
LGTM! Thanks!

jdemel:
@daniestevez Thanks for your patience. Merging now.

PR #619, opened 2023-02-28
Add 32f_s32f_multiply_32f RISC-V manually optimized assembly

ThomasHabets:
I don't know volk wants hand optimized assembly when there's no SIMD or vector instructions used, but a <strike>35% gain is a 35% gain</strike>56% gain is a 56% gain, right?

### DIFF
--- a/gen/archs.xml
+++ b/gen/archs.xml
@@ -1,4 +1,9 @@
-<!-- archs appear in order of significance for blind, de-facto version ordering -->
+<!--
+Archs appear in order of significance for blind, de-facto version ordering.
+
+In other words when available, an arch further down will be used. "generic" is
+at the top, as a last resort.
+-->
 <grammar>
 
 <arch name="generic"> <!-- name is required-->
@@ -171,6 +176,9 @@
     <flag compiler="clang">-mavx512cd</flag>
     <flag compiler="msvc">/arch:AVX512CD</flag>
     <alignment>64</alignment>
+</arch>
+
+<arch name="riscv64">
 </arch>
 
 </grammar>
--- a/gen/machines.xml
+++ b/gen/machines.xml
@@ -29,6 +29,10 @@
 <archs>generic 32|64| mmx| sse sse2 sse3 ssse3 orc|</archs>
 </machine>
 
+<machine name="sifive_u74">
+<archs>generic riscv64 orc|</archs>
+</machine>
+
 <machine name="sse4_a">
 <archs>generic 32|64| mmx| sse sse2 sse3 sse4_a popcount orc|</archs>
 </machine>
--- a/kernels/volk/asm/riscv/volk_32f_s32f_multiply_32f_sifive_u74.s
+++ b/kernels/volk/asm/riscv/volk_32f_s32f_multiply_32f_sifive_u74.s
@@ -1,0 +1,82 @@
+        .text
+        .align 2
+        .type volk_32f_s32f_multiply_32f_sifive_u74, @function
+        .global volk_32f_s32f_multiply_32f_sifive_u74
+
+volk_32f_s32f_multiply_32f_sifive_u74:
+        # Input:
+        # a0  out
+        # a1  in
+        # fa0 scalar
+        # a2  size
+
+        # Main loop in 8x unrolled.
+
+        # Split counter into main and final loop.
+        # a5  main loop counter
+        # a2  closing loop counter
+        srli    a5,a2,3
+        andi    a2,a2,7
+        slli    a5,a5,5
+        beqz    a5,.dolastloop
+        add     a5,a0,a5
+
+        .align 2
+.loop:
+        flw     fa1,0(a1)
+        addi    a0,a0,32      # increment output (free, running on pipeline A)
+
+        flw     fa2,4(a1)
+        flw     fa3,8(a1)
+        flw     fa4,12(a1)
+        flw     fa5,16(a1)
+        flw     fa6,20(a1)
+        flw     fa7,24(a1)
+        flw     ft8,28(a1)
+        addi    a1,a1,32      # increment input (free, running on pipeline A)
+
+        fmul.s  fa1,fa1,fa0
+        fmul.s  fa2,fa2,fa0
+        fmul.s  fa3,fa3,fa0
+        fmul.s  fa4,fa4,fa0
+        fmul.s  fa5,fa5,fa0
+        fmul.s  fa6,fa6,fa0
+        fmul.s  fa7,fa7,fa0
+        fmul.s  ft8,ft8,fa0
+
+        fsw     fa1,-32(a0)
+        fsw     fa2,-28(a0)
+        fsw     fa3,-24(a0)
+        fsw     fa4,-20(a0)
+        fsw     fa5,-16(a0)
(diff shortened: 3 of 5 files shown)


drmpeg:
If ```liborc-0.4-0``` is installed, volk_profile doesn't find the new kernel.
```
RUN_VOLK_TESTS: volk_32f_s32f_multiply_32f(131071,1987)
generic completed in 2181.57 ms
a_generic completed in 2182.21 ms
u_orc completed in 3570.61 ms
Best aligned arch: generic
Best unaligned arch: generic
```

ThomasHabets:
Good catch. Weird, though. Sigh, back to reverse engineering code generators. :-(

ThomasHabets:
Ok, got it. `archs.xml` is ordered. The comment at the top intends to say that, but is extremely unclear.

ThomasHabets:
Ok, try now.

drmpeg:
```visionfive2``` is now being selected for ```volk_32f_s32f_multiply_32f```, but kernels that previously selected ```orc``` are not selecting ```orc``` now. For example:
old behavior:
```
volk_profile -R volk_16ic_deinterleave_real_8i

RUN_VOLK_TESTS: volk_16ic_deinterleave_real_8i(131071,1987)
generic completed in 1169.25 ms
u_orc completed in 1166.62 ms
Best aligned arch: u_orc
Best unaligned arch: u_orc
```
new behavior:
```
volk_profile -R volk_16ic_deinterleave_real_8i
 
RUN_VOLK_TESTS: volk_16ic_deinterleave_real_8i(131071,1987)
no architectures to test
```

**Comment by drmpeg** [comment]
volk_profile looks good now.

Just a minor nit. volk_32f_s32f_multiply_32f_visionfive2.s has mixed tabs and spaces and trailing white space on some lines.

I guess the preference is spaces, but there are tabs in some of the NEON asm files.

**Comment by ThomasHabets** [comment]
It should at least be consistent within a file. Switched to all spaces.

The checks seem to be failing due to trying to use the RISC-V function, so something is still not right. I confirmed that I'm getting the error on my x86-64 too.

So something needs to be fixed with either the cmake or the codegen.

**Comment by ThomasHabets** [comment]
I think the two failing tests are flaky. It's two different ones failing.

Potentially this is now good to go, if the arch/machine naming of riscv64/visionfive2 is a thing that makes sense.

Though I may be relying too much on any existing tests. I'll want to test manually a bit.

**Comment by drmpeg** [comment]
It looks good and I agree the failing tests seems spurious. ```visionfive2``` seems a little too specific to me (after all, I'm testing it on the HiFive Unmatched), but naming is difficult. Maybe sifive-u74 instead?

**Comment by ThomasHabets** [comment]
Ugly test program that seems to work:

```c++
#include<iostream>
#include<vector>
#include<cassert>
#include<cmath>
extern "C" void volk_32f_s32f_multiply_32f_visionfive2(float* cVector, const float* aVector, const float scalar, unsigned int num_point\s);

namespace {
  void  check(const std::vector<float> &a, const std::vector<float> &b)
  {
    bool match = true;
    for (int i = 0; i < a.size(); i++) {
      if (fabs(a[i] - b[i]) > 0.0001) {
        match = false;
        break;
      }
    }
    if (match) {
      return;
    }
    std::cerr << "Mismatch:\n";
    for (auto e : a) {
      std::cerr << e << " ";
    }
    std::cerr << "\n";
    for (auto e : b) {
      std::cerr << e << " ";
    }
    std::cerr << "\n";
  }
}

int main()
{
  std::vector<float> in, out;
  std::cerr << "empty\n";
  volk_32f_s32f_multiply_32f_visionfive2(out.data(), in.data(), 3.0, in.size());

  std::cerr << "one\n";
  in.push_back(2.0);
  out.clear();
  out.resize(in.size());
  volk_32f_s32f_multiply_32f_visionfive2(out.data(), in.data(), 3.0, in.size());
  assert(out == std::vector<float>{6.0});

  std::cerr << "two\n";
  in.push_back(2.2);
  out.clear();
  out.resize(in.size());
  volk_32f_s32f_multiply_32f_visionfive2(out.data(), in.data(), 3.0, in.size());
  check(out, std::vector<float>{6.0, 6.6});

  std::cerr << "three\n";
  in.push_back(-3.1);
  out.clear();
  out.resize(in.size());
  volk_32f_s32f_multiply_32f_visionfive2(out.data(), in.data(), 3.0, in.size());
  check(out, std::vector<float>{6.0, 6.6, -9.3});
}
```

drmpeg:
Tested with GNU Radio multiply const block and fast noise source. Output files match perfectly.

ThomasHabets:
Looks like unrolling more given even more speed. I'll rename it and will polish the assembly and update.

ThomasHabets:
Ok, now it's renamed, and unrolled to 8x. I also tried unrolling to 16x, but it didn't seem to add any value.

jdemel:
@ThomasHabets I'd like to add RiscV support. As far as I can see, this includes a few tweaks in places other than kernels as well. That's a valuable change. The 35% speed improvement is nice as well.

After reading through your conversation, I'm a bit worried if things break one way or the other. However, all tests seem to pass. (Including the RiscV test.) 
How would you spell it correctly? RiscV? RISC-V? riscv64? ...?

ThomasHabets:
56% less time now, so more than twice as fast as GCC.

* The risc-v.org site calls the whole thing RISC-V.
* Linux kernel source code uses "riscv" as directory names.
* `uname -m` is `riscv64`.
* The [risc-v ISA docs](https://riscv.org/wp-content/uploads/2017/05/riscv-spec-v2.2.pdf) calls it RV64.

Probably this assembly will actually build just fine under RV32, but I don't know if there is or will ever be Linux-compatible 32bit RISC-V, or if it's just for embedded stuff. (especially since we here assume RISC-V with floating point support)

While this code makes no assumptions about the width of general purpose registers or addresses, I think this may not be true in the future by default. It's also of course untested on 32bit RISC-V.

I think it's safest to go with riscv64 for now, but looks to me like bowing to a future consensus name should not be a breaking change.

jdemel:
I'd stick with `riscv64`. I wouldn't expect any `riscv32` CPUs that run VOLK. If this ever happens, we'd need to think of it.

As a side note: [RISC-V autovectorization might be available in GCC 14](https://www.phoronix.com/news/GCC-RISC-V-Auto-Vectorization)

Are you guys sure that all the subtle bugs, e.g. shadowing ORC is fixed? If so, I'd like to merge this PR.

ThomasHabets:
I'm as sure as I can be that the ORC issue is fixed. But this being my first contribution to volk it's not a huge amount of certainty.

Thanks for the GCC 14 link. Nice to hear. But it won't help current hardware, since I believe no hardware currently has the vector instructions.

But with (maybe) GCC 14, on hardware that has vector instructions, it's expected that this code won't be chosen by `volk_profile`. Potentially in the future even non-vector GCC may produce much better code, too.

But I guess the above goes for the whole volk project. A "sufficiently smart compiler" will be able to create AVX512 from vanilla C++ code, too, where appropriate.

jdemel:
Thanks! Merging it now. I assume this PR got due time to be reviewed.

PR #620, opened 2023-04-10
Add volk_32fc_x2_dot_prod_32fc_sifive_u74

ThomasHabets:
~41% less CPU used.

RUN_VOLK_TESTS: volk_32fc_x2_dot_prod_32fc(131071,1987) 
sifive_u74 completed in 1903.6 ms
generic completed in 2689.21 ms
a_generic completed in 2687.77 ms
Best aligned arch: sifive_u74
Best unaligned arch: sifive_u74

### DIFF
--- a/kernels/volk/asm/riscv/volk_32fc_x2_dot_prod_32fc_sifive_u74.s
+++ b/kernels/volk/asm/riscv/volk_32fc_x2_dot_prod_32fc_sifive_u74.s
@@ -1,0 +1,101 @@
+        .text
+        .align 2
+        .type   volk_32fc_x2_dot_prod_32fc_sifive_u74, @function
+        .global volk_32fc_x2_dot_prod_32fc_sifive_u74
+
+        #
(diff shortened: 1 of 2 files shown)


jdemel (approved):
LGTM. I really hope these never break. It'd be difficult to fix any of these.

It might be a good time to add smth like QEMU VMs and instructions to set up the whole build chain for different architectures.

ThomasHabets:
As assembly goes this is actually not that tricky. The most advanced instructions here are the fused multiply add. What makes it a bit harder to read is the interleaving.

Potentially the interleaving could in the future become a preprocessing stage, making the source easier to read.

Longer term future hardware should get vector instructions, which will hopefully will either be detected by a "sufficiently smart compiler", or get some `builtin` support like SIMD stuff has.

Edit:
Looks like clang-trunk is at least partly a sufficiently smart compiler for vector instructions: https://godbolt.org/z/bvPon5KWf

PR #623, opened 2023-05-07
cpu_features: Update submodule pointer

michael-roe:


jdemel:
Thanks for the update.

PR #631, opened 2023-09-17
Update cpu features

### LINKED ISSUE #607: cmake doesn't check for cpu_features library being installed
Build fails with the following error if cpu_features not installed, which cmake should check:

Consolidate compiler generated dependencies of target volk_obj
[ 88%] Built target volk_obj
[ 89%] Linking C shared library libvolk.so
/usr/bin/ld: cannot find -lcpu_features
collect2: error: ld returned 1 exit status
make[2]: *** [lib/CMakeFiles/volk.dir/build.make:168: lib/libvolk.so.2.5.2] Error 1
make[1]: *** [CMakeFiles/Makefile2:239: lib/CMakeFiles/volk.dir/all] Error 2
make: *** [Makefile:146: all] Error 2

### LINKED ISSUE #629: Build error -  Target "CpuFeatures::cpu_features" not found.
```
CMake Error at lib/CMakeLists.txt:524 (target_include_directories):
  Error evaluating generator expression:

    $<TARGET_PROPERTY:CpuFeatures::cpu_features,INTERFACE_INCLUDE_DIRECTORIES>

  Target "CpuFeatures::cpu_features" not found.
```
This **CpuFeatures::cpu_features** should be **CpuFeature::cpu_features**

Also this if/else block seem redundant to me, in both branches it is including cpu_features far as I can tell
https://github.com/gnuradio/volk/blob/efa2220e6f0b50724618b006457f97eb748327e7/lib/CMakeLists.txt#L523C1-L531C10

ashley-b:
- Update cpu_features to v0.9.0
- Added BUILD_EXECUTABLE for cpu_features to build list_cpu_features for
  github workflow tests
- Added ENABLE_INSTALL to install cpu_feature for old behaviour
- Renamed CpuFeature -> CpuFeatures to match new name
- Removed duplicated cmake logic


ashley-b:
Related issue #629

CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ if (VOLK_CPU_FEATURES)
       FORCE)
     set(BUILD_SHARED_LIBS_SAVED "${BUILD_SHARED_LIBS}")
     set(BUILD_SHARED_LIBS OFF)
+    set(ENABLE_INSTALL ON)
```
jdemel: I suggest to go for `set(ENABLE_INSTALL OFF)` because `cpu_features` should never be installed with VOLK. This causes issue #607 . Since we don't need the `cpu_features` library and don't expose any of its APIs, the `NO install` solution would be preferred.

jdemel (changes requested):
One tiny change. But besides this looks good. Thanks for the PR!

CMakeLists.txt
ashley-b: I  did try leaving this off before I submit this PR, but it break's the static CI build. This make's sense since the finial stage link needs the cpu_feature library.

One solution to this would be to ENABLE_INSTALL for cpu_features if ENABLE_STATIC_LIBS is set

jdemel: I try to add a bit more context here. I suppose package maintainers, e.g. for Debian, explicitly add patches to make sure `cpu_features` is not installed together with with VOLK.
Furthermore, VOLK should not slip into a maintainer role for a dependency. Even if this implies to update the submodule pointer every once in a while. 
As a consequence, we should try to:
1. hide all dependencies as private and not expose them in our API (if possible)
2. not install anything that is not directly VOLK.

Having written all of this, what would be a reasonable approach here?
It's "good news" that only the static build breaks. Right now, I see that the static test fails because one of the tests fails. I suggest to set `ENABLE_INSTALL OFF` and then we would be able to discuss the situation in more detail.
I would really prefer to not install anything related to `cpu_features`.

ashley-b: All good, I've added the change.

As a follow up
It does the right thing when cpu_features is installed as a system package on Fedora distro., the volk shared library dynamical links to the the system cpu_features and it does not build the submodule version

jdemel: I asked in the gnuradio matrix chat. I received this hint from @marcusmueller
https://discourse.cmake.org/t/cmake-error-install-export-projecttargets-includes-target-xxx-which-requires-target-yyy-that-is-not-in-any-export-set/7344/3

jdemel: If I read this discussion correctly, this line:
```cmake
target_link_libraries(volk_static PRIVATE CpuFeatures::cpu_features)
```
though, I haven't had the time to fully understand it yet.

noc0lour:
I hope you don't mind me adding another commit fixing the issues with the build. Though it would be useful to add a test building an "external" binary linking to the dynamic & static builds to see that everything works correctly with this linking. Maybe something for the issue tracker.

jdemel (approved):
All tests pass. LGTM. Fixes important things. Thank you all!

jdemel:
@ashley-b if you're OK with the state of this PR including @noc0lour 's change, I'd merge it.

ashley-b:
@jdemel Sorry for the delay, Im happy with the PR. 
Not sure about @noc0lour commit. I could not see any changes in the github Actions build/tests or local builds on my system.

Am not a CMake expert, but I not sure BUILD_INTERFACE has any effect in target_link_libraries.

PR #637, opened 2023-10-04
fix compiler warnings

Ka-zam:
Fix some annoying compiler warnings when using uninitialized variables.


jdemel (approved):
This is a deprecated kernel because we don't test it and we don't know if anyone actually uses it.

https://github.com/gnuradio/volk/blob/a26a1b808ddd75baeebe660e324f23c36dec6416/tmpl/volk.tmpl.h#L66

Getting rid of the compiler warnings seems right.

PR #639, opened 2023-10-12
Disable ORC kernels that are broken on ARM

### LINKED ISSUE #622: ORC implementation of volk_32fc_magnitude_32f returns incorrect results on ARM
Some users have noticed that Gqrx's AM demodulator does not work properly on the Raspberry Pi:

* https://github.com/gqrx-sdr/gqrx/issues/570
* https://github.com/gqrx-sdr/gqrx/issues/1191

I dug into this, and found that on 32-bit ARM, `volk_32fc_magnitude_32f`'s `u_orc` implementation produces an output of `NaN` when the input is zero. A sample of `NaN` then corrupts the state of an IIR filter, causing all of its future output to be `NaN`.

It appears the root cause is that ORC's `sqrtf` instruction produces `NaN` when the input value is zero, on 32-bit ARM. ORC's test suite even makes an explicit exception for this case:

https://gitlab.freedesktop.org/gstreamer/orc/-/blob/72b4699314e0a6eeeb29cea33d0410612c80f533/orc-test/orctest.c#L587-591

Thus it would seem that it is not safe for VOLK to pass zero into `sqrtf`.

Several other kernels also use `sqrtf`:

* volk_16ic_magnitude_16i
* volk_16ic_magnitude_32f
* volk_32f_sqrt_32f

I would presume they are similarly broken on 32-bit ARM.

argilo:
Fixes #622.

Both `volk_32f_sqrt_32f_a_orc_impl` and `volk_32fc_magnitude_32f_a_orc_impl` use the ORC `sqrtf` instruction, which is broken on ARM (`sqrt(0) = NaN`):

https://gitlab.freedesktop.org/gstreamer/orc/-/blob/72b4699314e0a6eeeb29cea33d0410612c80f533/orc-test/orctest.c#L587-591

Here I've disabled the broken kernels on ARM.

### DIFF
--- a/kernels/volk/volk_32f_sqrt_32f.h
+++ b/kernels/volk/volk_32f_sqrt_32f.h
@@ -168,6 +168,7 @@ volk_32f_sqrt_32f_generic(float* cVector, const float* aVector, unsigned int num
 
 
 #ifdef LV_HAVE_ORC
+#ifndef LV_HAVE_NEON /* ORC sqrtf instruction is broken */
 
 extern void volk_32f_sqrt_32f_a_orc_impl(float*, const float*, unsigned int);
 
@@ -177,6 +178,7 @@ volk_32f_sqrt_32f_u_orc(float* cVector, const float* aVector, unsigned int num_p
     volk_32f_sqrt_32f_a_orc_impl(cVector, aVector, num_points);
 }
 
+#endif /* LV_HAVE_NEON */
 #endif /* LV_HAVE_ORC */
 
 #endif /* INCLUDED_volk_32f_sqrt_32f_a_H */
--- a/kernels/volk/volk_32fc_magnitude_32f.h
+++ b/kernels/volk/volk_32fc_magnitude_32f.h
@@ -440,6 +440,7 @@ static inline void volk_32fc_magnitude_32f_neon_fancy_sweet(
 
 
 #ifdef LV_HAVE_ORC
+#ifndef LV_HAVE_NEON /* ORC sqrtf instruction is broken */
 
 extern void volk_32fc_magnitude_32f_a_orc_impl(float* magnitudeVector,
                                                const lv_32fc_t* complexVector,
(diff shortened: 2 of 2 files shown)


argilo:
Oops, this doesn't actually build properly on ARM. Changing to draft while I figure out another way.

argilo:
Hmm, it looks like `volk_kernel_defs.py` just finds all the `LV_HAVE_*` strings to determine which architectures are supported. I don't immediately see a way exclude an ORC kernel on a single platform, so I'll close this off unless I come up with something.

argilo:
Could we consider removing these ORC kernels for all platforms?

jdemel:
The ORC kernels are usually slower than the others. As far as I can recall, the generic kernel is often faster as well. They cause trouble. I suggest we can remove them. Especially because of the `sqrt(0) == NaN` issue.  Feel free to reopen and remove the ORC kernels.

argilo:
> The ORC kernels are usually slower than the others.

That's true on architectures that have native kernels (x86 & ARM), but I expect the ORC kernels could give a significant performance boost on other architectures. That's why I hesitated to remove them completely.

I notice that NEON appears before ORC in `archs.xml`. If I understand correctly, that means that ORC kernels will be preferred if `volk_profile` has not been used.  Moving NEON below ORC would at least prevent the broken ORC kernels from running by default.

As a test, I ran `volk_profile` on a Raspberry Pi 4 (aarch64), and the only two cases where ORC beat NEON were:

* `volk_16ic_deinterleave_real_8i` (144 vs 136 ms)
* `volk_32f_x2_min_32f` (525 vs 461 ms)

In the other 17 cases, NEON was faster than ORC, often by a small margin, but sometimes by a significant margin.

So I think it would make sense to move NEON below ORC in `archs.xml`. That would also (mostly) resolve #622, although I suppose you could still select the broken ORC kernels manually. It's unlikely that `volk_profile` would choose them, since they're significantly slower than NEON.

argilo:
I opened #640 to swap the ordering of ORC & NEON, since I think that makes sense regardless of what we do with the ORC kernels that are broken on ARM.

PR #640, opened 2023-10-12
Prefer NEON kernels over ORC

argilo:
Currently ORC appears below NEON in `archs.xml`, which means that ORC kernels will be preferred if `volk_profile` has not been executed to profile the kernels. As noted in https://github.com/gnuradio/volk/pull/639#issuecomment-1759754136, in most cases NEON kernels outperform ORC kernels (sometimes by a significant margin), so I think it makes sense to move NEON below ORC. This will also have the side benefit of mostly resolving #622, because the ORC kernels that are broken on ARM (`volk_32f_sqrt_32f_a_orc_impl` and `volk_32fc_magnitude_32f_a_orc_impl`) will no longer be selected by default, and are unlikely to be selected by `volk_profile` as they are significantly slower than their NEON counterparts.

### DIFF
--- a/gen/archs.xml
+++ b/gen/archs.xml
@@ -19,29 +19,6 @@ at the top, as a last resort.
   <flag compiler="clang">-mfloat-abi=hard</flag>
 </arch>
 
-<arch name="neon">
-  <flag compiler="gnu">-funsafe-math-optimizations</flag>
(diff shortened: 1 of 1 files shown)


jdemel (approved):
LGTM.

Regarding #622 , we should probably remove the offending ORC implementation entirely. We can't rely on it to work as expected.

PR #641, opened 2023-10-12
Fix include guards

argilo:
The `volk_32f_s32f_add_32f_generic` and `volk_32fc_s32f_magnitude_16i_generic` kernels are defined outside their header files' include guards, which prevents `volk_kernel_defs.py` from finding them.

After moving these kernels inside the include guards, they become accessible.


jdemel (approved):
LGTM. Very reasonable.

PR #642, opened 2023-10-12
Add length checks to volk_8u_x2_encodeframepolar_8u

### LINKED ISSUE #414: Problem with POLAR encoding/decoding blocks
It seems newer version of volk got issue with POLAR encoding/decoding blocks as well.
if you open examples in gr-fec/examples/polar_code_example.grc & fecapi_polar_decoders.grc you can observe that source info bits and decoded bits are not same.
it seems the bug comes from the same point because #395 and #3376 in gnuradio git have same issues.
workaround as other issues with same effect:
1. run volk_profile
2. change volk_8u_x3_encode_polar_8u_x2 in ~/.volk/volk_config to generic generic
3. restart gnuradio

argilo:
Fixes #414.

The various `volk_8u_x2_encodeframepolar_8u` implementations have warnings about the input size:

https://github.com/gnuradio/volk/blob/a26a1b808ddd75baeebe660e324f23c36dec6416/kernels/volk/volk_8u_x2_encodeframepolar_8u.h#L157-L158

https://github.com/gnuradio/volk/blob/a26a1b808ddd75baeebe660e324f23c36dec6416/kernels/volk/volk_8u_x2_encodeframepolar_8u.h#L430-L431

https://github.com/gnuradio/volk/blob/a26a1b808ddd75baeebe660e324f23c36dec6416/kernels/volk/volk_8u_x2_encodeframepolar_8u.h#L692-L693

https://github.com/gnuradio/volk/blob/a26a1b808ddd75baeebe660e324f23c36dec6416/kernels/volk/volk_8u_x2_encodeframepolar_8u.h#L964-L965

So let's heed those warnings by falling back to the generic implementation if the input size is too small.

Doing so also removes the need for extra size checks in `volk_32f_8u_polarbutterfly_32f.h`.

After this change, all failures disappear. I tested with the following:

```
for i in {1..64}; do apps/volk_profile -n -i 1 -v $i -R encodepolar | grep fail; done
```

I also verified that all valgrind warnings disappear after these changes. I tested with:

```
for i in {1..64}; do valgrind apps/volk_profile -n -i 1 -v $i -R encodepolar | grep fail; done
```

### DIFF
--- a/kernels/volk/volk_32f_8u_polarbutterfly_32f.h
+++ b/kernels/volk/volk_32f_8u_polarbutterfly_32f.h
@@ -232,11 +232,7 @@ static inline void volk_32f_8u_polarbutterfly_32f_u_avx(float* llrs,
         unsigned char* u_temp = u + 2 * frame_size;
         memcpy(u_temp, u + u_num - stage_size, sizeof(unsigned char) * stage_size);
 
-        if (stage_size > 15) {
-            volk_8u_x2_encodeframepolar_8u_u_ssse3(u_target, u_temp, stage_size);
-        } else {
-            volk_8u_x2_encodeframepolar_8u_generic(u_target, u_temp, stage_size);
-        }
+        volk_8u_x2_encodeframepolar_8u_u_ssse3(u_target, u_temp, stage_size);
 
         src_llr_ptr = llrs + (max_stage_depth + 1) * frame_size + row - stage_size;
         dst_llr_ptr = llrs + max_stage_depth * frame_size + row;
@@ -331,11 +327,7 @@ static inline void volk_32f_8u_polarbutterfly_32f_u_avx2(float* llrs,
         unsigned char* u_temp = u + 2 * frame_size;
         memcpy(u_temp, u + u_num - stage_size, sizeof(unsigned char) * stage_size);
 
-        if (stage_size > 15) {
-            volk_8u_x2_encodeframepolar_8u_u_ssse3(u_target, u_temp, stage_size);
-        } else {
-            volk_8u_x2_encodeframepolar_8u_generic(u_target, u_temp, stage_size);
-        }
+        volk_8u_x2_encodeframepolar_8u_u_avx2(u_target, u_temp, stage_size);
 
         src_llr_ptr = llrs + (max_stage_depth + 1) * frame_size + row - stage_size;
         dst_llr_ptr = llrs + max_stage_depth * frame_size + row;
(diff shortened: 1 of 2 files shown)


kernels/volk/volk_32f_8u_polarbutterfly_32f.h
```diff
@@ -331,11 +327,7 @@ static inline void volk_32f_8u_polarbutterfly_32f_u_avx2(float* llrs,
         unsigned char* u_temp = u + 2 * frame_size;
         memcpy(u_temp, u + u_num - stage_size, sizeof(unsigned char) * stage_size);
 
-        if (stage_size > 15) {
-            volk_8u_x2_encodeframepolar_8u_u_ssse3(u_target, u_temp, stage_size);
-        } else {
-            volk_8u_x2_encodeframepolar_8u_generic(u_target, u_temp, stage_size);
-        }
+        volk_8u_x2_encodeframepolar_8u_u_avx2(u_target, u_temp, stage_size);
```
argilo, line 330: May as well call the AVX2 version here, since we're in `volk_32f_8u_polarbutterfly_32f_u_avx2`.

argilo:
I also confirmed that GNU Radio's `qa_polar_decoder_sc_systematic_test` test passes after this change, even when `VOLK_GENERIC=1` is removed:

```diff
diff --git a/cmake/Modules/GrTest.cmake b/cmake/Modules/GrTest.cmake
index b5b2c196ec..f31e199067 100644
--- a/cmake/Modules/GrTest.cmake
+++ b/cmake/Modules/GrTest.cmake
@@ -67,7 +67,7 @@ function(GR_ADD_TEST test_name)
     # We add it to the beginning of the list to use locally-built modules before installed ones.
     list(INSERT pypath 0 "${PROJECT_BINARY_DIR}/test_modules")
 
-    set(environs "VOLK_GENERIC=1" "GR_DONT_LOAD_PREFS=1" "srcdir=${srcdir}"
+    set(environs "GR_DONT_LOAD_PREFS=1" "srcdir=${srcdir}"
                  "GR_CONF_CONTROLPORT_ON=False")
     list(APPEND environs ${GR_TEST_ENVIRONS})
 
```

jdemel (approved):
LGTM. When I implemented these kernels, these very small sizes seemed to be useless for an FEC code. But people stumble over it anyways. Thus, we should fix that.

argilo:
If GNU Radio's `qa_polar_decoder_sc_systematic_test` test uses an unrealistic code size, maybe it should be fixed.

argilo:
Another motivation for making this change is that it would allow a CI test to sweep over small vector lengths (which are prone to bugs) in all kernels.

PR #643, opened 2023-10-13
New kernel for reciprocal

Ka-zam:
New kernel that calculates the reciprocal:

```
volk_profile -R rec -i 10000

RUN_VOLK_TESTS: volk_32f_reciprocal_32f(131071,10000)
generic completed in 119.092 ms
a_avx2_fma completed in 94.7562 ms
a_avx completed in 97.1054 ms
a_avx_div completed in 118.053 ms
u_avx2_fma completed in 97.3939 ms
u_avx completed in 98.4622 ms
u_avx_div completed in 129.018 ms
Best aligned arch: a_avx2_fma
Best unaligned arch: u_avx2_fma
```


include/volk/volk_avx2_fma_intrinsics.h
```diff
@@ -16,6 +16,16 @@
 #define INCLUDE_VOLK_VOLK_AVX2_FMA_INTRINSICS_H_
 #include <immintrin.h>
 
+/*
+ * First order Newton-Raphson approximation of 1 / x
+ */
+static inline __m256 _mm256_rcp1_avx2_fma_ps(const __m256 x)
```
jdemel: why do you call this function `rcp1`? a more descriptive name would help others in the future.

kernels/volk/volk_32f_reciprocal_32f.h
```diff
@@ -0,0 +1,228 @@
...
+#define INCLUDED_volk_32f_reciprocal_32f_a_H
+
+#ifdef LV_HAVE_GENERIC
+static inline void
+volk_32f_reciprocal_32f_generic(float* out, const float* in, unsigned int num_points)
+{
+    unsigned int i = 0;
+    for (; i < num_points; i++) {
+        out[i] = 1.f / in[i];
+    }
```
jdemel: I suggest to initialize the loop variable in the for loop.

and you can work with `*out++` etc. if you want.

kernels/volk/volk_32f_reciprocal_32f.h
```diff
@@ -0,0 +1,228 @@
...
+        __m256 x_inv = _mm256_rcp1_avx2_fma_ps(x);
+
+        _mm256_store_ps(out, x_inv);
+        out += 8;
+    }
+
+    number = eighth_points * 8;
+    for (; number < num_points; number++) {
+        *out++ = 1.f / (*in++);
+    }
```
jdemel: This would probably just be a call to the generic implementation.

kernels/volk/volk_32f_reciprocal_32f.h
```diff
@@ -0,0 +1,228 @@
...
+#endif /* LV_HAVE_GENERIC */
+
+
+#if LV_HAVE_AVX2 && LV_HAVE_FMA
+#include <immintrin.h>
+#include <volk/volk_avx2_fma_intrinsics.h>
+static inline void
+volk_32f_reciprocal_32f_a_avx2_fma(float* out, const float* in, unsigned int num_points)
+{
+    unsigned int number = 0;
```
jdemel: You can just move this variable into the for loop initialization part. This makes the code more readable. The same comment applies to the other kernels.

kernels/volk/volk_32f_reciprocal_32f.h
```diff
@@ -0,0 +1,228 @@
...
+    const unsigned int eighth_points = num_points / 8;
+    const __m256 ONE = _mm256_set1_ps(1.0f);
+
+    for (; number < eighth_points; number++) {
+        __m256 x = _mm256_loadu_ps(in);
+        in += 8;
+
+        __m256 x_inv = _mm256_div_ps(ONE, x);
```
jdemel: I wonder why you would just use a `div` in this kernel but implement an approximation for the other kernels.

jdemel (changes requested):
This looks like an interesting kernel. I have a few comments.
My biggest worry would be to see this kernel fail randomly with some `1.0f / 1e-xx` value.

kernels/volk/volk_32f_reciprocal_32f.h
Ka-zam: Agreed. I have nothing against pointers but it does look a bit cleaner like this I think. Performance is identical.

kernels/volk/volk_32f_reciprocal_32f.h
Ka-zam: I used a different variable and there seems to be no performance penalty.

kernels/volk/volk_32f_reciprocal_32f.h
Ka-zam: Ah, just for testing. Removed.

Ka-zam:
> This looks like an interesting kernel. I have a few comments. My biggest worry would be to see this kernel fail randomly with some `1.0f / 1e-xx` value.

I ran the code for some special values: 
```
	x	x		generic		div		rcp		avx2_fma		avx
-0.000000e+00	 -0x0p+0	    -inf	    -inf	    -inf	    -nan	    -nan
+0.000000e+00	 +0x0p+0	    +inf	    +inf	    +inf	    -nan	    -nan
+3.673420e-40	+0x1p-131	    +inf	    +inf	    +inf	    -inf	    -inf
+5.877472e-39	+0x1p-127	+1.701412e+38	+1.701412e+38	    +inf	    -inf	    -inf
+1.175494e-38	+0x1.fffffp-127	+8.507063e+37	+8.507063e+37	    +inf	    -inf	    -inf
+1.175494e-38	+0x1p-126	+8.507059e+37	+8.507059e+37	+8.504982e+37	+8.507059e+37	+8.507059e+37
+4.253530e+37	+0x1p+125	+2.350989e-38	+2.350989e-38	+2.350415e-38	+2.350989e-38	+2.350989e-38
    +inf	    +inf	+0.000000e+00	+0.000000e+00	+0.000000e+00	    -nan	    -nan
```

div is the _mm256_div_ps instruction
rcp is the _mm256_rcp_ps instruction

So, the proposed implementation fails for special values more or less spectactularly. However, it's 20% faster for valid inputs and more accurate than _mm256_rcp_ps.

_mm256_rcp_ps is 10% faster still but not as accurate in the valid domain.

Ka-zam:
I did some more experiments and tried to add checks for nan, 0.0 and inf. Turns out it's not worth it and then it's better to just use div_ps()

So, 30% faster but 
- returns nan instead of inf for 0.0 input
- returns nan instead of 0.0 for inf input

jdemel:
How do we prevent that this kernel is constantly causing random CI failures? Values close to `0.0f` are likely in our tests. We'd need to make sure that we don't see flaky kernel issues raised.

Also, the `inf` vs `nan` issue needs to be discussed. If I'd use this kernel I'd expect that `1.0f / inf = 0.0f`. This is not consistent across all implementations. This would probably cause issues in the future.
We had a similar issue in #622 and eventually had to remove some kernel implementations #658 . I'm worried this kernel will see the same fate. We should find a solution for this issue before we merge it.

Ka-zam:
Values close to 0.f are no problem, it's only a very slight problem for subnormals where there's decreased accuracy.

```inf``` vs ```nan``` is more of a problem. Perhaps that can be looked at but the problem is we can't do too much work before the div_ps( 1.f / x )  instruction is faster.

I will investigate.

PR #654, opened 2023-10-14
Use absolute tolerance for sum_of_poly kernel

### LINKED ISSUE #650: qa_volk_32f_x3_sum_of_poly_32f sometimes fails on ARM
Example of failing CI run: https://github.com/gnuradio/volk/actions/runs/6498731447/job/17650685199

Errors exceed the allowed tolerance. I'm not sure whether the tolerance is too low, or the kernels aren't up to snuff.

I was able to reproduce on a Raspberry Pi by running the test in a loop:

```
while ctest -R qa_volk_32f_x3_sum_of_poly_32f --output-on-failure; do :; done
```

The failures occur often for `a_neon` and somewhat less frequently for `neonvert`.

argilo:
Fixes #650.

The "sum of poly" kernel output is the sum of many positive & negative values, and the final sum can end up being close to zero. This causes test failures, because the relative error becomes large when the sum is near zero. Here I've switched to an absolute tolerance. I tested the generic kernel against a double-precision implementation and found that the error can be as high as 120 (which isn't surprising, since 131071 * 5 = 655355 polynomial terms are summed), so I set the tolerance at 1000 to keep it a safe distance away.


jdemel (approved):
LGTM.

This issue is one of the reasons why we should think of a better test environment. However, implementing more manual tests is tedious work.

In any case, we should try to strike a balance between more automated tests and more tailoring where necessary. 

My comments are obviously out of scope of this PR.

marcusmueller (approved):
yeah. This kernel is not the most precise anyway, I don't think anyone who needs precision would use it.

Issue #646, opened 2023-10-13
qa_volk_32f_s32f_convertpuppet_8u is flaky


argilo:
CI often fails, and the culprit seems to be `qa_volk_32f_s32f_convertpuppet_8u`. This was added in #617.

/cc @daniestevez 

The failure can be demonstrated locally:
```
$ while ctest -R convertpuppet --output-on-failure; do :; done
Test project /home/argilo/git/volk/build
    Start 49: qa_volk_32f_s32f_convertpuppet_8u
1/1 Test #49: qa_volk_32f_s32f_convertpuppet_8u ...   Passed    0.04 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.05 sec
Test project /home/argilo/git/volk/build
    Start 49: qa_volk_32f_s32f_convertpuppet_8u
1/1 Test #49: qa_volk_32f_s32f_convertpuppet_8u ...   Passed    0.03 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.03 sec
Test project /home/argilo/git/volk/build
    Start 49: qa_volk_32f_s32f_convertpuppet_8u
1/1 Test #49: qa_volk_32f_s32f_convertpuppet_8u ...   Passed    0.03 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.04 sec
Test project /home/argilo/git/volk/build
    Start 49: qa_volk_32f_s32f_convertpuppet_8u
1/1 Test #49: qa_volk_32f_s32f_convertpuppet_8u ...   Passed    0.03 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.03 sec
Test project /home/argilo/git/volk/build
    Start 49: qa_volk_32f_s32f_convertpuppet_8u
1/1 Test #49: qa_volk_32f_s32f_convertpuppet_8u ...   Passed    0.03 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.03 sec
Test project /home/argilo/git/volk/build
    Start 49: qa_volk_32f_s32f_convertpuppet_8u
1/1 Test #49: qa_volk_32f_s32f_convertpuppet_8u ...   Passed    0.03 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.03 sec
Test project /home/argilo/git/volk/build
    Start 49: qa_volk_32f_s32f_convertpuppet_8u
[...]

PR #655, opened 2023-10-14
Fix undefined behaviour in dot product kernels

### LINKED ISSUE #651: qa_volk_32fc_32f_dot_prod_32fc sometimes fails on "static on ubuntu-latest"
Example CI failures:

* https://github.com/gnuradio/volk/actions/runs/6518230858/job/17703441242
* https://github.com/gnuradio/volk/actions/runs/6498731447/job/17650683791
* https://github.com/gnuradio/volk/actions/runs/6497341447/job/17646235509
* https://github.com/gnuradio/volk/actions/runs/6172971032/job/16756885920

The failures are always in the `a_sse` and `u_sse` kernels, and only seem to occur in the "static on ubuntu-latest" run.

It can be seen that the output value is completely wrong.

```
2023-09-13T14:21:54.1649140Z 76: offset 0 in1: -1.30911 + -115.38j  in2: -3.19421 + -116.214j tolerance was: 0.01
2023-09-13T14:21:54.1659998Z 76: volk_32fc_32f_dot_prod_32fc: fail on arch a_sse
2023-09-13T14:21:54.1724783Z 76: offset 0 in1: -1.30911 + -115.38j  in2: -3.19421 + -116.214j tolerance was: 0.01
2023-09-13T14:21:54.1735504Z 76: volk_32fc_32f_dot_prod_32fc: fail on arch u_sse
```

### LINKED ISSUE #659: Undefined behaviour in volk_16i_32fc_dot_prod_32fc
As noted in https://github.com/gnuradio/volk/pull/655#issuecomment-1763224488, the `volk_16i_32fc_dot_prod_32fc` kernels have undefined behaviour due to strict pointer aliasing violations.

argilo:
Fixes #651.
Fixes #659.

In tests, the dot product is a summation of 131071 terms, ~~so a fair amount of error can accumulate. From the past CI failures it can be seen that the error can reach a magnitude of 2, so I've set a tolerance of 10 to leave a safe margin.~~

Since the summation can be close to zero, it makes sense to check the absolute error instead of the relative error. The `ccompare` function did not have an absolute mode, so I added one.

**Update**: The test failures and large errors identified in #651 were due to undefined behaviour in the dot product kernels, which sometimes manifested in trailing samples/taps not being included in the calculation. I've updated the pull request to include fixes for that problem.


argilo:
The test failure is unrelated; it's due to #646.

argilo:
On my machine I can get away with an absolute tolerance of `1e-2` even if I change the order of floating point operations in the generic kernel by reversing the loop. So I'm confused why the errors have been so much larger in CI. There might be something more to the failures. Let's hold off on merging this until we know what's going on.

jdemel:
@argilo you could convert this PR into a draft. This would clearly indicate that it is still being worked on.

argilo:
Aha! The `volk_32fc_32f_dot_prod_32fc` kernels misbehave because they cast a `float*` to `lv_32fc_t*` and then dereference it, which is undefined behaviour.

https://github.com/gnuradio/volk/blob/a26a1b808ddd75baeebe660e324f23c36dec6416/kernels/volk/volk_32fc_32f_dot_prod_32fc.h#L64-L65

https://github.com/gnuradio/volk/blob/a26a1b808ddd75baeebe660e324f23c36dec6416/kernels/volk/volk_32fc_32f_dot_prod_32fc.h#L78

I got rid of the float array and replaced it with `lv_32fc_t`, following the model that the accumulator kernels use.

Tests now pass on all platforms, even with the tighter `1e-2` absolute tolerance.

argilo:
The test failures are unrelated; both are due to #646.

argilo:
`volk_16i_32fc_dot_prod_32fc` has the same bug, but I'll leave that one for later.

argilo:
GNU Radio uses both `volk_32fc_32f_dot_prod_32fc` and `volk_16i_32fc_dot_prod_32fc` for FIR filtering, so it will be good to get them cleaned up and tested with tighter tolerances.

argilo:
> `volk_16i_32fc_dot_prod_32fc` has the same bug, but I'll leave that one for later.

Actually, I'll throw it in here because it will also need the new absolute tolerance code.

lib/kernel_tests.h
```diff
@@ -71,7 +71,7 @@ std::vector<volk_test_case_t> init_test_list(volk_test_params_t test_params)
     QA(VOLK_INIT_TEST(volk_16ic_x2_dot_prod_16ic, test_params))
     QA(VOLK_INIT_TEST(volk_16i_s32f_convert_32f, test_params))
     QA(VOLK_INIT_TEST(volk_16i_convert_8i, test_params))
-    QA(VOLK_INIT_TEST(volk_16i_32fc_dot_prod_32fc, test_params_inacc))
+    QA(VOLK_INIT_TEST(volk_16i_32fc_dot_prod_32fc, test_params.make_absolute(1e-1)))
```
argilo, line 74: The `16i` input is in the range -7 .. +7 (compared to -1 .. +1 for the 32f input in the other kernel) so the tolerance needs to be larger here.

argilo:
The test failure in the latest run is due to #650. At least there's some variety. :smile:

argilo:
Some information about why casting from `float[]` to `std::complex` is Undefined Behaviour:

https://stackoverflow.com/questions/23198943/is-it-legal-to-cast-float-to-stdcomplexfloat

kernels/volk/volk_16i_32fc_dot_prod_32fc.h
```diff
@@ -147,8 +147,7 @@ static inline void volk_16i_32fc_dot_prod_32fc_u_sse(lv_32fc_t* result,
     unsigned int number = 0;
     const unsigned int sixteenthPoints = num_points / 8;
 
-    float res[2];
-    float *realpt = &res[0], *imagpt = &res[1];
+    lv_32fc_t returnValue;
```
jdemel: Compilers would potentially complain about uninitialized variables here.
I suggest: `lv_32fc_t returnValue = lv_cmake(0.0f, 0.0f);`
I'd update the other occurrences as well.

lib/kernel_tests.h
jdemel: Does it fail due to inaccuracy sometimes?

jdemel (changes requested):
Almost LGTM. Only a very few comments. Thanks!

lib/kernel_tests.h
argilo: Yes.

kernels/volk/volk_16i_32fc_dot_prod_32fc.h
argilo: I'm not sure compilers would be confused by this particular case, but it would make the code easier to understand for humans. I'll add initializations and change all the updates to `+=`.

argilo:
Updated, and rebased on main. It might be a while until CI catches up. :smile:

argilo:
CI has now caught up. ✅

jdemel (approved):
LGTM

PR #638, opened 2023-10-12
Remove unused ORC code

argilo:
Several .orc files are unused, and a couple ORC kernels have been disabled for a long time. I've removed all the unused code here.

PR #658, opened 2023-10-14
Remove ORC kernels that use sqrtf

### LINKED ISSUE #622: ORC implementation of volk_32fc_magnitude_32f returns incorrect results on ARM

argilo:
Fixes #622
(as well as https://github.com/gqrx-sdr/gqrx/issues/570 and https://github.com/gqrx-sdr/gqrx/issues/1191).

The `sqrtf` ORC instruction is dangerous, because on at least one platform (ARM) it produces sqrt(0) = NaN.

@jdemel suggested in #640 that we remove affected kernels completely, so I've done that here. This may potentially reduce performance on platforms other than x86 & ARM, but it's probably worth it to not have incorrect output on ARM.

There are a couple other unused kernels that also have the `sqrtf` instruction, but they will be removed in #638.


argilo:
The test failures are unrelated; they're due to #646 and #651.

marcusmueller (approved):

jdemel (approved):
LGTM

PR #660, opened 2023-10-18
Fix undefined behaviour in volk_8u_x4_conv_k7_r2_8u

argilo:
Undefined Behaviour Sanitizer (enabled with `-fsanitize=undefined`) reports the following Undefined Behaviour during a `volk_profile` run:

`/home/argilo/git/volk/kernels/volk/volk_8u_x4_conv_k7_r2_8u.h:113:38: runtime error: left shift of 2 by 30 places cannot be represented in type 'int'`

This left shift results in signed integer overflow, which is Undefined Behaviour. To fix it, I changed `decision0` and `decision1` to `unsigned int`, which allows them to accommodate 32-bit positive integers without overflow. These variables never store negative numbers, so the change shouldn't break anything.

### DIFF
--- a/kernels/volk/volk_8u_x4_conv_k7_r2_8u.h
+++ b/kernels/volk/volk_8u_x4_conv_k7_r2_8u.h
@@ -82,7 +82,8 @@ static inline void BFLY(int i,
                         decision_t* d,
                         unsigned char* Branchtab)
 {
-    int j, decision0, decision1;
+    int j;
+    unsigned int decision0, decision1;
     unsigned char metric, m0, m1, m2, m3;
 
     int NUMSTATES = 64;


argilo:
The test failures are unrelated; they're from #646 and #651.

marcusmueller (approved):

daniestevez:
Just seen this PR. Hold on a second. This doesn't look right. `decision0` and `decision1` are supposed to hold one bit (either `0` or `1`), as per
```c
    decision0 = (signed int)(m0 - m1) > 0;
    decision1 = (signed int)(m2 - m3) > 0;
```
so the problem must be elsewhere. I'm going to look at this in more detail and let you know.

**Edit:** Okay. I now realize the fix is correct. The problem was (as you said; should have read more attentively) that when the two-bit value `(decision0 | decision1 << 1)` is packed in the MSB position of an `unsigned int`, this shifts data into the sign bit, which is UB (the usual arithmetic shift vs logical shift problematic). I agree that `decision0` and `decision1` should be `unsigned int`, not only because it fixes this UB, but because doing bitwise operations on signed values is usually discouraged, and also because `d->w`, which is where this values get packed into, is also `unsigned int`.

So, TLDR, this PR also looks good to me.

argilo:
That's correct. When `i` is 15, `decision1` is shifted 1 + 2*15 = 31 bits to the left, which pushes it into the sign bit.

Thanks for having a look over the change.

jdemel (approved):
LGTM

Also, thanks to everyone for the discussion and review on this one.

PR #664, opened 2023-10-19
Fix undefined behaviour in volk_32u_reverse_32u

argilo:
Undefined Behaviour Sanitizer (UBSAN, `-fsanitize=undefined`) reports two instances of overflowing left shifts in volk_32u_reverse_32u.

In both cases, an `unsigned char` is promoted to `int` and then left-shifted by 24 bits, but `int` is not large enough to hold the result. I've added an explicit cast to `uint32_t`, which resolves the Undefined Behaviour warnings.


marcusmueller (approved):

jdemel (approved):
LGTM

PR #665, opened 2023-10-20
convert 32f->32i: fix compiler warnings about loss of int precision

marcusmueller:
This feels like someone else probably has an open PR about it already, but I can't find it.

Compilers warn that converting INT_MAX to float leads to INT_MAX+1 as float value.

That's absolutely not surprising: bits(single precision float mantissa) < bits(int), so an int of value 2^N - 1 must inherently be non-representable for N close to bits(int), and MAX_INT = 2³²-1.

So, the upround that I'm doing here explicitly is always done, on every machine I could think of, and so what this compiler warning silencing does is always done, anyways.

Finally makes VOLK a zero-warning compile on my clang16 x68_64 avx2 machine.


argilo (approved):
Looks good, and this shouldn't change the behaviour.

jdemel (approved):
LGTM

PR #668, opened 2023-10-22
Fix register size warnings in reverse kernel

argilo:
When building VOLK on 64-bit ARM, many compiler warnings are printed because the register size of the `rbit` instruction is incorrect. To fix that, I've added an aarch64 version of the `DO_RBIT` macro which includes the `w` template modifier to select 32-bit registers.

Also, it appears the `arm` version of the reverse kernel was inaccessible (on both 32-bit and 64-bit ARM) because `volk_kernel_defs.py` does not understand nested `#ifdef LV_HAVE_*`. I separated the `#ifdef`s and now the `arm` version of the kernel runs on both 32-bit and 64-bit ARM.

### DIFF
--- a/kernels/volk/volk_32u_reverse_32u.h
+++ b/kernels/volk/volk_32u_reverse_32u.h
@@ -363,10 +363,20 @@ volk_32u_reverse_32u_neonv8(uint32_t* out, const uint32_t* in, unsigned int num_
     }
 }
 
-#else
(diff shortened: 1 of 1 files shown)


argilo:
The test failure is unrelated; it's due to #669.

jdemel (approved):
LGTM

Issue #666, opened 2023-10-20
volk_malloc should not printf


marcusmueller:
`volk_malloc` is a service routine; not being able to allocate memory is unfortunate, but just like normal `libc` malloc, the consumer needs to check whether an allocation worked, either way. The "printf debugging" shouldn't be "normally on", as a library shouldn't do freaky things to a user's stderr on its own – *especially* not using `fprintf` in an environment where output buffering might already be handled by the C++ runtime to begin with.

jdemel:
I assume you refer to:
https://github.com/gnuradio/volk/blob/fe463375f72170a56254bfc5bd2bfc6872828dd9/lib/volk_malloc.c#L41
and other occurrences of `fprintf` in that file and function.
As far as I remember, it's always been there.

I could imagine that simply removing the `fprintf` lines would be reasonable.

PR #672, opened 2023-10-22
Add length checks to puppets

argilo:
Almost all kernels are now testable all the way down to vector length zero, with a couple exceptions which I've fixed here:

* `volk_32f_8u_polarbutterflypuppet_32f` fails below length 2 due to division by zero in `maximum_frame_size`
* `volk_8u_x3_encodepolarpuppet_8u` generates an error message below length 1 (`VOLK: Error allocating memory: either size or alignment is 0`)

I've added length checks to both, which should open the door to testing kernels over a wide range of vector lengths.

I also too the opportunity to add a missing carriage return to the allocation error message; it's the only such message that lacks a carriage return.


argilo:
Related: #666

marcusmueller (approved):

jdemel (approved):
LGTM

PR #675, opened 2023-10-22
Include ORC in neonv8 machine definition

argilo:
I noticed that neonv8 kernels do not run in CI, nor do they run on my Raspberry Pi if `liborc-0.4-dev` is installed. It looks like this happens because VOLK chooses the `neon_orc` machine over the ORC-less `neonv8` machine.

The `neonv8` machine is the only one in `machines.xml` that lacks `orc|`; adding this corrects the problem, and VOLK then chooses the `neonv8_orc` machine and executes neonv8 kernels.

### DIFF
--- a/gen/machines.xml
+++ b/gen/machines.xml
@@ -13,7 +13,7 @@
 </machine>
 
 <machine name="neonv8">
-<archs>generic neon neonv8</archs>
+<archs>generic neon neonv8 orc|</archs>
 </machine>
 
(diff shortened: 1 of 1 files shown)


argilo:
CI logs show that neonv8 kernels are now executing on aarch64.

The test failure is unrelated; it's due to #676.

jdemel (approved):
LGTM

PR #196, opened 2018-09-15
ARMV8 kernels added

hcab14:
Recently I got a raspberry pi3b+ and installed a 64bit operation system. Thanks to @balister's work (https://github.com/gnuradio/volk/commit/e98e9277409cd658f1a5b65ffaa0caab842bebce) support for ARMV8 kernels is available:
- `volk_32u_reverse_32u` and `volk_64u_byteswappuppet_64u` failed on this system and the commits in this pull request fix this problem, along with 
- fixes for some compiler warnings for other NEON kernels

```
pi@raspberrypi:~/volk/build$ uname -a
Linux raspberrypi 4.14.34-v8+ #1 SMP PREEMPT Tue Apr 17 02:45:42 PDT 2018 aarch64 GNU/Linux
```

```
pi@raspberrypi:~/volk/build$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/aarch64-linux-gnu/6/lto-wrapper
Target: aarch64-linux-gnu
[...]

PR #677, opened 2023-10-23
Generate random integers with uniform_int_distribution

### LINKED ISSUE #688: Poor random numbers in uint tests.
https://github.com/gnuradio/volk/blob/main/lib/qa_utils.cc#L54  Contains: `float int_max = float(uint64_t(2) << (type.size * 8));`.  If `type.size` is 8, then `int_max = uint64_t(2) << 64`.  Shift a 64-bit number by 64 is undefined behavior.  It appears that gcc masks the shift amount and no shift is performed.  Thus the random numbers generated are between 0 and 2 for unsigned 64-bit numbers and -1 and 1 for signed 64-bit numbers.

This is not ideal and causes some tests to not actually exercise algorithm.  See   https://github.com/gnuradio/volk/issues/605

argilo:
Fixes #688.

Random integer generation in `load_random_data` has a lot of bugs:

* `float(uint64_t(2) << (type.size * 8))` performs a left shift by 64 bits when the type size is 8, and this results in Undefined Behaviour. On my system, the only generated `uint64_t` values are 0, 1, and 18446744073709551615.
* The `float(uint64_t(2) << (type.size * 8))` calculation is also incorrect (twice as large as it should be) when the type size is smaller than 8, resulting in Undefined Behaviour when floating-point random values are cast to integers.
* Floating-point values have fewer bits of precision than 32-bit integers, so most generated `int32_t` and `uint32_t` values are multiples of 256.
* The only generated `uint16_t` values are 0, 1, 2, 3, 4, 5, 6, 7, 65529, 65530, 65531, 65532, 65533, 65534, and 65535.

Here I've fixed all these problems by replacing `std::uniform_real_distribution` with `std::uniform_int_distribution`.

I had to keep `int16_t` generation in the range -7 .. +7, because various kernels added in https://github.com/gnuradio/volk/pull/77 fail otherwise (due to differing integer overflow between implementations).

In a separate commit, I also fixed the compiler warnings in qa_utils.cc:

* unused parameter warnings for `icompare`, which I accidentally introduced in #655
* ignored type qualifier on a cast

### DIFF
--- a/lib/qa_utils.cc
+++ b/lib/qa_utils.cc
@@ -485,7 +485,7 @@ bool ccompare(t* in1, t* in2, unsigned int vlen, float tol, bool absolute_mode)
 }
 
 template <class t>
-bool icompare(t* in1, t* in2, unsigned int vlen, unsigned int tol, bool absolute_mode)
+bool icompare(t* in1, t* in2, unsigned int vlen, unsigned int tol)
 {
     bool fail = false;
     int print_max_errs = 10;
@@ -573,7 +573,7 @@ bool run_volk_tests(volk_func_desc_t desc,
     vlen = vlen + vlen_twiddle;
 
     const float tol_f = tol;
-    const unsigned int tol_i = static_cast<const unsigned int>(tol);
+    const unsigned int tol_i = static_cast<unsigned int>(tol);
 
     // first let's get a list of available architectures for the test
     std::vector<std::string> arch_list = get_arch_list(desc);
@@ -799,14 +799,12 @@ bool run_volk_tests(volk_func_desc_t desc,
                             fail = icompare((int64_t*)test_data[generic_offset][j],
                                             (int64_t*)test_data[i][j],
                                             vlen * (both_sigs[j].is_complex ? 2 : 1),
-                                            tol_i,
-                                            absolute_mode);
+                                            tol_i);
                         } else {
                             fail = icompare((uint64_t*)test_data[generic_offset][j],
                                             (uint64_t*)test_data[i][j],
                                             vlen * (both_sigs[j].is_complex ? 2 : 1),
-                                            tol_i,
-                                            absolute_mode);
+                                            tol_i);
                         }
                         break;
                     case 4:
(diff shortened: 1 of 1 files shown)


argilo:
This has revealed that `volk_64u_byteswap_neonv8` is broken. I guess we'll have to fix that first.

lib/qa_utils.cc
```diff
@@ -51,44 +51,69 @@ void load_random_data(void* data, volk_type_t type, unsigned int n)
...
+                    std::numeric_limits<uint8_t>::min(),
+                    std::numeric_limits<uint8_t>::max());
+                for (unsigned int i = 0; i < n; i++)
+                    ((uint8_t*)data)[i] = uniform_dist(rnd_engine);
+            }
+            break;
+        default:
+            throw "load_random_data: no support for data size > 8 or < 1"; // no
+                                                                           // shenanigans
+                                                                           // here
```
marcusmueller, line 116: :+1:

marcusmueller (approved):

marcusmueller:
> This has revealed that `volk_64u_byteswap_neonv8` is broken. I guess we'll have to fix that first.

I'll argue that falling back to the generic implementation is OK here and we should not prioritize fixing it (instead of removing it) over fixing *all our test cases*

marcusmueller:
From a [build](https://github.com/gnuradio/volk/actions/runs/6616346149/job/17970422866?pr=679) on aarch64 g++12 Ubuntu22.04:
```
RUN_VOLK_TESTS: volk_64u_byteswappuppet_64u(131071,1)
generic_decompose completed in 0.947894 ms
generic completed in 0.596396 ms
neonv8 completed in 4.90347 ms
Best aligned arch: generic
Best unaligned arch: generic
Test #125: qa_volk_64u_byteswappuppet_64u .......................   Passed    0.17 sec
```

(generic_decompose is the current generic, generic is my new [proposed](https://github.com/gnuradio/volk/pull/679) one.)

Same picture on the older g++8 of Ubuntu 20.04. 

In other words, yeah, let's kill that broken neonv8 kernel. It serves no purpose that generic (even the old one) doesn't do better, in all our current builds.

argilo:
It looks like the neonv8 kernel was even slower back when it was added: https://github.com/gnuradio/volk/pull/196

argilo:
I'll rebase this after #680 is merged.

argilo:
I rebased on `main` (to pick up #680) and the tests now pass.

Issue #674, opened 2023-10-22
Undefined Behaviour in volk_8u_x4_conv_k7_r2_8u


argilo:
UBSAN reports undefined behaviour in this code:

https://github.com/gnuradio/volk/blob/42f57cd67506e7fb6a7795af3948b803de0085f4/kernels/volk/volk_8u_x4_conv_k7_r2_8u.h#L112-L113

Here `w` is an `unsigned int[2]`, and the `s * (sizeof(decision_t) / sizeof(unsigned int))` calculation intentionally reaches outside the bounds of the array to access the `s`th element of the outer `decision_t` array `d`.

It looks like this hack became necessary because `sizeof(decision_t)` is 8, and yet it is marked as being 16-byte aligned; because of this, accessing `d[1]` (or any other odd offset) fails due to misalignment. I suspect that the 16-byte alignment of `decision_t` is incorrect or unnecessary.

jdemel:
Thanks for finding this issue. Bugs in the FEC decoder code are a nightmare because or test infrastructure is unsuitable for this kernel.

Not directly related to the bug at hand, but:
I had the idea to introduce [gtest](https://github.com/google/googletest) to VOLK. Writing more extensive tests for this kernel with specific input might be worthwhile.

argilo:
It looks like it will not be possible to avoid Undefined Behaviour in this kernel without changing the API. Decisions are returned as an array of `unsigned char` and yet they're actually packed as `unsigned int`. VOLK and GNU Radio both use a `decision_t` union to switch between the two in an unsafe way.

The decisions are packed bits. I think storing them in `unsigned char` would make more sense. Even though the function signature would remain the same, this would be an API change because the bytes would be in order, while currently each group of four bytes is reversed (on little-endian platforms, anyway). So we'd have to deprecate this kernel and create a new one.

PR #678, opened 2023-10-23
Add UBSAN to ASAN builds

argilo:
Adding the Undefined Behaviour Sanitizer (UBSAN) to ASAN builds will allow more bugs to be uncovered. I already found a number of bugs this way:

* #660
* #664
* #674 
* #677

### DIFF
--- a/cmake/Modules/VolkBuildTypes.cmake
+++ b/cmake/Modules/VolkBuildTypes.cmake
@@ -187,9 +187,9 @@ endif(NOT WIN32)
 # NOTE: This is not defined on Windows systems.
 ########################################################################
 if(NOT WIN32)
-  SET(CMAKE_CXX_FLAGS_ASAN "-Wall -Wextra -g -O2 -fsanitize=address -fno-omit-frame-pointer" CACHE STRING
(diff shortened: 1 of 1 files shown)


marcusmueller (approved):

jdemel:
This changes runtime behavior. A quick search indicated that this change may affect performance. Do we want this to happen? Maybe it is sufficient to add a CI test with this option enabled.

argilo:
ASAN changes runtime behaviour and reduces performance in the same way; it's only meant to be used for finding bugs. The intent here is to add more checks so that more bugs can be caught. I think that's compatible with the intent of #192, which added an ASAN build type in the first place, and seems to have been aimed at finding bugs by hand.

Running CI tests with ASAN+UBSAN at least once for each platform we care about (x86, ARM) would definitely help catch more bugs. The `UBSAN_OPTIONS="halt_on_error=1"` environment variable would need to be set, since UBSAN's default behaviour is to print an error message and carry on.

jdemel:
Got it.

CMake provides `-DCMAKE_BUILD_TYPE=ASAN`. We just add the options here. The way this cmake file works is not obvious.

I merge this PR now. We can add the CI tests in another PR.

PR #679, opened 2023-10-23
granular parallel generic kernel for 64u_byteswap

marcusmueller:
This simplifies (at least to me) understanding what the generic kernel does, and it's also about 1.5 times faster on my machines.

### DIFF
--- a/kernels/volk/volk_64u_byteswap.h
+++ b/kernels/volk/volk_64u_byteswap.h
@@ -56,6 +56,30 @@
 #include <inttypes.h>
 #include <stdio.h>
 
+#ifdef LV_HAVE_GENERIC
+/* Adapted from https://graphics.stanford.edu/~seander/bithacks.html#ReverseParallel
+ * Where they reverse the bits in an N-bit word. But who's stoppng me from doing the same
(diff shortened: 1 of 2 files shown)


marcusmueller:
is a good addition to make the byteswap somewhat performant on non-x86 platforms, especially in light of #680

jdemel:
Does this PR touch the intend of #606 ? I know, the concern different implementations.

marcusmueller:
No, it's unrelated. However, #680 addressed that quite directly.

jdemel (commented):
This PR LGTM.

However, this whole kernel has issues. The include guards are at least confusing. The `_a` include guard is around the `_u` kernels and vice versa. Tail handling creates copypasta code. Loop variables are defined outside of loops. All in all, this kernel needs even more clean up. This is beyond this PR though.

One concern: We just removed all the `_a_generic` kernels. The diff looks like you rename one of these kernels.

Could you rebase your PR first before we merge it?

jdemel (changes requested):
Could you rebase this PR onto the current `main`? I'd say we can go with it then.

PR #682, opened 2023-10-23
Coverage

marcusmueller:
This adds support for Coverage reporting.

Requirements:

- gcc or clang
- gcov
- gcovr

Usage:

```shell
mkdir build-coverage && cd build-coverage
cmake -DCMAKE_BUILD_TYPE=Coverage ..
make -j
make coverage
firefox coverage/index.html
```

Demo of the result on my oldish desktop machine: https://doppler.inpha.se/coverage/


marcusmueller:
The patch to CodeCoverage.cmake is something I'm trying to upstream: bilke/cmake-modules#82

jdemel (commented):
LGTM. It's unfortunate that we can't include thie `CodeCoverage.cmake` but need to copy it.

Could you add a reference to the origin of the `CodeCoverage.cmake` file? This might make updates easier.

marcusmueller:
@jdemel done, but you won't be updating this anytime soon.

jdemel:
> @jdemel done, but you won't be updating this anytime soon.

That is actually a good reason to put the info in there.

CMakeLists.txt
```diff
@@ -202,6 +202,22 @@ endif(ENABLE_ORC)
 ########################################################################
 add_subdirectory(docs)
 
+########################################################################
+# Setup Coverage
+########################################################################
+if(${CMAKE_BUILD_TYPE} STREQUAL "Coverage")
```
noc0lour, line 238: With the CodeCoverage.cmake setting the Build_Type to anything other than Debug results in a warning. Either removing that check from there or leaving the `CMAKE_BUILD_TYPE` alone and just adding a switch to enable coverage would be an option.

CMakeLists.txt
```diff
@@ -202,6 +202,22 @@ endif(ENABLE_ORC)
...
+    # Use CodeCoverage based on Lars Bilke's
+    # https://github.com/bilke/cmake-modules , with extensive
+    # modifications as done in the past for GNU Radio.
+    include(CodeCoverage)
+    append_coverage_compiler_flags()
+    setup_target_for_coverage_gcovr_html(
```
noc0lour, line 244: Why only html output and why gcovr? IIRC lcov is the more mature and complete tool. Allowing regular coverage output we can hook that into codecov.io pipelines and add information to pullrequests and pushes to track coverage as we go.

noc0lour (changes requested):
The direct HTML output is nice, but actually the regular coverage info can be post-processed with tools to give you HTML as you wish. Therefore it would be better if we would produce regular coverage output and not HTML.

Would it be possible for you to add a github workflow which will run and produce codecov artifacts and uploads them with codecov-action so we can hook it into codecov.io?

CMakeLists.txt
marcusmueller: gcov vs lcov:

TBH, because I looked into lcov first, couldn't make it work, and then resorted to doing what was possible within limited time :)

html:

because I wanted this to be immediately useful for insights, and hence jumped on a minimum effort way of getting readable output. Proper CI integration, including things like complaining when you add code that is specifically not well-covered, is something that needs to come – but having this seemed better than not having it.

marcusmueller: gcov vs lcov:

TBH, because I looked into lcov first, couldn't make it work, and then resorted to doing what was possible within limited time :)

html:

because I wanted this to be immediately useful for insights, and hence jumped on a minimum effort way of getting readable output. Proper CI integration, including things like complaining when you add code that is specifically not well-covered, is something that needs to come – but having this seemed better than not having it.

CMakeLists.txt
marcusmueller: this is honestly just a mirror of what GR does; looking back, yes, having extra flags instead of a build type of its own would have been a worthwhile option, but I foreseeable won't have time to implement that.

noc0lour:
@marcusmueller If you don't mind I'll have a crack at it to make it useful for user output & ingestion into statistics/CI pipeline. Because with the way it is right now it will not be immediately useful to integrate...

marcusmueller:
Rebased the PR. @jdemel ?

PR #684, opened 2023-10-24
Add back volk_32f_exp_32f test

argilo:
I'm not sure why, but the `volk_32f_exp_32f` test was removed in #350, so the kernel is now untested. Here I've added back the test, which appears to run fine on my system.


marcusmueller (approved):

jdemel (approved):
LGTM

PR #689, opened 2023-10-28
Fix random numbers

### LINKED ISSUE #688: Poor random numbers in uint tests.

jsallay:
Resolve https://github.com/gnuradio/volk/issues/688

Note that we need to merge https://github.com/gnuradio/volk/pull/606 beforehand because the armv8 unit test fails if we fix the random numbers.

### DIFF
--- a/lib/qa_utils.cc
+++ b/lib/qa_utils.cc
@@ -52,7 +52,7 @@ void load_random_data(void* data, volk_type_t type, unsigned int n)
         }
     } else {
         float int_max, int_min;
-        switch(type.size) {
+        switch (type.size) {
         case 8:
             if (type.is_signed) {
                 int_max = static_cast<float>(std::numeric_limits<int64_t>::max());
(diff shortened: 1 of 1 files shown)


argilo:
I already have a pull request open to fix this (and some other bugs in random integer generation): #677

jdemel (approved):
LGTM

jdemel:
It looks like we need to decide whether we want to merge this PR or #677 . At the moment I would prefer #677 because it addresses a bit more.

@jsallay I hope you don't mind if we close this PR and merge @argilo's PR.

argilo:
#677 does include all the improvements made here, but one small difference is that it keeps the `int16_t` range at -7 .. +7 (for now) since some tests would become flaky if it was increased to the full numeric range of `int16_t`.

jsallay:
I don't have a problem with closing this.

jdemel:
Alright, I'm closing this PR in favor of #677 now.

PR #692, opened 2023-11-04
Remove broken volk_16i_max_star_16i_neon protokernel

argilo:
A compiler warning points out that the `vst1q_s16` call on line 142 results in a stack buffer overrun because `candidate` is not long enough to hold the result, which is an array of shorts. I verified with ASAN that a buffer overrun does in fact occur.

Since the `volk_16i_max_star_16i` kernel is deprecated, we may as well just remove this broken protokernel.


jdemel (approved):
LGTM.

jdemel (approved):
LGTM

PR #693, opened 2023-11-04
Fix compiler warnings

argilo:
When building with `-DCMAKE_BUILD_TYPE=ASAN` a number of compiler warnings are shown. I've cleaned up most of them here.

### DIFF
--- a/apps/volk_option_helpers.cc
+++ b/apps/volk_option_helpers.cc
@@ -188,6 +188,7 @@ void option_list::parse(int argc, char** argv)
                     } catch (std::exception& exc) {
                         throw std::exception();
                     };
+                    break;
                 case STRING:
                     std::cout << this_option->printval << std::endl;
                     break;
@@ -225,15 +226,8 @@ void option_list::help()
             help_line += this_option->shortform + " [ " + this_option->longform + " ]";
         }
 
-        switch (help_line.size() / 8) {
-        case 0:
-            help_line += "\t";
-        case 1:
-            help_line += "\t";
-        case 2:
-            help_line += "\t";
-        case 3:
-            help_line += "\t";
+        while (help_line.size() < 32) {
+            help_line += " ";
         }
         help_line += this_option->msg;
         std::cout << help_line << std::endl;
--- a/kernels/volk/volk_32f_8u_polarbutterflypuppet_32f.h
+++ b/kernels/volk/volk_32f_8u_polarbutterflypuppet_32f.h
@@ -81,6 +81,8 @@ static inline void volk_32f_8u_polarbutterflypuppet_32f_generic(float* llrs,
                                                                 unsigned char* u,
                                                                 const int elements)
 {
+    (void)input; // suppress unused parameter warning
+
     if (elements < 2) {
         return;
     }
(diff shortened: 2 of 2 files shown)


apps/volk_option_helpers.cc
```diff
@@ -188,6 +188,7 @@ void option_list::parse(int argc, char** argv)
                     } catch (std::exception& exc) {
                         throw std::exception();
                     };
+                    break;
```
argilo, line 191: The fall-through here was a bug; it caused a blank line to be printed if the `-R`, `-j`, or `-p` options were used with `volk_profile`.

apps/volk_option_helpers.cc
```diff
@@ -225,15 +226,8 @@ void option_list::help()
...
 
-        switch (help_line.size() / 8) {
-        case 0:
-            help_line += "\t";
-        case 1:
-            help_line += "\t";
-        case 2:
-            help_line += "\t";
-        case 3:
-            help_line += "\t";
```
argilo, line 236: This code is just meant to line up the tables that are displayed by `volk_profile --help` and `volk_config-info --help`. The new version gets rid of the fall-through warnings, and is simpler.

kernels/volk/volk_32f_8u_polarbutterflypuppet_32f.h
```diff
@@ -81,6 +81,8 @@ static inline void volk_32f_8u_polarbutterflypuppet_32f_generic(float* llrs,
                                                                 unsigned char* u,
                                                                 const int elements)
 {
+    (void)input;
+
```
argilo: It's intentional that `input` is unused, so I added `(void)input` to suppress the unused parameter warnings.

jdemel: C23 provides something for this exact situation:
[`maybe_unused`](https://en.cppreference.com/w/c/language/attributes/maybe_unused)
I suggest to add a comment to the `(void)input;` lines because this will make it more verbose. Otherwise, someone may remove these lines in the future because the look superfluous.

jdemel (changes requested):
I found one little thing that could improve the code for future reference.

kernels/volk/volk_32f_8u_polarbutterflypuppet_32f.h
marcusmueller: as to give a differing opinion: `(void)input;` is the canonical way of silencing this warning and thus of marking a variable as unused. It's as concise as it gets! (and pretty unambiguous; if there's a `(void)input;` in the code, and you are about to remove it, you would reasonably ask yourself *why it is in the code*, and [stackoverflow](https://stackoverflow.com/search?q=%22%28void%29%22++%5BC%5D) has you [covered](https://stackoverflow.com/questions/4647665/why-cast-an-unused-function-parameter-value-to-void).

marcusmueller: (also, the goal here would be to enable `-Werror -Wunused-parameter` so that a removal of that line would become a build fail)

marcusmueller (approved):

jdemel (approved):
LGTM.

Issue #663, opened 2023-10-19
aarch64: volk_32fc_s32f_power_spectrum_32f fail


marcusmueller:
Found during a Fedora scratch build.
[build.log.txt](https://github.com/gnuradio/volk/files/13042347/build.log.txt)

Sum of poly is already covered in #650 , I'm surprised I didn't see that:

<s>suspiciously fast in neonvert, and also wrong:
```
 72/139 Test  #72: qa_volk_32f_x3_sum_of_poly_32f .......................***Failed    0.01 sec
RUN_VOLK_TESTS: volk_32f_x3_sum_of_poly_32f(131071,1)
generic completed in 0.216802 ms
a_neon completed in 0.328402 ms
neonvert completed in 0.0616 ms
offset 0 in1: -8937.34 in2: -9080.31 tolerance was: 0.01
volk_32f_x3_sum_of_poly_32f: fail on arch a_neon
Best aligned arch: neonvert
Best unaligned arch: neonvert
```
</s>
and power spectrum's tolerance seems ambitious, but I haven't looked at that

```
 95/139 Test  #95: qa_volk_32fc_s32f_power_spectrum_32f .................***Failed    0.01 sec
RUN_VOLK_TESTS: volk_32fc_s32f_power_spectrum_32f(131071,1)
generic completed in 0.472402 ms
neon completed in 0.286401 ms
offset 114951 in1: -48.291 in2: -48.291 tolerance was: 1e-06
volk_32fc_s32f_power_spectrum_32f: fail on arch neon
Best aligned arch: neon
Best unaligned arch: neon

**Comment by argilo** [comment]
This one fails in CI too:

* https://github.com/gnuradio/volk/actions/runs/6603595907/job/17936864893
* https://github.com/gnuradio/volk/actions/runs/6603892571/job/17937473189

```
  97: RUN_VOLK_TESTS: volk_32fc_s32f_power_spectrum_32f(131071,1)
  97: generic completed in 14.0574 ms
  97: neon completed in 11.6342 ms
  97: offset 108359 in1: -48.2963 in2: -48.2963 tolerance was: 1e-06
  97: volk_32fc_s32f_power_spectrum_32f: fail on arch neon
```

argilo:
As an experiment, I tried increasing the tolerance to `2e-06` and no failures occurred on my Raspberry Pi after a full day running the `power_spectrum` test.

argilo:
[...]

PR #694, opened 2023-11-06
volk_32fc_x2_divide_32fc: add documentation about numerical accuracy

### LINKED ISSUE #608: volk_32fc_x2_divide_32fc gives nan in aarch64 for 0.0 / 1e-20
Investigating a bug in gr-satellites (https://github.com/daniestevez/gr-satellites/issues/416), we have found that the `volk_32fc_x2_divide_32fc` kernel gives `nan + 1j nan` when calculating the division of `0.0 + 1j 0.0` by `1e-20 + 1j 0.0` in aarch64. This seems to happen both in the `generic` kernel and with the `neon` kernel (though replicating these tests on another aarch64 machine would be a good idea). The testing has been done in a Rasberry Pi 4 with the [volk 2.4.1 package in Debian Bullseye](https://packages.debian.org/bullseye/libvolk2.4).

I don't have an aarch64 machine at hand, but I've tested on an x86_64 and an armv7 machine and the result is `0.0 + 1j 0.0` as we would expect.

I'm using the following program to test:
```c
#include <volk/volk.h>
#include <complex.h>
#include <stdio.h>
#include <string.h>

int main(int argc, char **argv) {
    size_t alignment = volk_get_alignment();
    float complex *z = volk_malloc(512 * sizeof(float complex), alignment);
    float complex *w = volk_malloc(512 * sizeof(float complex), alignment);
    float complex *x = volk_malloc(512 * sizeof(float complex), alignment);
    memset(z, 0, 512 * sizeof(float complex));
    memset(w, 0, 512 * sizeof(float complex));
    z[0] = 0.0f;
    w[0] = 1e-20f;
    volk_32fc_x2_divide_32fc(x, z, w, 512);
    printf("%e + 1j %e\n", creal(x[0]), cimag(x[0]));
    return 0;
}
```
This is to be built as
```
gcc -O3 -Wall -o test_divide_volk test_divide_volk.c -lvolk
```

If we replace `1e-20` by `5.5e-20`, the test works in aarch64 (it gives zero rather than nan).

For gr-satellites this is not a huge problem, because the reason this popped up is because we have an AGC that does roughly `signal / (rms_amplitude(signal) + 1e-20))`. The `1e-20` in the denominator is there to prevent `0 / 0` when the input signal is zero. We can increase this small `1e-20` to whatever value is safe for all architectures.

However, I would like to chase this rabbit hole further and understand why this problem happens, whether something can be fixed or we have to live with this (for instance if setting up the kernel so that `0 / 1e-20` gives correct results would hurt performance), and what is a safe value for all architectures (apparently `5.5e-20` works on aarch64, but it still could be too small on other machines).

daniestevez:
This adds some documentation to the complex divide kernel about the potential pitfall in some architectures of obtaining inf or nan with a small denominator (even if the numerator is zero) due to reduced range because of -fcx-limited-range and -freciprocal-math.

This closes #608.

### DIFF
--- a/kernels/volk/volk_32fc_x2_divide_32fc.h
+++ b/kernels/volk/volk_32fc_x2_divide_32fc.h
@@ -54,6 +54,44 @@
  *   volk_free(input_vector);
  *   volk_free(out);
  * \endcode
+ *
+ * \b Numerical accuracy
+ *
+ * The division of complex numbers (a + bj)/(c + dj) is calculated as
+ * (a*c + b*d)/(c*c + d*d) + (b*c - a*d)/(c*c + d*d) j.
+ *
+ * Since Volk is built using unsafe math optimizations
+ * (-fcx-limited-range -funsafe-math-optimizations with gcc), after computing
+ * the two numerators and the denominator in the above formula, the compiler
+ * can either:
+ *
+ * a) compute the two divisions;
+ * b) compute the inverse of the denominator, 1.0/(c*c + d*d), and then
+ * multiply this number by each of the numerators.
+ *
+ * Under gcc, b) is allowed by -freciprocal-math, which is included in
+ * -funsafe-math-optimizations.
+ *
+ * Depending on the architecture and the estimated cost of multiply and divide
+ * instructions, the compiler can perform a) or b). gcc performs a) under x86_64
+ * and armv7, and b) under aarch64.
+ *
+ * To avoid obtaining inf or nan in some architectures, care should be taken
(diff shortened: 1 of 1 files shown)


argilo (approved):

daniestevez:
Force-pushed to fix clang format.

Force-pushed again. Hopefully now it's correct.

argilo:
The test failure is unrelated; it's due to #663.

jdemel (approved):
LGTM

Issue #649, opened 2023-10-13
qa_volk_32f_x2_pow_32f is flaky


argilo:
Example of a failed CI run: https://github.com/gnuradio/volk/actions/runs/6497341447/job/17646238510

Failures can be reproduced locally with:

```
ctest -R qa_volk_32f_x2_pow_32f --output-on-failure --repeat until-fail:100000
```

Log:

```
Test project /home/argilo/git/volk/build
    Start 71: qa_volk_32f_x2_pow_32f
1/1 Test #71: qa_volk_32f_x2_pow_32f ...........***Failed    0.06 sec
RUN_VOLK_TESTS: volk_32f_x2_pow_32f(131071,1)
a_avx2_fma completed in 1.77732 ms
a_avx2 completed in 2.54741 ms
a_sse4_1 completed in 4.50744 ms
generic completed in 1.8077 ms
u_sse4_1 completed in 4.52563 ms
u_avx2_fma completed in 1.75473 ms
u_avx2 completed in 2.27722 ms
offset 126757 in1: -1.81431 in2: 1.8141 tolerance was: 0.01
volk_32f_x2_pow_32f: fail on arch a_avx2_fma
offset 126757 in1: -1.81431 in2: 1.8141 tolerance was: 0.01
volk_32f_x2_pow_32f: fail on arch a_avx2
offset 126757 in1: -1.81431 in2: 1.8141 tolerance was: 0.01
volk_32f_x2_pow_32f: fail on arch a_sse4_1
offset 126757 in1: -1.81431 in2: 1.8141 tolerance was: 0.01
volk_32f_x2_pow_32f: fail on arch u_sse4_1
offset 126757 in1: -1.81431 in2: 1.8141 tolerance was: 0.01
volk_32f_x2_pow_32f: fail on arch u_avx2_fma
offset 126757 in1: -1.81431 in2: 1.8141 tolerance was: 0.01
volk_32f_x2_pow_32f: fail on arch u_avx2
Best aligned arch: u_avx2_fma
Best unaligned arch: u_avx2_fma


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.07 sec

The following tests FAILED:
	 71 - qa_volk_32f_x2_pow_32f (Failed)
```

argilo:
I think the problem here is that the test inputs are bad. The test framework feeds it uniformly distributed values between -1 and +1, and failures occur when the base value is -1. It doesn't make sense to allow negative bases.

[...]

PR #696, opened 2023-11-09
Fix documentation for the clamp kernel

argilo:
Two bugs prevented the clamp kernel's documentation from working correctly:

* The page name was incorrect.
* It was not added as a subpage in kernels.dox.

I've fixed both issues here, and the documentation now appears in the expected place.


argilo:
The test failure is unrelated; it's due to #649.

PR #698, opened 2023-11-09
Fix encodepolar documentation

argilo:
The page name for the `volk_8u_x3_encodepolar_8u_x2` kernel is incorrect. The dispatcher prototype name is also incorrect, and the sample code calls the `_generic` protokernel. I've corrected all these issues here.


jdemel (approved):
LGTM. Thanks for spotting these issues!

Issue #669, opened 2023-10-22
qa_volk_32f_x2_dot_prod_16i is flaky


argilo:
Example of a failed CI run: https://github.com/gnuradio/volk/actions/runs/6601430584/job/17932450061

Failures can be reproduced locally with:

```
ctest -R qa_volk_32f_x2_dot_prod_16i --output-on-failure --repeat until-fail:100000
```

Log:

```
Test project /home/argilo/git/volk/build
    Start 64: qa_volk_32f_x2_dot_prod_16i
1/1 Test #64: qa_volk_32f_x2_dot_prod_16i ......***Failed    0.07 sec
RUN_VOLK_TESTS: volk_32f_x2_dot_prod_16i(131071,1)
generic completed in 0.653972 ms
u_sse completed in 0.267041 ms
u_avx2_fma completed in 0.14574 ms
u_avx completed in 0.142483 ms
offset 0 in1: 168 in2: 167 tolerance was: 0
volk_32f_x2_dot_prod_16i: fail on arch u_sse
offset 0 in1: 168 in2: 167 tolerance was: 0
volk_32f_x2_dot_prod_16i: fail on arch u_avx2_fma
offset 0 in1: 168 in2: 167 tolerance was: 0
volk_32f_x2_dot_prod_16i: fail on arch u_avx
Best aligned arch: u_avx
Best unaligned arch: generic


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.07 sec

The following tests FAILED:
	 64 - qa_volk_32f_x2_dot_prod_16i (Failed)
```

It can be seen that the results are off by one. I would guess that the floating point result was very close to 167.5, resulting in a rounding difference, similar to #646. We may need to increase the tolerance to 1.

marcusmueller:
uff.

jdemel:
I remember @noc0lour fought hard to fix this or make the CI tests more reliable at least. Or does my memory deceive me here?

Anyways, yes, these kernels might benefit from +-1 tolerance. I wouldn't worry about this kind of deviation in a communication system. In other areas this might be an issue though.

argilo:
I added a debug statement and observed that failure occurs when the result is very close to an integer value (e.g. -112.000359). That makes sense, because the kernels just cast from `float` to `int16_t`.

The cast seems like an undesirable thing to do in a signal processing environment though, since it truncates toward zero! I think the result should be rounded.

argilo:
[...]

PR #702, opened 2023-11-10
Use absolute tolerances for accumulator and dot product

argilo:
While looking for flaky tests with `ctest --repeat until-fail:<n>`, the `qa_volk_32f_accumulator_s32f` and `qa_volk_32f_x2_dot_prod_32f` tests turned out to be flaky. Failures occur when the sum or dot product is close to zero, in which case the relative error becomes large. As with other similar kernels, I've switched these to use absolute tolerances. I took the minimum value at which tests began passing, then doubled that to leave some breathing room.

In the longer term, these tests should be switched to use a tolerance that is proportional to the vector length.


argilo:
The test failure is due to #669.

argilo:
I ran a longer `ctest --repeat until-fail:<n>`, and `volk_32fc_accumulator_s32fc` failed as well. I'll update this PR to include that, and also switch the remaining `dot_prod` kernels over to absolute tolerances since they are likely to fail the same way.

argilo:
Updated to include `volk_32fc_accumulator_s32fc`, `qa_volk_32fc_x2_dot_prod_32fc`, and `qa_volk_32fc_x2_conjugate_dot_prod_32fc`.

I ran 100000 iterations of all five tests, and no failures occurred.

jdemel (approved):
LGTM

Issue #676, opened 2023-10-22
qa_volk_16ic_x2_dot_prod_16ic is flaky


argilo:
Example of a failed CI run: https://github.com/gnuradio/volk/actions/runs/6606089847/job/17941931107

Failures can be reproduced locally with:
```
ctest -R qa_volk_16ic_x2_dot_prod_16ic --output-on-failure --repeat until-fail:100000
```

I would guess that the failure is due to different arithmetic (modular vs. saturating) across the kernels.

The test logic intentionally uses very small integers to avoid overflow/saturation conditions, but this is not foolproof when the input vectors are long:

https://github.com/gnuradio/volk/blob/42f57cd67506e7fb6a7795af3948b803de0085f4/lib/qa_utils.cc#L75-L77

I think the kernels should be adjusted so they perform the same type of arithmetic. (Saturating arithmetic everywhere?)

Log:

```
18: RUN_VOLK_TESTS: volk_16ic_x2_dot_prod_16ic(131071,1)
18: generic completed in 0.304401 ms
18: a_sse2 completed in 0.0966 ms
18: u_sse2 completed in 0.0554 ms
18: u_avx2 completed in 0.0587 ms
18: a_avx2 completed in 0.0393 ms
18: offset 0 in1: -32533 in2: -32768 tolerance was: 0
18: volk_16ic_x2_dot_prod_16ic: fail on arch a_sse2
18: offset 0 in1: -32533 in2: -32768 tolerance was: 0
18: volk_16ic_x2_dot_prod_16ic: fail on arch u_sse2
18: offset 0 in1: -32533 in2: -32768 tolerance was: 0
18: volk_16ic_x2_dot_prod_16ic: fail on arch u_avx2
18: offset 0 in1: -32533 in2: -32768 tolerance was: 0
18: volk_16ic_x2_dot_prod_16ic: fail on arch a_avx2
```

jdemel:
I agree that saturation arithmetic should be applied everywhere. It'll cause less distortion in algorithms that use convolutions.

PR #706, opened 2023-11-11
Decrease the range for signed 16-bit integer testing

argilo:
Mitigates #676.

The `qa_volk_16ic_x2_dot_prod_16ic` test is now the most frequent source of CI failures which does not already have an open pull request. The `volk_16ic_x2_dot_prod_16ic` kernel was added in #77, and because it does not handle integer overflow consistently, the test suite was modified at that time to limit signed 16-bit integer testing to the range -7 .. +7. Unfortunately, this is not sufficient to prevent overflows in the `qa_volk_16ic_x2_dot_prod_16ic`.

It would be quite a bit of work to rewrite the `volk_16ic_x2_dot_prod_16ic` kernel to have consistent overflow handling between the protokernels, so I'm proposing to decrease the signed integer range slightly (to -6 .. +6). This does not remove test failures entirely, but reduces the failure rate by about 50x.

### DIFF
--- a/lib/qa_utils.cc
+++ b/lib/qa_utils.cc
@@ -84,7 +84,7 @@ void load_random_data(void* data, volk_type_t type, unsigned int n)
             break;
         case 2:
             if (type.is_signed) {
-                std::uniform_int_distribution<int16_t> uniform_dist(-7, 7);
+                std::uniform_int_distribution<int16_t> uniform_dist(-6, 6);
                 for (unsigned int i = 0; i < n; i++)
                     ((int16_t*)data)[i] = uniform_dist(rnd_engine);
             } else {


jdemel (approved):
LGTM. Even though, this affects other kernels as well.

The saturation vs. overflow arithmetic that is mentioned in #676 is not resolved here. Thus, I leave that issue open.

argilo:
> Even though, this affects other kernels as well.

That's correct; this will slightly reduce the effectiveness of testing for all kernels with 16i inputs, but I think it's better than having a flaky test which causes CI to be ignored. Hopefully we can come up with a proper fix for #676, after which the full range of 16i values can be tested.

PR #709, opened 2023-11-14
Fix truncate-toward-zero distortion

### LINKED ISSUE #673: Truncate-toward-zero distortion in some kernels
As noted in https://github.com/gnuradio/volk/issues/669#issuecomment-1774158247 and https://github.com/gnuradio/volk/issues/669#issuecomment-1774160329, the following kernels cast directly from `float` to an integer type:

* `volk_32f_x2_dot_prod_16i`
* `volk_32fc_s32f_deinterleave_real_16i`

The truncate-toward-zero of the cast introduces distortion, as noted here: https://www.cs.cmu.edu/~rbd/papers/cmj-float-to-int.html

Most kernels avoid this problem by applying `rintf` to round the floating-point number before conversion.

I have not checked whether any other integer-output kernels have this defect.

argilo:
Fixes #673.

Most kernels with integer outputs use `rintf` to round the output. But `volk_32f_x2_dot_prod_16i` and `volk_32fc_s32f_deinterleave_real_16i` cast directly to `int16_t`, which introduces truncate-toward-zero distortion. Here I've switched the affected kernels to use `rintf` instead.

### DIFF
--- a/kernels/volk/volk_32f_x2_dot_prod_16i.h
+++ b/kernels/volk/volk_32f_x2_dot_prod_16i.h
@@ -67,7 +67,7 @@ static inline void volk_32f_x2_dot_prod_16i_generic(int16_t* result,
         dotProduct += ((*aPtr++) * (*bPtr++));
     }
 
-    *result = (int16_t)dotProduct;
+    *result = (int16_t)rintf(dotProduct);
 }
 
 #endif /*LV_HAVE_GENERIC*/
@@ -141,7 +141,7 @@ static inline void volk_32f_x2_dot_prod_16i_a_sse(int16_t* result,
         dotProduct += ((*aPtr++) * (*bPtr++));
     }
 
-    *result = (short)dotProduct;
+    *result = (short)rintf(dotProduct);
 }
 
 #endif /*LV_HAVE_SSE*/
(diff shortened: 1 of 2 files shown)


argilo:
GNU Radio uses `volk_32f_x2_dot_prod_16i` in the Float->Short variants of the Decimating & Interpolating FIR Filter blocks. Distortion can be observed with the following flow graph:
Screenshot from 2023-11-14 13-26-17

Before:
Screenshot from 2023-11-14 13-26-27

After:
Screenshot from 2023-11-14 13-35-46

argilo:
It's a bit easier to see what's going on with the floating point signal in the picture as well:

Screenshot from 2023-11-14 20-32-55

Before:

Screenshot from 2023-11-14 20-32-21

After:

Screenshot from 2023-11-14 20-27-46

jdemel (approved):
LGTM

argilo:
It turns out there are two GNU Radio tests which assert the truncate-toward-zero behaviour:

https://github.com/gnuradio/gnuradio/blob/9b22fd38b8259f33928c7b3ff7868e0bc4b60d1e/gr-filter/python/filter/qa_fir_filter.py#L160-L188

Those tests now fail. I'll have to think about what the best solution is.

argilo:
The failing tests were fixed in https://github.com/gnuradio/gnuradio/pull/6980.

PR #714, opened 2023-12-08
Remove broken sse_32 protokernels

argilo:
`volk_32fc_x2_conjugate_dot_prod_32fc` fails in [i386 Debian CI](https://buildd.debian.org/status/fetch.php?pkg=volk&arch=i386&ver=3.1.0-1&stamp=1701982056&raw=0) and locally. Only the `a_sse_32` protokernel is broken:

```
108: offset 0 in1: 184.314 + -210.344j  in2: 183.667 + -210.042j tolerance was: 0.02
108: volk_32fc_x2_conjugate_dot_prod_32fc: fail on arch a_sse_32
```

I had a look over the code, and it appears the problem is that inline assembler writes to `input`, `taps`, and `*result` without telling the compiler about that. I made a few attempts at fixing it, but I'm not sure whether it's worth saving this ugly code. I doubt this protokernel sees much use, since it is only available on 32-bit systems, and many newer protokernels are now available on 32-bit x86.

It appears `volk_32fc_x2_dot_prod_32fc_a_sse_64` had a similar bug, as it was commented out in 2012 and remains commented out today:

https://github.com/gnuradio/gnuradio/commit/43224ac7be478cd93f46720377d95f1e53aef87e

I've removed that protokernel here as well. After this change, there are no more `sse_32` protokernels left in VOLK.


argilo:
I checked whether GNU Radio uses the `volk_32fc_x2_conjugate_dot_prod_32fc` kernel, and it only recently started doing so in https://github.com/gnuradio/gnuradio/pull/3306. That's another good indication that the buggy `volk_32fc_x2_conjugate_dot_prod_32fc_a_sse_32` protokernel hasn't seen any use.

jdemel (approved):
LGTM. Less assembly should improve maintenance as well. I wonder if this kernel is ever used nowadays.

PR #716, opened 2023-12-09
Fix flaky mod_range test

argilo:
`volk_32fc_s32f_magnitude_16i` fails in locally on i386:

```
offset 9838 in1: 323.859 in2: 327 tolerance was: 1e-06
volk_32f_s32f_mod_rangepuppet_32f: fail on arch u_sse
offset 9838 in1: 323.859 in2: 327 tolerance was: 1e-06
volk_32f_s32f_mod_rangepuppet_32f: fail on arch a_sse
offset 9838 in1: 323.859 in2: 327 tolerance was: 1e-06
volk_32f_s32f_mod_rangepuppet_32f: fail on arch u_sse2
offset 9838 in1: 323.859 in2: 327 tolerance was: 1e-06
volk_32f_s32f_mod_rangepuppet_32f: fail on arch a_sse2
offset 9838 in1: 323.859 in2: 327 tolerance was: 1e-06
volk_32f_s32f_mod_rangepuppet_32f: fail on arch u_avx
offset 9838 in1: 323.859 in2: 327 tolerance was: 1e-06
volk_32f_s32f_mod_rangepuppet_32f: fail on arch a_avx
```

This happens because the input range (-1 .. +1) maps to the output ranges (325.664 .. 327) and (323.859 .. 324.523), with a wrap-around occurring when the input is equal to 0.336. When an input close to 0.336 occurs, floating-point error could cause the output to be close to either 327 or 323.859, and the test can fail.

To avoid this situation, I've adjusted the lower bound such that the wrap-around condition is avoided. After this change, (-1 .. +1) maps to the continuous range (324.624 .. 326.624) and the test passes reliably on i386.

### DIFF
--- a/kernels/volk/volk_32f_s32f_mod_rangepuppet_32f.h
+++ b/kernels/volk/volk_32f_s32f_mod_rangepuppet_32f.h
@@ -19,7 +19,7 @@ static inline void volk_32f_s32f_mod_rangepuppet_32f_generic(float* output,
                                                              unsigned int num_points)
 {
     volk_32f_s32f_s32f_mod_range_32f_generic(
-        output, input, bound - 3.141f, bound, num_points);
+        output, input, bound - 3.131f, bound, num_points);
 }
 #endif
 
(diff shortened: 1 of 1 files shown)


marcusmueller (approved):

jdemel (approved):
LGTM

PR #721, opened 2023-12-11
Express version information in decimal

### LINKED ISSUE #720: VOLK_VERSION is too hard to actually use.
Maybe it would be better as a string.

The following code demonstrates that following the comments leads to unexpected results.
// -----
#include <volk/volk.h>
#include <iostream>

int main(int argc, char *argv[])
{
  // No complaints yet...
  std::cout << "Using Volk: " <<
    VOLK_VERSION_MAJOR << "." <<
    VOLK_VERSION_MINOR << "." <<
    VOLK_VERSION_MAINT << "\n";
  /*
   * VOLK_VERSION % 100 is the MAINT version
   * (VOLK_VERSION / 100) % 100 is the MINOR version
   * (VOLK_VERSION / 100) / 100 is the MAJOR version
   */
  std::cout << "Using VOLK_VERSION: " <<
    VOLK_VERSION << ":  " <<
    (VOLK_VERSION  / 100)/100 << "  " <<
    (VOLK_VERSION /100) %100 << "  " <<
    (VOLK_VERSION % 100) << " ...Hmm. Not Decimal.\n";
  // Uh... No.
  unsigned int v = VOLK_VERSION;
  std::cout << "It's octal: 0o0" << (v>>12)%8
	   << ((v>>9)%8)*8 <<  ((v>>6)%8) 
	   << ((v>>3)%8)*8 <<  v%8 << " ?\n";
  std::cout << "Or, base64: " << (v>>12)%64 << "."
	   << ((v>>6)%64) << "." << v%64 << "\n";
  // cout is interpreting these values as decimal, what could possibly go wrong?
  return(0);
}
// -----

Some observed output:

Using Volk: 3.0.0
Using VOLK_VERSION: 12288:  1  22  88 ...Hmm. Not Decimal.
It's octal: 0o030000 ?
Or, base64: 3.0.0.

Using Volk: 2.5.2
Using VOLK_VERSION: 8514:  0  85  14 ...Hmm. Not Decimal.
It's octal: 0o020502 ?
Or, base64: 2.5.2

So, yeah. Maybe revise code and/or comments in the include file,
/usr/include/volk/volk_version.h

argilo:
Fixes #720.

VOLK's version header (created in #346) unintentionally expresses version numbers in octal, because it prefixes the (decimal) numbers with zeroes. This will fail (with `error: invalid digit "8" in octal constant`) as soon as any component of the version number reaches 8.

The documentation in the header states that `VOLK_VERSION` is decimal, and this is a more convenient format, so I think it makes sense to switch to that. GNU radio already uses `#if VOLK_VERSION >= 030100`, but this condition will continue to be true if VOLK switches to decimal because 30100<sub>10</sub> is greater than 030100<sub>8</sub> (= 12352<sub>10</sub>).

I considered instead changing the header documentation to indicate that the format is octal, but it would be difficult to implement correct header generation in CMake, because CMake's `math` only supports decimal and hexadecimal output formats.


argilo:
Before:
```
#define VOLK_VERSION_MAJOR 03
#define VOLK_VERSION_MINOR 01
#define VOLK_VERSION_MAINT 00

[...]

#define VOLK_VERSION 030100
```

After:
```
#define VOLK_VERSION_MAJOR 3
#define VOLK_VERSION_MINOR 1
#define VOLK_VERSION_MAINT 0

[...]

#define VOLK_VERSION 30100
```

argilo:
GNSS-SDR is the one other user of VOLK that I know of, and it doesn't use the `VOLK_VERSION` macros.

marcusmueller (approved):
Let's do this, before the current state causes any more damage.

jdemel (approved):
LGTM

jdemel:
This code might be affected:
https://github.com/jj1bdx/airspy-fmradion/blob/c6ca61c600dc48f0e0af2322c61fe6ab16e25c3b/sfmbase/MultipathFilter.cpp#L139

argilo:
> This code might be affected

That will continue working, for the same reason that GNU Radio's checks will. Future decimal versions numbers will be greater than 030100<sub>8</sub>.

PR #725, opened 2023-12-15
Remove SSE protokernels written in assembly

argilo:
Only three protokernels remain that are written entirely in assembly:

* `volk_32fc_x2_conjugate_dot_prod_32fc_a_sse`
* `volk_32fc_x2_dot_prod_32fc_u_sse_64`
* `volk_32fc_x2_dot_prod_32fc_a_sse_64`

These are holdovers from the early days of GNU Radio, and I don't think keeping them is worth the maintenance cost. Both `volk_32fc_x2_conjugate_dot_prod_32fc` and `volk_32fc_x2_dot_prod_32fc` have SSE3 protokernels which perform better, and SSE3 has been around since 2004.


marcusmueller (approved):

jdemel (approved):
LGTM. I think we discussed expected CPU features lately. In essence, 32bit x86 is probably hard to find nowadays. It's reasonable to remove the maintenance burden here.
64bit x86 requires at least SSE2. A fact that caused issues in the past as well.

PR #726, opened 2023-12-15
Build and test neonv7 protokernels on armv7

argilo:
The neon/neonv7 kernels are not built or tested in CI, because a toolchain file is not passed to CMake. Here I've added one. I had to change the compiler to the OS default, because the toolchain file specifies:

```
set(CMAKE_CXX_COMPILER g++)
set(CMAKE_C_COMPILER  gcc)
```

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -96,7 +96,8 @@ jobs:
             compiler: { name: clang-14, cc: clang-14, cxx: clang++-14 }
           - arch: armv7
             distro: ubuntu22.04
-            compiler: { name: g++-12, cc: gcc-12, cxx: g++-12 }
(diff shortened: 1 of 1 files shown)


argilo:
If desired, we could add configurations to test other toolchain files as well.

jdemel:
Thanks for this PR. It'd be great to test our toolchain files. We had issues regarding them in the past that were difficult to answer. Tests might give an indicator on how to use them etc.

jdemel (approved):
LGTM. I was worried for a bit that the `cmakeargs` part would interfere but this shouldn't be a problem in case it is left empty.

argilo:
Yes, that should be fine. From GitHub's docs:

> If a configuration variable has not been set, the return value of a context referencing the variable will be an empty string.

PR #727, opened 2023-12-15
Remove inline assembler from volk_32fc_convert_16ic_neon

argilo:
In #296, NEON code was removed from volk_32fc_convert_16ic and replaced with significantly slower inline assembler, so that the `neon` protokernel would produce identical output to the others when rounding numbers whose fractional part is exactly 0.5. Elsewhere in VOLK, we prioritize speed and accept that floating-point calculations will not produce identical results. In particular, every other kernel that converts floating-point values to integers has its test tolerance set to 1 to account for the fact that rounding results may be different when the fractional part of the floating-point value is close to 0.5.

I propose to restore the NEON code and set the output tolerance to 1, as we do elsewhere. In addition to speeding up the calculation, this will remove the maintenance burden associated with inline assembler. (See #589, #593 for a case where it has already caused trouble.)

Before:
```
RUN_VOLK_TESTS: volk_32fc_convert_16ic(131071,1987)
neon completed in 1874.38 ms
generic completed in 5753.56 ms
Best aligned arch: neon
Best unaligned arch: neon
```

After:
```
RUN_VOLK_TESTS: volk_32fc_convert_16ic(131071,1987)
neon completed in 980.321 ms
generic completed in 5742.38 ms
Best aligned arch: neon
Best unaligned arch: neon
```


jdemel (approved):
LGTM. I think we changed our test policy. In the past we were more aiming towards accuracy. Now, we strike a balance that is more focused towards "does it perform well in a communications engineering context".

argilo:
Agreed. Over time, various changes like #342 have been moving VOLK in the direction of higher DSP performance, and I think that's the right direction for the users that I know of (GNU Radio, Gqrx, GNSS-SDR, airspy-fmradion).

PR #728, opened 2023-12-16
Use bit shifts in generic and byte_shuffle reverse

argilo:
Using bit shifts instead of bit fields seems to be much more compiler-friendly, and results in large speedups with both GCC & Clang, across x86_64, aarch64, and armhf. When compiling with Clang, the generic protokernel comes close to being the fastest protokernel on all three architectures.

| Compiler | generic before | generic after | byte_shuffle before | byte_shuffle after |
|----- | -----: | -----: | -----: | -----: |
| GCC x86_64 | 2687 | 389 | 3193 | 167 |
| Clang x86_64 | 1398 | 46 | 4883 | 168 |
| GCC aarch64 (Pi 5) | 3406 | 1887 | 4007 | 458 |
| Clang aarch64 (Pi 5)| 1187 | 78 | 5089 | 111 |
| GCC armhf (Pi 4) | 6674 | 4331 | 5771 | 963 |
| Clang armhf (Pi 4) | 3322 | 321 | 6994 | 1022 |

### DIFF
--- a/kernels/volk/volk_32u_reverse_32u.h
+++ b/kernels/volk/volk_32u_reverse_32u.h
@@ -28,50 +28,6 @@
  * \endcode
  */
 #ifndef INCLUDED_VOLK_32u_REVERSE_32u_U_H
-struct dword_split {
-    int b00 : 1;
-    int b01 : 1;
-    int b02 : 1;
-    int b03 : 1;
-    int b04 : 1;
-    int b05 : 1;
-    int b06 : 1;
-    int b07 : 1;
-    int b08 : 1;
-    int b09 : 1;
-    int b10 : 1;
-    int b11 : 1;
-    int b12 : 1;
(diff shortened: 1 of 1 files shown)


marcusmueller:
Obvious win! It's interesting how much clang and gcc diverge here, and also, that there's no consistency in which implementation wins. Fun fact, looking at disassebly, here: clang reworks the the generic bit shift implementation into the `bintree_permute_top_donw` code, basically, for AVX2 (which explains why run times are near identical).

marcusmueller (approved):

argilo:
Clang is also smart enough to use `rbit` on ARM, which explains why the generic protokernel does so well.

jdemel (approved):
LGTM. A nice performance win. Also, this is an interesting data point that Clang might produce faster code than GCC.

PR #731, opened 2023-12-17
volk_32fc_s32f_atan2_32f: Add NaN tests for avx2 and avx2_fma code

### LINKED ISSUE #730: v3.1.0 volk_32fc_s32f_atan2_32f.h avx2 and avx2_fma kernels return NaN for an input element 0+0j
## Synopsis

VOLK v3.1.0 volk_32fc_s32f_atan2_32f() returns -nan for the following kernels when an input element is 0+0j:

* avx2 (a_avx2, u_avx2)
* avx2_fma (a_avx2_fma, u_avx2_fma)

VOLK v3.1.0 volk_32fc_s32f_atan2_32f() returns zero for the following kernels when an input element is 0+0j:

* generic
* polynomial

(where `j^2 = -1`)

Tested on Ubuntu 22.04.3 x86_64.

Test details are available at: <https://gist.github.com/jj1bdx/62e27aac4b54a29dfafe210c73c49b0e>

## What should be done

The avx2 and avx2_fma kernels should behave the same as the generic and polynomial kernels, i.e., returning zero when the input is 0+0j. I have to screen this irregularity in my own application as in [this example](https://github.com/jj1bdx/airspy-fmradion/blob/d780acbb28dd8cd095660dc8c36ab2452e738ca2/sfmbase/FmDecode.cpp#L282-L283), which affects the overall performance and introduces unnecessary complexity. 

I would appreciate it if the original author of the kernel would kindly help fix this bug.

## References

* Issue #636
* [Speeding up atan2f by 50x](https://mazzo.li/posts/vectorized-atan2.html)
  - In the [Edge Cases](https://mazzo.li/posts/vectorized-atan2.html#edge-cases) section of this article, the author explicitly states that: "All our functions do not handle inputs containing infinity, or when both x and y are 0, while conforming implementations must handle them specially". So, this is assumed to be *a known issue*.

jj1bdx:
This adds a NaN test after the division operation of volk_32fc_s32f_atan2_32f avx2 and avx2_fma code.

This patch solves the issue in #730.

Using volk_profile, the processing time will increase by 17% on my environment of Intel NUC NUC10i7FNH with Intel(R) Core(TM) i7-10710U CPU. I believe this is a necessary cost to maintain full compatibility with the past (i.e., VOLK v3.0.0 and previous) implementations.

### DIFF
--- a/kernels/volk/volk_32fc_s32f_atan2_32f.h
+++ b/kernels/volk/volk_32fc_s32f_atan2_32f.h
@@ -117,6 +117,7 @@ static inline void volk_32fc_s32f_atan2_32f_a_avx2_fma(float* outputVector,
     const __m256 pi_2 = _mm256_set1_ps(0x1.921fb6p0f);
     const __m256 abs_mask = _mm256_castsi256_ps(_mm256_set1_epi32(0x7FFFFFFF));
     const __m256 sign_mask = _mm256_castsi256_ps(_mm256_set1_epi32(0x80000000));
+    const __m256 zero = _mm256_setzero_ps();
 
     unsigned int number = 0;
     unsigned int eighth_points = num_points / 8;
@@ -133,6 +134,8 @@ static inline void volk_32fc_s32f_atan2_32f_a_avx2_fma(float* outputVector,
             _mm256_and_ps(y, abs_mask), _mm256_and_ps(x, abs_mask), _CMP_GT_OS);
         __m256 input = _mm256_div_ps(_mm256_blendv_ps(y, x, swap_mask),
                                      _mm256_blendv_ps(x, y, swap_mask));
+        __m256 nan_mask = _mm256_cmp_ps(input, input, _CMP_UNORD_Q);
+        input = _mm256_blendv_ps(input, zero, nan_mask);
         __m256 result = _m256_arctan_poly_avx2_fma(input);
 
         input =
@@ -174,6 +177,7 @@ static inline void volk_32fc_s32f_atan2_32f_a_avx2(float* outputVector,
     const __m256 pi_2 = _mm256_set1_ps(0x1.921fb6p0f);
     const __m256 abs_mask = _mm256_castsi256_ps(_mm256_set1_epi32(0x7FFFFFFF));
     const __m256 sign_mask = _mm256_castsi256_ps(_mm256_set1_epi32(0x80000000));
+    const __m256 zero = _mm256_setzero_ps();
 
     unsigned int number = 0;
     unsigned int eighth_points = num_points / 8;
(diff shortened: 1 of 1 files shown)


jj1bdx:
volk_profile benchmark results:

With old v3.1.0 code:

```
RUN_VOLK_TESTS: volk_32fc_s32f_atan2_32f(131071,1987)
generic completed in 7078.28 ms
polynomial completed in 2318.11 ms
a_avx2_fma completed in 175.141 ms
a_avx2 completed in 175.124 ms
u_avx2_fma completed in 174.965 ms
u_avx2 completed in 176.404 ms
Best aligned arch: u_avx2_fma
Best unaligned arch: u_avx2_fma
```

With the code proposed in this pull request:

```
RUN_VOLK_TESTS: volk_32fc_s32f_atan2_32f(131071,1987)
generic completed in 7075.16 ms
polynomial completed in 2343.78 ms
a_avx2_fma completed in 204.169 ms
a_avx2 completed in 204.303 ms
u_avx2_fma completed in 204.389 ms
u_avx2 completed in 204.4 ms
Best aligned arch: a_avx2_fma
Best unaligned arch: u_avx2_fma
```

Processing time increase rate: 204/175 ~= 1.166

jdemel:
According to [Wikipedia atan2](https://en.wikipedia.org/wiki/Atan2), one would expect a NaN for `0.0+0.0j`. Changing this behavior in VOLK would potentially cause trouble.

Before the new AVX2 and AVX2 FMA kernels, we would basically always use the [atan2 function from the standard](https://en.cppreference.com/w/c/numeric/math/atan2).

```c
atan2f(0.0, 0.0) == 0.0;
```
Thus, this fix makes sense to align the ouputs.

argilo:
Normally VOLK strives for maximum performance, but I agree that we should avoid returning NaN, because that can be catastrophic in DSP contexts. (For instance, if an IIR filter receives a single NaN sample, then all future outputs become NaN.)

marcusmueller:
I must admit I especially don't like the fallback on the generic implementation at the end of loops here. This kernel is used for phase estimations, and minor implementation details (like whether to jump to -pi or +pi for zero x, positive or negative y) makes it hard to debug this and to use it. Could we maybe just for the "leftover" iteration fill a set MM register with the inputs that are available, run the regular SIMD on it, then instead of storing the whole result, just cherry-pick these elements that are actually based on input?

marcusmueller:
might be worth checking what sleef does here.

jdemel (approved):
LGTM. This PR fixes situation where a NaN may corrupt a system. We should really avoid that.

The other concerns, e.g. how to handle the tail case, are valid but out of scope.

jj1bdx:
@jdemel Thanks for merging!

PR #724, opened 2023-12-14
Remove unnecessary volatiles from volk_32fc_s32f_magnitude_16i

argilo:
Several variables in `volk_32fc_s32f_magnitude_16i` were marked as volatile in an attempt to resolve a flaky test:

https://github.com/gnuradio/volk/pull/282#issuecomment-532558206
https://github.com/gnuradio/volk/pull/282#issuecomment-532790680
https://github.com/gnuradio/volk/pull/285

These changes did not actually resolve the flaky test; it still failed on i386, and the test itself was later fixed in #713. Therefore, we can now revert the additions of `volatile` that were made in #285.

PR #734, opened 2023-12-17
Remove __VOLK_VOLATILE

argilo:
The last usage of `__VOLK_VOLATILE` was removed in #724. I don't think VOLK has any good reason to use `volatile`; it's intended for use in hardware environments where things other than the currently running program might modify the contents of memory. In the past, VOLK has occasionally misused `volatile` to block compiler optimizations, but this hurts performance and doesn't make broken code any less broken. So I think it would make sense to remove `__VOLK_VOLATILE`.


marcusmueller (approved):

jdemel (approved):
LGTM

PR #636, opened 2023-10-04
new kernels for atan2

Ka-zam:
New kernels for atan2 based on the recently merged arctan work. Almost 40x speedup.

With this PR:
```
$ volk_profile -R atan2

RUN_VOLK_TESTS: volk_32fc_s32f_atan2_32f(131071,1987)
generic completed in 5091.65 ms
polynomial completed in 2138.51 ms
a_avx2_fma completed in 131.781 ms
a_avx2 completed in 131.696 ms
u_avx2_fma completed in 131.963 ms
u_avx2 completed in 132.086 ms
Best aligned arch: a_avx2
Best unaligned arch: u_avx2_fma
```

Without:
```
$ volk_profile -R atan2

RUN_VOLK_TESTS: volk_32fc_s32f_atan2_32f(131071,1987)
a_sse4_1 completed in 5159.66 ms
a_sse completed in 5168.12 ms
generic completed in 5201.91 ms
Best aligned arch: a_sse4_1
Best unaligned arch: generic
```

PR #735, opened 2023-12-17
Remove references to simdmath library

### LINKED ISSUE #732: Remove LV_HAVE_LIB_SIMDMATH?
As noted in https://github.com/gnuradio/volk/pull/636#issuecomment-1761066135, some kernels use `LV_HAVE_LIB_SIMDMATH` and include `simdmath.h`. My best guess is that this is related to some experimental GNU Radio work done around 2009: https://github.com/gnuradio/volk/pull/636#issuecomment-1859183031. At this point, maybe we should just delete the protokernels that use `LV_HAVE_LIB_SIMDMATH`. Two of them were already removed in #636.

argilo:
Fixes #732.

A number of protokernels depend on `simdmath.h`, and fall back to a generic implementation if it is not present. My best guess is that this was a proprietary library that was used for some experiments around 2009: https://github.com/gnuradio/volk/pull/636#issuecomment-1859183031

Since these protokernels are effectively equivalent to the existing generic versions, I think we should remove them and get rid of the maintenance burden.


marcusmueller (approved):

jdemel (approved):
LGTM. Thanks for trying to figure out where this is coming from.

PR #736, opened 2023-12-17
Fix bugs in convolutional decoder

### LINKED ISSUE #473: volk_8u_x4_conv_k7_r2_8u generic kernel inconsistent with SSE implementation

### LINKED ISSUE #733: volk_8u_conv_k7_r2puppet is broken
`volk_8u_conv_k7_r2puppet`, which tests the `volk_8u_x4_conv_k7_r2_8u` kernel, has its input and output arguments (`dec` & `syms`) reversed. As a result, it always tests `volk_8u_x4_conv_k7_r2_8u` with `syms = [0, 0, ... 0]`, which is a very poor test input and has allowed serious bugs to slip through, like the one fixed in #458.

If the argument order is fixed, then even the `spiral` protokernel begins failing. I'm having a look now to see why its behaviour differs from the `generic` protokernel.

argilo:
Fixes #733.
Fixes #473.
Reverts #475.

VOLK's convolutional decoder has many bugs:

1. The test puppet has the input & output reversed, so it always tests the kernel with zeroes as input. Once this is corrected, many more bugs become apparent.
2. The `spiral` and `neonspiral` protokernels only renormalize the branch metrics if the first branch metric is greater than 210. This threshold is much too high, and results in constant overflows and incorrect decisions. The threshold was previously removed from the generic protokernel, so I've done the same for the others. This causes a ~20% performance penalty, but it's necessary for correct output.
3. The adjustment to `METRICSHIFT` made in #475 changed the scaling of metrics for the generic protokernel but not the SIMD protokernels. This causes inaccuracy, because the SIMD protokernels use the generic calculation to when processing the tail end of the input.
4. The protokernels differ in how they add metrics for the two parity bits:
    * The generic protokernel uses `(metric1 >> 1) + (metric2 >> 1) = (metric1 + metric2) >> 1`.
    * The SIMD protokernels use `_mm_avg_epu8`, which calculates `(metric1 + metric2 + 1) >> 1`.
5. The generic protokernel uses `>` for metric comparison, while the SIMD protokernels use `>=`.

After all these bugs are fixed, the convolutional decoder is tested with random input and produces identical output between the three protokernels.

I left the `avx2` protokernel commented out for now because I haven't yet investigated why it's broken. But it at least now fails if it is uncommented. (Previously it passed when it received zeroes as input.)

### DIFF
--- a/kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
+++ b/kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
@@ -97,8 +97,8 @@ static inline int chainback_viterbi(unsigned char* data,
 #include <stdio.h>
 #include <xmmintrin.h>
 
-static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms,
-                                                      unsigned char* dec,
+static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* dec,
+                                                      unsigned char* syms,
                                                       unsigned int framebits)
 {
     if (framebits < 12) {
@@ -181,8 +181,8 @@ static inline void volk_8u_conv_k7_r2puppet_8u_spiral(unsigned char* syms,
 
 #include "volk/sse2neon.h"
 
-static inline void volk_8u_conv_k7_r2puppet_8u_neonspiral(unsigned char* syms,
-                                                          unsigned char* dec,
+static inline void volk_8u_conv_k7_r2puppet_8u_neonspiral(unsigned char* dec,
+                                                          unsigned char* syms,
                                                           unsigned int framebits)
 {
     if (framebits < 12) {
@@ -266,8 +266,8 @@ static inline void volk_8u_conv_k7_r2puppet_8u_neonspiral(unsigned char* syms,
 //#include <immintrin.h>
 //#include <stdio.h>
 //
-// static inline void volk_8u_conv_k7_r2puppet_8u_avx2(unsigned char* syms,
-//                                                    unsigned char* dec,
+// static inline void volk_8u_conv_k7_r2puppet_8u_avx2(unsigned char* dec,
+//                                                    unsigned char* syms,
 //                                                    unsigned int framebits)
 //{
 //    if (framebits < 12) {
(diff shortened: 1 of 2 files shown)


argilo:
I also removed the unused `threshold` argument from the `renormalize` function, which cleans up one of the few remaining compiler errors.

marcusmueller (approved):
I really can't claim I understand the SPIRAL code, but your forcing of renormalization seems to make sense to me, unless this leads to rounding of many different errors to 0 and thus reduced performance in high-SNR?

Anyway, if this actually does correctly work now, we're one big step ahead. And we should totally merge this PR.

argilo:
> your forcing of renormalization seems to make sense to me, unless this leads to rounding of many different errors to 0

There's no rounding; the renormalization is just an integer subtraction to keep the path metrics from overflowing an `unsigned char`.

argilo:
> I really can't claim I understand the SPIRAL code

Now that I've read and understood it, I'm not sure it's doing much good compared to a human-readable version. It's not that different, except it unrolls the main loop (so that two butterfly operations are done in each iteration) and it splits out every single operation onto its own line, e.g.:

```
        a73 = (4 * i9);
        b6 = (syms + a73);
        a75 = *(b6);
```

A human would write `a75 = syms[4 * i9]` and I'm sure the compiler would do just as well (if not better) with that.

Aang23:
@argilo as mentioned on Matrix, I did all my tests with this PR. Both using `generic` and `spiral`. Everything behaves as expected on my end :-)

argilo:
@Aang23 I've pushed the AVX2 to fixes to this branch: https://github.com/argilo/volk/tree/k7-r2-avx2

I'm not sure whether to include those in this PR. Maybe they should wait until later. Unfortunately the AVX2 code is slower than the spiral implementation. The renormalization code is getting to be quite costly since it happens on every iteration, and isn't very SIMD-friendly.

jdemel (approved):
LGTM. FEC code is hard.

PR #737, opened 2023-12-18
Get the sse2neon.h file to a git submodule to avoid random copies.

balister:
sse2neon.h is maintained at https://github.com/DLTcollab/sse2neon. This commit replaces the file copied from this repo with a submodule adding the repo to the volk source tree. The files that use the header file need the include path changing from volk/sse2neon.n to volk/sse2neon/sse2neon.h.


jdemel (approved):
LGTM. Makes sense to add a submodule instead of a copy.

PR #742, opened 2024-01-12
cmake: Switch to GNUInstallDirs

### LINKED ISSUE #723: volk CMakeLists.txt uses CMAKE_INSTALL_LIBDIR without including the 'GNUInstallDirs' module
I am running Linux Fedora Core here, where for 64bit libraries LIBDIR is expected have the suffix '64' (i.e. `/usr/local/lib64`).
I noticed today that instead volk installed the shared library and the cmake config files under `/usr/local/lib`.

I then looked at the volk configuration for cmake (`CMakeLists.txt`), and I saw that there's code there to decide between `lib` vs `lib64` (https://github.com/gnuradio/volk/blob/main/CMakeLists.txt#L205-L210), however that codes uses the variable `CMAKE_INSTALL_LIBDIR`.
According to the cmake documentation (https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html) that variable is defined in the `GNUInstallDirs` module, but I don't see that cmake module included anywhere, and indeed when I added a `message()` statement to print the value of `CMAKE_INSTALL_LIBDIR` right before the check above, the value turned out to be empty.

Thanks,
Franco

jdemel:
The custom code to determine `lib` vs `lib64` folder names on different distros was error prone. There's a better tool for that. Let's use it and potentially fix more issues that arise.

Fixes #723


argilo:
I tested this with PyBOMBS and it works fine there.

jdemel:
All tests pass. It's a little change. Let's merge it.

fventuri:
I just cloned the 'main' branch, built, and installed volk here on Fedora Core, and this time the shared libraries and related files get correctly installed in the right place ('/usr/local/lib64'):

```
-- Installing: /usr/local/lib64/cmake/volk/VolkConfig.cmake
-- Installing: /usr/local/lib64/cmake/volk/VolkConfigVersion.cmake
-- Installing: /usr/local/lib64/cmake/volk/VolkTargets.cmake
-- Installing: /usr/local/lib64/cmake/volk/VolkTargets-release.cmake
-- Installing: /usr/local/lib64/libvolk.so.3.1.0
-- Up-to-date: /usr/local/lib64/libvolk.so.3.1
-- Up-to-date: /usr/local/lib64/libvolk.so
```

Thanks for fixing it!
Franco

PR #743, opened 2024-01-12
cmake: Fix 64bit host CPU detection

### LINKED ISSUE #683: lib/CMakeLists.txt: check_type_size("void*[8]" SIZEOF_CPU BUILTIN_TYPES_ONLY) empty
This leads to the folliowing cmake failure in a ninja/clang/Coverage build attempt

```
CMake Error at lib/CMakeLists.txt:248 (if):
  if given arguments:

    "EQUAL" "8"

  Unknown arguments specified
```

This checking seems redund, see CMAKE_SIZEOF_VOID_P.

jdemel:
In cases where we don't cross-compile, we might want to detect if a CPU is 32bit or 64bit. CMake provides functionality for this case starting in CMake 3.10. Let's use it.
From the top of my head, the oldest supported distribution are:
- Ubuntu 20.04 with CMake 3.16
- Debian buster with CMake 3.13.

Fixes #683


jdemel:
Since this PR bumps the minimum required CMake version, this should go into a feature release, i.e. VOLK v3.2.

PR #755, opened 2024-02-01
Replace sse2neon with native NEON

argilo:
`volk_8u_x4_conv_k7_r2_8u_neonspiral` is the only protokernel that needs sse2neon, which has proven to be a troublesome dependency. Here I've rewritten the protokernel in native NEON, eliminating the need for sse2neon.

Now that the `volk_8u_x4_conv_k7_r2_8u` kernel is adequately tested, we can be confident that the updated protokernel is correct.

Performance is more or less unchanged:

Old:
```
neonspiral completed in 4557.97 ms
generic completed in 29817.1 ms
```

New:
```
neonspiral completed in 4591.51 ms
generic completed in 29832.3 ms
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
 [submodule "cpu_features"]
 	path = cpu_features
 	url = https://github.com/google/cpu_features.git
-[submodule "include/volk/sse2neon"]
-	path = include/volk/sse2neon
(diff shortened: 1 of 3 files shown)


jdemel:
Merging this PR and waiting for #756 to be rebased. I appreciate that the code seems to simplify.

PR #758, opened 2024-02-24
docs: Fix upload to webserver

noc0lour:
This also enables storing of different versions of the documentation in subdirectories. 

Currently they will be available in `libvolk.org/test-docs/{branch,tag}/{refname}/`. 
So, currently there is a version in www.libvolk.org/test-docs/branch/noc0lour/build-docs-ci/ from this branch. 
Whenever a new version is built it will be created in a new subdirectory (based on the current date) and the symlink is switched to it on the webserver.

I did not add a version selector to the documentation HTML (maybe someone is especially motivated to do so), so access to different branches & versions is limited to people who know. Probably we could link to the versions from the main page.


jdemel:
Thanks for taking care of it. Looks good.

PR #764, opened 2024-06-25
cmake: Suppress invalid escape sequence warnings with Python 3.12

### LINKED ISSUE #763: Warnings with Python 3.12
With Python 3.12 (on Ubuntu 24.04), some new warnings have appeared.

```<string>:1: SyntaxWarning: invalid escape sequence '\W'```

It seems that the strings located here:

https://github.com/gnuradio/volk/blob/main/cmake/Modules/VolkPython.cmake#L147

and here:

https://github.com/gnuradio/volk/blob/main/lib/CMakeLists.txt#L108

need to made "raw". Like so:

```-print(re.sub('\\W', '_', '${desc} ${reldir} ' + unique))"```
```+print(re.sub(r'\\W', '_', '${desc} ${reldir} ' + unique))"```

```-                        "import re; print(re.sub('\\W', '_', '${have_flag}'))"```
```+                        "import re; print(re.sub(r'\\W', '_', '${have_flag}'))"```

drmpeg:


jdemel:
@drmpeg Thank you very much! The PR looks good to me. Merging.

PR #766, opened 2024-07-27
cmake/pkgconfig: use CMAKE_INSTALL_FULL_* variables

doronbehar:
See: https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,12 +220,6 @@ include(GNUInstallDirs)
 ########################################################################
 # Setup the package config file
 ########################################################################
-#set variables found in the pc.in file
-set(prefix ${CMAKE_INSTALL_PREFIX})
-set(exec_prefix "\${prefix}")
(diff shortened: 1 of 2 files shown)


jdemel (approved):

jdemel:
Besides, cleaning up the code, does this PR fix anything in particular?

doronbehar:
> Besides, cleaning up the code, does this PR fix anything in particular?

Yes. It fixes the build on NixOS. Our build system has automatic checks for pkg config files, that make sure our pkg-config files have valid paths. Not knowing the subtle difference between `CMAKE_INSTALL_FULL_*` variables and the variables that were used before this patch, is a common mistake detected by our build system, and without this check and this patch, the pkg config files generated by volk before this patch are incorrect.

You may be able to create an invalid path if you'd build volk with a non trivial prefix - as done in NixOS (And Guix) for every package. See: https://nixos.org/guides/how-nix-works/

PR #767, opened 2024-08-19
Add googletest based tests

### LINKED ISSUE #356: Test with short, medium and large vectors
Since we sometimes see issues with kernels on very short vector sizes and sometimes on larger vector sizes, we should run all tests on ~3 vector sizes. I suggest 127, 8091, and 2^17-1 (current default). I hope that catches a few more bugs.

### LINKED ISSUE #690: Move to better test infrastructure
In a lot of issues and PRs we discuss problems with our current tests.

We need to discuss a way forward to improve this situation.

One option would be to introduce [gtest](https://github.com/google/googletest). We would write specific tests for some kernels first and adopt an approach where we slowly move to the new system.

jdemel:
This PR adds googletest as a dependency and starts to implement new tests. This should offer more flexibility and thus, better support for more complex kernels and better coverage of corner cases. Finally, different vector sizes should be easily supported.

For now, this is a second set of tests that still needs to evolve but the foundation is in place.

Finally, the CI system receives an update with newer compiler versions and older versions being removed. Here, the target is to test on Ubuntu versions that still receive general support. At the moment the oldest distro would be Ubuntu 20.04.


jdemel:
This PR was sitting here for long enough. Merging it. At this point, some changes are kinda urgent to fix the failing flyci test.

PR #774, opened 2024-10-22
Implement RISC-V Vector 1.0 kernels

### LINKED ISSUE #772: RISC-V Vector 1.0 Support: If and where to start?
Hello,

RISC-V's Vector extension was ratified a few years back and recently vector-supporting boards have come out, many based on the octo-core Spacemit K1/M1. I have been using one such board for a while now, with gnuradio and more, but the performance is quite lacking: when profiling with volk_profile, the generic is around an order of magnitude faster than the alternatives

All that to say, I think RVV 1.0 has many instructions useful for volk and I am willing to help but I have no idea where to start. 
If this is in the project's plans, is there a roadmap?

camel-cdr:
This PR adds full RVV vectorization to all kernels, excluding deprecated ones and `volk_32f_s32f_power_32f`  and `volk_32fc_s32f_power_32fc`  which don't have any vectorized kernels in for other architectures, presumably due to precision requirements.

All the tests pass in qemu with VLEN 128 up to 1024, and were additionally tested on all input sizes from 0 to 1000. They have been integrated into CI, with both clang and gcc.

I've attached the output of volk_profile running on the [SpacemiT X60](https://camel-cdr.github.io/rvv-bench-results/bpi_f3/index.html) cores of the Banana Pi BPI-F3 SOC: [X60-gcc-14.txt](https://github.com/user-attachments/files/17481827/X60-gcc-14.txt). [X60-clang-18.txt](https://github.com/user-attachments/files/17481828/X60-clang-18.txt)
The average speedup across all kernels compared to the previously fastest one was 3.8x.

The code is written using RVV 1.0 intrinsics, following the frozen [v0.12](https://github.com/riscv-non-isa/rvv-intrinsic-doc/releases/tag/v0.12.0) spec, which is supported by gcc 14 and clang 18 and above. The build system was adjusted to detect support for RVV intrinsics, to make sure we don't break builds on older compilers.

I tried to maximize LMUL without causing register spills, while avoiding lane crossing permutations.

Segmented load/stores varies in performance a lot on current systems, compare the vsseg graph between the [C910](https://camel-cdr.github.io/rvv-bench-results/milkv_pioneer/ascii_to_utf16.html) and [X60 ](https://camel-cdr.github.io/rvv-bench-results/bpi_f3/ascii_to_utf16.html). To get the best performance I didn't use it in the rvv target, but created an additional pseudo target rvvseg, with alternative implementations using segmented load/stores. 

I didn't modify the existing kernels, except for the following cases:

* `volk_32u_popcntpuppet_32u.h` and `volk_64u_popcntpuppet_64u.h` had a bug where they didn't compute anything.
* `volk_8u_conv_k7_r2puppet_8u.h` created a lookup table with 256 entries once, but only used for a total of 64 lookups, so I replaced it with the direct calculation.

This should resolve https://github.com/gnuradio/volk/issues/772


cmake/Toolchains/rv64gcv-linux-gnu.cmake
```diff
@@ -0,0 +1,34 @@
...
+    CACHE INTERNAL "size tool")
+
+set(CMAKE_FIND_ROOT_PATH ${BINUTILS_PATH})
+
+set(QEMU_VLEN $ENV{VLEN})
+if(NOT QEMU_VLEN)
+    set(QEMU_VLEN "128")
+endif()
+
+set(CMAKE_CROSSCOMPILING_EMULATOR "qemu-riscv64-static -L /usr/riscv64-linux-gnu/ -cpu rv64,v=on,vlen=${QEMU_VLEN},rvv_ta_all_1s=on,rvv_ma_all_1s=on")
```
luhenry: It's generally good to enable `zba` and `zbb` as well as they are available on all boards that have `v` and they bring up some nice performance boost on scalar code. You can enable with `zba=true,zbb=true`

camel-cdr: Ah, good idea, I'll add that.
I didn't think to include it, because `google/cpu_features` currently doesn't support detecting the bitmanip extensions.
I tried to patch this upstream, but couldn't figure out how to sign the CLA without a google account: https://github.com/google/cpu_features/pull/369

Their extension detection is fundamentally broken anyways, see: https://github.com/google/cpu_features/pull/368 (It would currently parse `rv64gc_xmycustomextensionwithavsomewhere` as `rv64gcv`)

Maybe you could forward this to the Google people at RISE.
`google/cpu_features` should probably use hwprobe instead of erroneously parsing cpuinfo, and it would be amazing if support for profiles and extension groups was added.

drmpeg:
Wow, I gotta get my BPI-F3 set up. I'll be testing with GNU Radio.

camel-cdr:
Thanks to [cloud-v.co](https://cloud-v.co/) I got access to a machine with [C910](https://camel-cdr.github.io/rvv-bench-results/milkv_pioneer/) cores, which doesn't support RVV 1.0, but rather the XTheadVector custom extension based on the RVV 0.7.1 draft specification, with VLEN=128.
gcc allows you to compile (some) RVV 1.0 intrinsics directly to XTheadVector, which allowed me to benchmark on the CPU as well.
These numbers are interesting because the C910 is an dual issue out-of-order CPU compared to the in-order X60.

The XTheadVector target for gcc isn't perfect, so I couldn't get every kernel to compile properly, but here are the results from the once that worked: 
```
                                  before   RVV speedup                                           before  RVV speedup                                before  RVV speedup
32f_s32f_convert_16i                4457   128 34.770   32f_reciprocal_32f                         1570  393 3.99455   64f_x2_min_64f                 1760 1070 1.64526
32f_sqrt_32f                       12923   381 33.843   32f_x2_min_32f                             1702  427 3.97885   32fc_magnitude_32f             1574 1000 1.57403
32f_log2_32f                       14068   469 29.972   16ic_s32f_magnitude_32f                    1651  509 3.2431    64f_x2_max_64f                 1768 1162 1.52126
32f_s32f_convertpuppet_8u           4057   152 26.646   32i_s32f_convert_32f                        412  130 3.1648    32f_index_max_16u               138   90 1.51793
32f_expfast_32f                     3141   127 24.621   32f_sin_32f                                6903 2183 3.16171   32f_s32f_multiply_32f           265  206 1.28825
32f_s32f_convert_8i                 3235   138 23.365   32fc_s32f_magnitude_16i                    3016  954 3.16168   16ic_deinterleave_16i_x2        588  515 1.14099
16i_s32f_convert_32f                1509    69 21.566   32f_x2_powpuppet_32f                       9276 2957 3.13613   32fc_x2_divide_32fc            3177 2832 1.1216
32fc_convert_16ic                   8910   471 18.909   32f_cos_32f                                6726 2191 3.06938   32f_x2_add_32f                  407  395 1.03269
16ic_magnitude_16i                 10076   574 17.531   16ic_s32f_deinterleave_32f_x2              3048 1006 3.02943   32f_x2_multiply_32f             416  403 1.03164
32f_s32f_convert_32i                2096   130 16.024   8ic_deinterleave_real_16i                   266   88 2.99781   32f_x2_subtract_32f             422  451 0.935161
32f_s32f_clamppuppet_32f            1845   123 14.889   32f_s32f_add_32f                            273   99 2.75712   32fc_deinterleave_64f_x2       1786 2032 0.878856
8ic_x2_multiply_conjugate_16ic      2224   234 9.4952   32f_x2_divide_32f                          1575  586 2.68736   32fc_conjugate_32fc             412  483 0.853893
32f_exp_32f                         7848   857 9.1584   16ic_deinterleave_real_16i                  269  103 2.59342   32fc_index_max_32u              787  949 0.829749
32f_asin_32f                       17823  2289 7.7839   32f_index_min_32u                           290  115 2.52049   64f_x2_add_64f                  694 1017 0.682631
32f_acos_32f                       18806  2476 7.5945   32f_s32f_normalize                          311  127 2.44517   32fc_deinterleave_real_32f      279  430 0.649423
32f_8u_polarbutterflypuppet_32f   107451 14240 7.5455   32f_index_max_32u                           281  115 2.44065   64f_x2_multiply_64f             705 1107 0.637096
32f_tan_32f                        15314  2428 6.3061   32fc_x2_s32fc_multiply_conjugate_add2_32fc 1172  520 2.25319   32i_x2_and_32i                  265  426 0.623177
8ic_deinterleave_real_8i             271    52 5.1638   32f_x2_fm_detectpuppet_32f                  959  426 2.24964   32fc_x2_add_32fc                707 1175 0.602147
16i_convert_8i                       262    52 4.9657   32f_convert_64f                             278  126 2.19516   32i_x2_or_32i                   261  442 0.590342
32f_s32f_stddev_32f                  393    84 4.6301   16ic_s32f_deinterleave_real_32f             397  185 2.14181   64f_convert_32f                 266  454 0.586313
32f_x2_s32f_interleave_16ic         3825   862 4.4377   32fc_index_max_16u                          458  218 2.10224   32fc_deinterleave_imag_32f      267  484 0.55243
32f_stddev_and_mean_32f_x2           839   199 4.2042   32fc_s32f_deinterleave_real_16i            1968  938 2.09685   32fc_deinterleave_32f_x2        920 1724 0.533851
8ic_deinterleave_16i_x2              527   125 4.1961   16ic_deinterleave_real_8i                   262  125 2.08729   32f_64f_multiply_64f            445  847 0.526016
16ic_convert_32fc                    524   126 4.1332   32f_binary_slicer_32i                       266  131 2.02717   32fc_deinterleave_real_64f      266  538 0.495517
32fc_s32f_atan2_32f                 8978  2224 4.0358   32f_atan_32f                               1557  829 1.87756   32fc_magnitude_squared_32f      399  868 0.459629
32f_x2_max_32f                      1703   422 4.0325   32f_index_min_16u                           171   91 1.87702   32fc_x2_multiply_conjugate_32fc 751 2453 0.306232
8ic_x2_s32f_multiply_conjugate_32fc 1820   454 4.0008   32f_x2_interleave_32fc                      432  231 1.86693   32fc_32f_multiply_32fc          533 1795 0.297167
32f_tanh_32f                        1573   393 3.9980   8u_x3_encodepolarpuppet_8u                 3318 1803 1.83999   32fc_x2_multiply_32fc           737 2761 0.266857
```
Keep in mind that the codegen for XTheadVector is often worse than for RVV 1.0, because additional instructions need to be inserted.
Still these speedups looks quite good, some things are a lot faster, and a few are a lot slower than scalar, but that's what `volk_profile` is fore.
Interesting to note is that the `rvvseg` variant was, as expected, always slower than the `rvv` variants, because the segmented load/store implementation is quite bad on the processor.


Edit: The latest force push fixed register spills in two kernels

.github/workflows/run-tests-rvv.yml
```diff
@@ -0,0 +1,54 @@
...
+
+name: Run VOLK tests on different RVV configurations
+
+on: [push, pull_request]
+
+jobs:
+  Tests:
+    runs-on: ubuntu-24.04
+    steps:
+    - uses: actions/checkout@v4
```
jdemel, line 17:
```suggestion
    - uses: actions/checkout@v4
      with:
        submodules: "recursive"
```
This would probably make the best use of github actions to use integrated features.

include/volk/volk_rvv_intrinsics.h
jdemel, line 1: This file among others doesn't pass our formatting CI check yet. It should be really easy to do so though. clang-format with `-style=file` should fix it. Most IDEs probably help with config like "format on save".

camel-cdr:
I ran clang-format on all changed files, so the formatting issues should hopeful be fixed now.

include/volk/volk_rvv_intrinsics.h
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+#ifndef INCLUDE_VOLK_VOLK_RVV_INTRINSICS_H_
+#define INCLUDE_VOLK_VOLK_RVV_INTRINSICS_H_
+#include <riscv_vector.h>
+
+#define RISCV_SHRINK2(op, T, S, v)              \
+    __riscv_##op(__riscv_vget_##T##S##m1(v, 0), \
+                 __riscv_vget_##T##S##m1(v, 1), \
+                 __riscv_vsetvlmax_e##S##m1())
```
jdemel, line 22: Is there a specific reason why you use macros? If possible, I'd like to avoid them because they look like functions from the outside but aren't. They are another beast to tame. Functions are usually the better approach. I'd argue that compilers inline functions whenever told/possible nowadays without the need of macros. Or did you use macros as C-style template functions? How many different instantiations are we looking at now?

jdemel:
> I ran clang-format on all changed files, so the formatting issues should hopeful be fixed now.

Thanks! Also, I was doing a review and got an "outdated file" (or smth similar) error at some point. Was already worried that comments got lost. Luckily they aren't. Thanks for taking care of the formatting stuff.

kernels/volk/volk_16ic_deinterleave_16i_x2.h
```diff
@@ -375,4 +375,45 @@ static inline void volk_16ic_deinterleave_16i_x2_u_avx2(int16_t* iBuffer,
...
+#include <riscv_vector.h>
+
+static inline void volk_16ic_deinterleave_16i_x2_rvvseg(int16_t* iBuffer,
+                                                        int16_t* qBuffer,
+                                                        const lv_16sc_t* complexVector,
+                                                        unsigned int num_points)
+{
+    size_t n = num_points;
+    for (size_t vl; n > 0; n -= vl, complexVector += vl, iBuffer += vl, qBuffer += vl) {
+        vl = __riscv_vsetvl_e16m4(n);
```
jdemel, line 408: So no more epilogue/tail handling? We can just calculate how many values we can compute in parallel? That's really nice.

kernels/volk/volk_16ic_deinterleave_16i_x2.h
```diff
@@ -375,4 +375,45 @@ static inline void volk_16ic_deinterleave_16i_x2_u_avx2(int16_t* iBuffer,
...
+#ifdef LV_HAVE_RVVSEG
+#include <riscv_vector.h>
+
+static inline void volk_16ic_deinterleave_16i_x2_rvvseg(int16_t* iBuffer,
+                                                        int16_t* qBuffer,
+                                                        const lv_16sc_t* complexVector,
+                                                        unsigned int num_points)
+{
+    size_t n = num_points;
+    for (size_t vl; n > 0; n -= vl, complexVector += vl, iBuffer += vl, qBuffer += vl) {
```
jdemel, line 407: Can you default initialize `vl`? `size_t vl = 0`?

kernels/volk/volk_32f_index_min_32u.h
```diff
@@ -42,7 +42,7 @@
  *
  *   volk_32f_index_min_32u(out, in, N);
  *
- *   printf("minimum is %1.2f at index %u\n", in[*out], *out);
+ *   ("minimum is %1.2f at index %u\n", in[*out], *out);
```
jdemel: It looks like smth went missing here. Was this change intended?

kernels/volk/volk_32u_popcntpuppet_32u.h
```diff
@@ -18,10 +18,8 @@ static inline void volk_32u_popcntpuppet_32u_generic(uint32_t* outVector,
                                                      const uint32_t* inVector,
                                                      unsigned int num_points)
 {
-    unsigned int ii;
-    for (ii = 0; ii < num_points; ++ii) {
-        volk_32u_popcnt_generic(outVector + ii, *(inVector + ii));
-    }
+    for (size_t i = 0; i < num_points; ++i)
+        volk_32u_popcnt_generic(outVector + i, inVector[i]);
```
jdemel: I really prefer braces around one-liners. This bit me before and I usually have a hard time to read it.

kernels/volk/volk_32u_reverse_32u.h
```diff
@@ -337,4 +337,57 @@ volk_32u_reverse_32u_arm(uint32_t* out, const uint32_t* in, unsigned int num_poi
...
+#include <riscv_vector.h>
+
+static inline void
+volk_32u_reverse_32u_rva23(uint32_t* out, const uint32_t* in, unsigned int num_points)
+{
+    size_t n = num_points;
+    for (size_t vl; n > 0; n -= vl, in += vl, out += vl) {
+        vl = __riscv_vsetvl_e32m8(n);
+        vuint32m8_t v = __riscv_vle32_v_u32m8(in, vl);
+        __riscv_vse32(out, __riscv_vbrev(v, vl), vl);
```
jdemel, line 388: So `__riscv_vbrev` is an intrinsic for reversal that's introduced with RVA23? Neat.

kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
```diff
@@ -20,11 +20,14 @@ typedef union {
...
-    x ^= (x >> 16);
-    x ^= (x >> 8);
-    return Partab[x];
+    x ^= x >> 16;
+    x ^= x >> 8;
+    x ^= x >> 4;
+    x ^= x >> 2;
+    x ^= x >> 1;
+    return x & 1;
 }
```
jdemel, line 31: Could you give a quick explanation or reference for this change? I like it. It's for documentation purposes.

lib/CMakeLists.txt
```diff
@@ -93,12 +93,28 @@ execute_process(
...
+    HAVE_RVV_INTRINSICS
+    ${CMAKE_BINARY_DIR}
+    ${CMAKE_SOURCE_DIR}/cmake/Checks/check-rvv-intrinsics.c
+)
+if(HAVE_RVV_INTRINSICS)
+    message(STATUS "Checking RVV intrinsics - found")
+else()
+    message(STATUS "Checking RVV intrinsics - not found")
+endif()
+
```
jdemel, line 106: I assume CMake doesn't have a function for this check yet. Would it make sense to only check for RVV if `"${arch_prefix}" STREQUAL "rv"`

jdemel (commented):
That's a very impressive PR. Thank you!

I have a few minor comments. Besides, the PR looks good to me.

jdemel:
You probably need to rebase your PR on the latest `main`. The CI seems to have received a few changes that need fixing to make it work again.

.github/workflows/run-tests-rvv.yml
camel-cdr: Since you mentioned "outdated file", how do you prefer I implement the edits?
Should I continue amending the commit and force pushing, or create multiple commits, or something else?

jdemel: It's up to you. My personal preference is to have commits that tell a story.

The "outdated file" issue was a concurrency issue on the GitHub side of things. I wouldn't worry about it. It was just a funny coincidence.

include/volk/volk_rvv_intrinsics.h
camel-cdr: It needs to be generic, so it can't be a regular function.
The occurrences are:
```
RISCV_SHRINK2: 6
RISCV_SHRINK4: 27
RISCV_SHRINK8: 6
RISCV_PERM4: 1
RISCV_PERM8: 6
RISCV_LUT4: 4
```
The shrink macros could be removed in theory, but current hardware implements LMUL>1 reductions non-optimally.I measured the X60 to take ~54 cycles for vfredusum on LMUL=8, while it takes ~6.5 cycles on LMUL=1 and ~1 cycle for LMUL=1 vfadd. So the SHRINK8 macro can be used to get from ~54 cycles to ~6.5+7=13.5 cycles.
I don't expect this to be a problem in future hardware and IIRC all of the SHRINK macros are mostly used in epilogue code, where this may not matter as much.

The PERM and LUT macros are important, because the vrgather operation needs to do quadratically more work with higher LMUL. So it's best to unroll it at LMUL=1 if you don't need to cross register lanes.

kernels/volk/volk_16ic_deinterleave_16i_x2.h
camel-cdr: Yes, this works quite well for most of the kernels.

kernels/volk/volk_16ic_deinterleave_16i_x2.h
camel-cdr: I could, but this is how the reference implementations currently write it:
https://github.com/riscv-non-isa/rvv-intrinsic-doc/blob/main/examples/rvv_branch.c#L15

kernels/volk/volk_32u_reverse_32u.h
camel-cdr: Yes, I recommend using this amazing website as a reference for the intrinsics: https://dzaima.github.io/intrinsics-viewer/
Edit: To clarify, it was introduced in the Zvbb (vector bitmanip) extension, which became mandatory for the RVA23 profile.

kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
camel-cdr: I just noticed that the kernels are doing a lot of redundant work.
To calculate the parity of 64 numbers, the code first fills a lookup table of 256 numbers with the result of a parity calculation to optimize the parity calculation of the initial 64 numbers in the same codepath.

lib/CMakeLists.txt
camel-cdr: That would require moving the check into the loop and checking it for every rvv architecture, that seems a bit redundant and IMO it's easier to read when it's not crammed into the nested statements.

include/volk/volk_rvv_intrinsics.h
jdemel: Thanks for the explanation. I guess we need it for now then.

kernels/volk/volk_16ic_deinterleave_16i_x2.h
jdemel: OK. We can leave it as is. It's funny though that you contributed to this example and point to it now.

camel-cdr: Oh, true, but I only cleaned the examples up a bit, this paradigm was there before.

kernels/volk/volk_8u_conv_k7_r2puppet_8u.h
jdemel: I'm a bit worried because we might break a test in an already fragile kernel. Still, it makes sense, let's go with your solution.

.github/workflows/run-tests-rvv.yml
```diff
@@ -0,0 +1,55 @@
...
+        sudo apt-get install -y python3-mako cmake qemu-user-static g++-14-riscv64-linux-gnu clang-18
+        mkdir build
+        cd build
+    - name: Test gcc-14 VLEN=128
+      run: |
+        cd build; rm -rf *
+        CXX=riscv64-linux-gnu-g++-14 CC=riscv64-linux-gnu-gcc-14 VLEN=128 \
+        cmake -DCMAKE_TOOLCHAIN_FILE=../cmake/Toolchains/rv64gcv-linux-gnu.cmake ..
+        make -j$(nproc)
+        ARGS=-j$(nproc) make test
```
jdemel: I just noticed that this tests run without the `verbose` flag. That'd be great. e.g. one would get a bit more information which kernels are actually executed in each test.
A second little thing: looking at all the other CI tests, we use `ctest -V`. Since these tests usually run rather quickly, it is nice to have them run sequentially. Since the whole CI test here is reported to run for ~5min, I assume this `-j$(nproc)` doesn't contribute much to the overall runtime.

lib/CMakeLists.txt
jdemel: I think we can leave it as is. If anyone runs into issues because CMake is way to slow, we can still fix it.

drmpeg:
I got my BPI-F3 running today. I'm using Bianbu 2.0.1, but I've ripped out the desktop and just using it headless. Here's my volk_profile results.

[BPI-F3-gcc-14.txt](https://github.com/user-attachments/files/17606731/BPI-F3-gcc-14.txt)

Should have GNU Radio running tomorrow.

drmpeg:
Tested with GNU Radio. Works perfectly for at least these kernels:
```
volk_32fc_32f_multiply_32fc
volk_32fc_x2_add_32fc
volk_32f_s32f_multiply_32f
volk_32fc_magnitude_32f
volk_32fc_s32fc_multiply2_32fc
volk_32fc_s32fc_multiply_32fc
volk_32f_x2_subtract_32f
volk_32fc_x2_multiply_conjugate_32fc
volk_32f_x2_add_32f
volk_32fc_x2_multiply_32fc
```

drmpeg (approved):

jdemel (approved):

jdemel:
Thanks for this huge contribution! I'll do a release soon. I hope you don't mind that I mention your name (as it appears in the sign-off) in the release changelog.

camel-cdr:
> Thanks for this huge contribution! I'll do a release soon. I hope you don't mind that I mention your name (as it appears in the sign-off) in the release changelog.

Yeah, that be great.

PR #776, opened 2024-11-13
appveyor: Update to VS 2022/Python 3.12

tinyboxvk:
- Update VS from 2019 to VS 2022
- Update Python to 3.12
  - Python 3.8 is EOL
- Remove "\\" from cmake command
  - Resolves CMake Warning  `Ignoring extra path from command line:  "/"` and  `Ignoring extra path from command line:  "\"`

This also reduces build time by roughly 20 seconds from my testing (~3m to ~2m40s).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,16 +5,16 @@
 #
 # SPDX-License-Identifier: LGPL-3.0-or-later
 #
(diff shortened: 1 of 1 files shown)


jdemel (approved):
Thanks for your contribution! Looks good to me.

PR #777, opened 2024-11-13
CI: Update Android Build

tinyboxvk:
- Do not install dependencies that are already installed
- Use `actions/setup-java` to configure required Java version
- Bump Command-Line Tools to the latest version
  - Requires Java 17
- Bump NDK to latest LTS
  - https://developer.android.com/ndk/downloads
- Bump SDK Platform to 34


jdemel (approved):
Thanks for your contribution. I'm happy to see a new contributor! Looks good to me.

PR #790, opened 2025-06-03
Add tool to compare profiles

marcusmueller:
I had this lying around and $daytimejob used it, so why not put it out in the open?

I use a small python script to compare `volk_config`s from different machines.

### DIFF
--- a/scripts/tools/compare_volk_profiles
+++ b/scripts/tools/compare_volk_profiles
@@ -1,0 +1,69 @@
+#!/bin/env python3
+# Copyright 2022, 2025 Marcus Müller
+# SPDX-License-Identifier: GPL-3.0
+# Takes in a list of volk profiles, ignores empty and identical lines, prints a table of differences.
+# Hacky as hell
+
+from sys import argv
+
+
+def keepline(line: str) -> bool:
+    if not line:
+        return False
+    line = line.strip()
+    if line.startswith("#"):
+        return False
+    return True
+
+
+def kernel(line: str) -> str:
+    return line.split(" ")[0]
+
+
+def impls(line: str) -> tuple[str, str]:
+    return tuple(line.strip().split(" ")[1:])
+
+
+machines = [
+    {kernel(line): impls(line) for line in open(f_name) if keepline(line)}
+    for f_name in argv[1:]
+]
+kernels = [set(d.keys()) for d in machines]
+common_kernels = [
+    kernel for kernel in kernels[0] if all((kernel in ks for ks in kernels[1:]))
+]
+
+differing_kernels = dict()
+for kernel in common_kernels:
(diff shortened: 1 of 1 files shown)


scripts/tools/compare_volk_profiles
```diff
@@ -0,0 +1,71 @@
+#!/bin/env python3
+# Copyright 2022, 2025 Marcus Müller
+# SPDX-License-Identifier: GPL-3.0
```
jdemel, line 3: Would it be possible to have this file under LGPL?

scripts/tools/compare_volk_profiles
```diff
@@ -0,0 +1,71 @@
...
+# Takes in a list of volk profiles, ignores empty and identical lines, prints a table of differences.
+# Hacky as hell
+
+from sys import argv
```
jdemel, line 7: argparse would probably be the way to go. That'd kinda fix your "hacky as hell" comment.

scripts/tools/compare_volk_profiles
```diff
@@ -0,0 +1,71 @@
...
+#from json import dumps
```
jdemel: You probably want to remove this line since it is unused.

scripts/tools/compare_volk_profiles
```diff
@@ -0,0 +1,71 @@
...
+max_kernel_len = max(len(kernel) for kernel in common_kernels)
+max_impl_len = max(
+    max(max(len(alignment) for alignment in impl) for impl in kernel.values())
+    for kernel in differing_kernels.values()
+)
+
+# print(dumps(machines, indent=2))
+# print(dumps(common_kernels))
+# print(dumps(differing_kernels, indent=2))
+# print(max_impl_len)
```
jdemel: you probably want to remove these lines.

jdemel (commented):
Mostly LGTM.

The license of the file is the part I'm a bit worried about. Other than that, you can fix the comments or not. It'd be OK.

scripts/tools/compare_volk_profiles
marcusmueller: no. It's also not something that ends up in the binaries, at all, so no need.

scripts/tools/compare_volk_profiles
marcusmueller: No, that's not what's hacky here; and this script simply takes positional arguments, all of which are identical. So, not winning anything by pulling in one of the more problematic python standard library modules.

scripts/tools/compare_volk_profiles
marcusmueller: (It contains code that was given to me under the assumption I'd release it under a strong copyleft. So, not my call to make. Take it or leave it; it's not that important functionality.)

PR #803, opened 2025-12-05
More useful reporting and fail summary

Ka-zam:
Be more explicit with test parameters and format better, in this case tol should probably be tightened up quite a bit:
```
RUN_VOLK_TESTS: volk_16i_32fc_dot_prod_32fc(vlen=131071, iter=1987, tol=1e-01)
arch           |        time |    throughput |   max_err |
---------------+-------------+---------------+-----------+
generic        |  91.5739 ms |  51194.3 MB/s |         - |
u_sse          |  77.0061 ms |  60879.2 MB/s |   7.4e-06 |
u_avx2_fma     |  23.3280 ms | 200962.8 MB/s |   1.0e-05 | *
u_avx2         |  23.7096 ms | 197728.9 MB/s |   1.0e-05 |
a_sse          |  76.0999 ms |  61604.1 MB/s |   7.4e-06 |
a_avx2         |  24.1527 ms | 194100.7 MB/s |   1.0e-05 |
a_avx2_fma     |  24.6400 ms | 190262.3 MB/s |   1.0e-05 |
Best aligned arch  :    u_avx2_fma (3.93x)
Best unaligned arch:    u_avx2_fma (3.93x)
--------------------------------------------------------------------------------
```

Show input and output values for failed kernels (I forced an add of 0.1 to the result):
```
$ volk_profile -R acos

RUN_VOLK_TESTS: volk_32f_acos_32f(vlen=131071, iter=1987, tol=1e-02)
arch           |        time |    throughput |   max_err |
---------------+-------------+---------------+-----------+
generic        |1789.6891 ms |   1164.2 MB/s |         - |
a_avx2_fma     | 248.0666 ms |   8399.3 MB/s |   1.2e-01 |
a_avx          | 246.8268 ms |   8441.5 MB/s |   5.4e-04 |
a_sse4_1       | 491.0029 ms |   4243.5 MB/s |   5.4e-04 |
u_avx2_fma     | 246.8047 ms |   8442.2 MB/s |   5.4e-04 |
u_avx          | 246.4952 ms |   8452.8 MB/s |   5.4e-04 | *
u_sse4_1       | 489.6534 ms |   4255.2 MB/s |   5.4e-04 |
Best aligned arch  :    u_avx (7.26x)
Best unaligned arch:    u_avx (7.26x)

volk_32f_acos_32f: fail on arch a_avx2_fma
  index |       in0 |  expected |    actual |  rel_err |      tol
--------+-----------+-----------+-----------+----------+---------
 131064 |    0.6881 |    0.8119 |    0.9119 |  1.2e-01 |  1.0e-02
 131065 |   -0.9130 |    2.7213 |    2.8213 |  3.7e-02 |  1.0e-02
 131066 |   -0.2300 |    1.8028 |    1.9028 |  5.5e-02 |  1.0e-02
 131067 |   -0.8256 |    2.5421 |    2.6421 |  3.9e-02 |  1.0e-02
 131068 |    0.2114 |    1.3578 |    1.4578 |  7.4e-02 |  1.0e-02
 131069 |    0.3655 |    1.1967 |    1.2967 |  8.4e-02 |  1.0e-02
 131070 |    0.2410 |    1.3274 |    1.4274 |  7.5e-02 |  1.0e-02
--------------------------------------------------------------------------------
```

### DIFF
--- a/lib/qa_utils.cc
+++ b/lib/qa_utils.cc
@@ -440,63 +440,64 @@ inline void run_cast_test3_s32fc(volk_fn_3arg_s32fc func,
 }
 
 template <class t>
-bool fcompare(t* in1, t* in2, unsigned int vlen, float tol, bool absolute_mode)
+bool fcompare(t* expected,
+              t* actual,
+              unsigned int vlen,
+              float tol,
+              bool absolute_mode,
+              std::vector<unsigned int>& fail_indices,
+              double& max_rel_err)
 {
     bool fail = false;
-    int print_max_errs = 10;
+    max_rel_err = 0.0;
     for (unsigned int i = 0; i < vlen; i++) {
-        // Check for special values (NaN, inf)
-        bool in1_special = std::isnan(((t*)(in1))[i]) || std::isinf(((t*)(in1))[i]);
-        bool in2_special = std::isnan(((t*)(in2))[i]) || std::isinf(((t*)(in2))[i]);
+        t exp_val = expected[i];
+        t act_val = actual[i];
 
-        if (in1_special || in2_special) {
+        // Check for special values (NaN, inf)
+        bool exp_special = std::isnan(exp_val) || std::isinf(exp_val);
+        bool act_special = std::isnan(act_val) || std::isinf(act_val);
+
+        bool this_fail = false;
+        double rel_err = 0.0;
+        if (exp_special || act_special) {
             // For NaN: both must be NaN (NaN != NaN, so use isnan)
             // For inf: both must be same signed infinity
             bool values_match =
-                (std::isnan(((t*)(in1))[i]) && std::isnan(((t*)(in2))[i])) ||
-                (((t*)(in1))[i] == ((t*)(in2))[i]);
+                (std::isnan(exp_val) && std::isnan(act_val)) || (exp_val == act_val);
             if (!values_match) {
-                fail = true;
-                if (print_max_errs-- > 0) {
-                    std::cout << "offset " << i << " in1: " << t(((t*)(in1))[i])
-                              << " in2: " << t(((t*)(in2))[i]);
-                    std::cout << " tolerance was: " << tol << std::endl;
-                }
+                this_fail = true;
+                rel_err = std::numeric_limits<double>::infinity();
             }
-            continue; // Skip normal comparison for special values
-        }
-
-        if (absolute_mode) {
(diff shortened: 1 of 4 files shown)


lib/qa_utils.cc
```diff
@@ -440,164 +440,324 @@ inline void run_cast_test3_s32fc(volk_fn_3arg_s32fc func,
...
+    if (fail_indices.empty())
+        return;
+
+    // Column widths
+    const int w_idx = 7;
+    const int w_val = 10;
+    const int w_sci = 9;
+
+    // Print header
+    std::cout << std::setw(w_idx) << "index";
```
jdemel: Since we rely on C++20 for our tests: https://github.com/gnuradio/volk/blob/5fdfbc340e718a2415a8923141f7964120a32fc5/tests/CMakeLists.txt#L29
You could use std::format to make your life easier.

jdemel (commented):
Thanks for the PR.

I'm a bit worried that this PR will change the behavior of our tests in some subtle way that will ultimately lead to new issues.
However, in the meantime we have infrastructure to add specialized tests to cover this kind of issue. Thus, we'd be able to catch these issues specifically.

The failing formatting test is probably trivial to fix.

lib/qa_utils.cc
Ka-zam: As I understand  it GCC11 and 12 don't implement std::format and those are in CI. But fmt::format is already a dependency so we could use that. Or drop GCC11/12.

Ka-zam:
> Thanks for the PR.
> 
> I'm a bit worried that this PR will change the behavior of our tests in some subtle way that will ultimately lead to new issues. However, in the meantime we have infrastructure to add specialized tests to cover this kind of issue. Thus, we'd be able to catch these issues specifically.
> 
> The failing formatting test is probably trivial to fix.

Well, I did some special input value injection for the atan2 issues and ultimately that's what found all the broken kernels for arg min/max as well. We could try to identify kernels suscpetible to NaN, Inf, out-of-domain values and test those specifically?

lib/qa_utils.cc
jdemel: fmt::format should be fine.

jdemel:
Can you fix the formatting CI? I'd argue we can merge this PR otherwise.

Ka-zam:
RVV kernel maps 0.0f input to NaN but generic clamps this to -127.0f, needs fixing.

PR #816, opened 2026-02-24
cmake: fix `FetchContent`/`add_subdirectory` subproject breakages

### LINKED ISSUE #815: [CMake error] `FetchContent_MakeAvailable(volk)` breaks subsequent `FetchContent_MakeAvailable`
### Summary

I'm trying to use VOLK (as a static library) in my project, and for that I'm using `FetchContent_MakeAvailable(volk)`.  
After `volk` is made available, a subsequent `FetchContent_MakeAvailable(...)` (e.g. `spdlog`) fails with:
```
CMake Error at /usr/share/cmake/Modules/FetchContent.cmake:1591 (message):
  Internal error: SUBBUILD_DIR not set
  /usr/share/cmake/Modules/FetchContent.cmake:2155:EVAL:2 (__FetchContent_doPopulation)
  /usr/share/cmake/Modules/FetchContent.cmake:2155 (cmake_language)
  /usr/share/cmake/Modules/FetchContent.cmake:2394 (__FetchContent_Populate)
  CMakeLists.txt:26 (FetchContent_MakeAvailable)
```

This appears to be caused by VOLK mutating global CMake directory variables.

### Environment
- CMake: `4.2.3`
- OS: Linux (x86_64)
- VOLK: `main` (`511fff053e190316919733fb148b70946f139a9e`)

### Minimal Reproducer

`CMakeLists.txt`:
```cmake
cmake_minimum_required(VERSION 3.24)
project(volk_fetchcontent_repro LANGUAGES CXX)

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)

include(FetchContent)

# Avoid unrelated cpu_features min-CMake issue on this setup
set(VOLK_CPU_FEATURES OFF CACHE BOOL "Disable cpu_features for repro" FORCE)

# 1) Fetch VOLK first
FetchContent_Declare(
  volk
  GIT_REPOSITORY https://github.com/gnuradio/volk.git
  GIT_TAG main
)
FetchContent_MakeAvailable(volk)

# 2) Then fetch another project (spdlog)
FetchContent_Declare(
  spdlog
  GIT_REPOSITORY https://github.com/gabime/spdlog.git
  GIT_TAG v1.17.0
)
FetchContent_MakeAvailable(spdlog)

add_executable(repro main.cpp)
target_link_libraries(repro PRIVATE volk spdlog::spdlog)
```

`main.cpp`:
```cpp
#include <volk/volk.h>
#include <spdlog/spdlog.h>

int main()
{
    spdlog::info("Includes from VOLK and spdlog are visible");
    return 0;
}
```

### Repro Command
```console
cmake --fresh -S . -B build
```

### Actual Result
Configuration fails at the second `FetchContent_MakeAvailable(spdlog)` with  `Internal error: SUBBUILD_DIR not set`.

### Expected Result
`FetchContent_MakeAvailable(spdlog)` should work normally after `FetchContent_MakeAvailable(volk)`.

### Additional Notes
-   `spdlog` alone configures successfully.
-   `spdlog` first, then `volk`, also configures successfully.
-   So this appears order-dependent (`volk` first triggers it).

### Suspected Root Cause
VOLK top-level `CMakeLists.txt` reassigns global CMake vars:
```cmake
set(CMAKE_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR}) #allows this to be a sub-project
set(CMAKE_BINARY_DIR ${CMAKE_CURRENT_BINARY_DIR}) #allows this to be a sub-project
```

These should not be overridden by a subproject and likely break `FetchContent` internals for later dependencies.

### Suggested Fix
Remove those assignments and use `CMAKE_CURRENT_SOURCE_DIR` / `CMAKE_CURRENT_BINARY_DIR` directly where needed.

xerpi:
## Summary

This MR fixes two CMake subproject issues in VOLK when used via `FetchContent_MakeAvailable()` / `add_subdirectory()`.

In short:
1. Remove `CMakeParseArgumentsCopy` usage/module to avoid shadowing CMake’s built-in `cmake_parse_arguments()`.
2. Replace top-level global path usage (`CMAKE_SOURCE_DIR` / `CMAKE_BINARY_DIR`) with project-scoped paths where needed.

These were causing failures when VOLK is consumed as a dependency, including:
1) `Internal error: SUBBUILD_DIR not set` (from `FetchContent.cmake`)
2) `Cannot find source file .../cmake/Checks/check-rvv-intrinsics.c` in subproject context

#### 1) Parser shadowing
`CMakeParseArgumentsCopy.cmake` defines `CMAKE_PARSE_ARGUMENTS(...)`.  
Because CMake command/function names are case-insensitive, this can shadow built-in `cmake_parse_arguments(...)`.

`FetchContent.cmake` relies on the built-in parser with `PARSE_ARGV`, so shadowing can break argument parsing and lead to:
`Internal error: SUBBUILD_DIR not set`.

#### 2) Global dir variables in subproject context
Some VOLK code used `CMAKE_SOURCE_DIR` / `CMAKE_BINARY_DIR` for internal paths.  
When VOLK is a subproject, those point to the parent project, not VOLK, which breaks path resolution.

## Related

- Fixes gnuradio/volk#815
- Related: https://github.com/gnuradio/gnuradio/issues/7716
- Related: https://github.com/gnuradio/gnuradio/pull/7996


cmake/Modules/VolkPython.cmake
```diff
@@ -196,14 +195,14 @@ function(VOLK_PYTHON_INSTALL)
         add_custom_command(
             DEPENDS ${pysrcfiles}
             OUTPUT ${pycfiles}
-            COMMAND ${PYTHON_EXECUTABLE} ${CMAKE_BINARY_DIR}/python_compile_helper.py
+            COMMAND ${PYTHON_EXECUTABLE} ${PROJECT_BINARY_DIR}/python_compile_helper.py
```
jdemel, line 198: Why are these changes necessary?

jdemel (commented):
This is an interesting use case. A have a question, but besides: LGTM.

cmake/Modules/VolkPython.cmake
xerpi: Good question 🙂
These changes are needed to make VOLK behave correctly when it is included via `FetchContent`, which effectively does:
```cmake
FetchContent_MakeAvailable(volk)
# internally equivalent to:
add_subdirectory(<volk_source> <volk_binary>)
```
In this case, VOLK is no longer the **top-level** project.
`CMAKE_SOURCE_DIR` and `CMAKE_BINARY_DIR` refer to the parent project, not VOLK itself. As a result, helper scripts like `python_compile_helper.py` end up being referenced relative to the parent build tree, which breaks path resolution.

This is different from `ExternalProject`, where VOLK would be configured and built in a completely separate build tree. In that case, `CMAKE_BINARY_DIR` would still refer to VOLK’s own build directory, so this issue wouldn’t appear (distros build VOLK as a standalone project, for example, as a `libvolk` package, so they don't have this issue).

xerpi: EDIT: Actually, this change is not 100% needed , since it's VOLK itself who's generating the `python_compile_helper.py` in the first place.
However, when VOLK is used as a sub-module, it's cleaner to generate this file on the VOLK-specific build dir (`${PROJECT_BINARY_DIR}`)  to not pollute the global/parent build dir (`${CMAKE_BINARY_DIR}`).

jdemel: Thanks for the explanation.

jdemel (approved):

PR #818, opened 2026-03-02
feat: add -T trials flag and profiling improvements to volk_profile

mtibbits:
## Motivation

A single timing trial gives no indication of variance — a noisy machine
or a cold cache can silently skew results. This PR adds support for
multiple timing trials so callers can see mean, stddev, and min, and
make more statistically grounded architecture selections.

### `apps/volk_profile.cc`
- Add `-T <n>` / `--trials <n>` option; defaults to 10 when `-T` is
  given without a value, 1 when the flag is absent entirely
- Refactor `read_results()` to use `std::istringstream` instead of
  manual `char[]` buffer parsing
- Add `json_escape()` helper and apply it to all string fields in JSON
  output (kernel names, arch names, units)
- Emit `time_stddev`, `time_min`, and `trial_times[]` in JSON output

### `apps/volk_option_helpers.cc`
- Add bounds check before consuming the next `argv` token for
  `INT_CALLBACK`, `FLOAT_CALLBACK`, and `STRING_CALLBACK` options;
  emit a warning and skip rather than crash on a missing value
- Simplify `BOOL_CALLBACK` parsing: remove the `try/catch` wrapper,
  handle unrecognized next-token as implicit `true` without consuming it

### `lib/qa_utils.h` / `lib/qa_utils.cc`
- Add `trials` field to `volk_test_params_t` with `trials()` /
  `set_trials()` accessors
- Add `time_stddev`, `time_min`, and `trial_times` fields to
  `volk_test_time_t`
- Run the timing loop `trials` times, accumulate per-trial results,
  compute mean/stddev/min before storing

### `lib/kernel_tests.h`
- Thread `trials` through the per-kernel test parameter structs

## Test plan

- [ ] `volk_profile -R volk_32f_x2_dot_prod_32f` — single trial, baseline
- [ ] `volk_profile -T 10 -R volk_32f_x2_dot_prod_32f` — 10 trials, verify stddev/min appear in JSON
- [ ] `volk_profile -T -R volk_32f_x2_dot_prod_32f` — `-T` without value defaults to 10
- [ ] `volk_profile --help` — verify `-T` appears in help output
- [ ] Existing `volk_profile` unit tests pass

### DIFF
--- a/apps/volk_option_helpers.cc
+++ b/apps/volk_option_helpers.cc
@@ -128,7 +128,7 @@ void option_list::parse(int argc, char** argv)
                 case INT_CALLBACK:
                     try {
                         if (arg_number + 1 >= argc) {
-                            std::cout << "Warning: option '" << argv[arg_number]
+                            std::cerr << "Warning: option '" << argv[arg_number]
                                       << "' expects a numeric value" << std::endl;
                             break;
                         }
@@ -139,7 +139,7 @@ void option_list::parse(int argc, char** argv)
                                 (next_arg[0] == '-' && next_arg[1] >= '0' &&
                                  next_arg[1] <= '9');
                             if (!is_number) {
-                                std::cout << "Warning: option '" << argv[arg_number]
+                                std::cerr << "Warning: option '" << argv[arg_number]
                                           << "' expects a numeric value" << std::endl;
                                 break;
                             }
@@ -147,7 +147,7 @@ void option_list::parse(int argc, char** argv)
                         int_val = atoi(argv[++arg_number]);
                         ((void (*)(int))this_option->callback)(int_val);
                     } catch (std::exception& exc) {
-                        std::cout << "An int option can only receive a number"
+                        std::cerr << "An int option can only receive a number"
                                   << std::endl;
                         throw std::exception();
                     };
@@ -155,7 +155,7 @@ void option_list::parse(int argc, char** argv)
                 case FLOAT_CALLBACK:
                     try {
                         if (arg_number + 1 >= argc) {
-                            std::cout << "Warning: option '" << argv[arg_number]
+                            std::cerr << "Warning: option '" << argv[arg_number]
                                       << "' expects a numeric value" << std::endl;
                             break;
                         }
@@ -167,7 +167,7 @@ void option_list::parse(int argc, char** argv)
                                  next_arg[1] <= '9') ||
                                 (next_arg[0] == '.');
                             if (!is_number) {
-                                std::cout << "Warning: option '" << argv[arg_number]
+                                std::cerr << "Warning: option '" << argv[arg_number]
                                           << "' expects a numeric value" << std::endl;
                                 break;
                             }
@@ -175,7 +175,7 @@ void option_list::parse(int argc, char** argv)
                         double double_val = atof(argv[++arg_number]);
                         ((void (*)(float))this_option->callback)(double_val);
                     } catch (std::exception& exc) {
-                        std::cout << "A float option can only receive a number"
+                        std::cerr << "A float option can only receive a number"
                                   << std::endl;
                         throw std::exception();
                     };
@@ -209,7 +209,7 @@ void option_list::parse(int argc, char** argv)
                 case STRING_CALLBACK:
                     try {
                         if (arg_number + 1 >= argc) {
-                            std::cout << "Warning: option '" << argv[arg_number]
+                            std::cerr << "Warning: option '" << argv[arg_number]
                                       << "' expects a value" << std::endl;
                             break;
                         }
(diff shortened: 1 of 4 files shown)


apps/volk_option_helpers.cc
```diff
@@ -127,6 +127,23 @@ void option_list::parse(int argc, char** argv)
...
+                            break;
+                        }
+                        {
+                            char* next_arg = argv[arg_number + 1];
+                            bool is_number =
+                                (next_arg[0] >= '0' && next_arg[0] <= '9') ||
+                                (next_arg[0] == '-' && next_arg[1] >= '0' &&
+                                 next_arg[1] <= '9');
+                            if (!is_number) {
+                                std::cout << "Warning: option '" << argv[arg_number]
```
marcusmueller: Also not you to introduce that, but error/warning messages should go to `std::cerr`, not `cout`. We might want to consistently fix this in all places here, as a separate PR.

mtibbits:
Apologies that I missed the std::cout usage.  I was focused on having it
not segfault.  I'll have this one modified and back up in a few minutes.

apps/volk_profile.cc
```diff
@@ -324,6 +316,30 @@ void write_results(const std::vector<volk_test_results_t>* results,
...
+                char buf[8];
+                snprintf(buf, sizeof(buf), "\\u%04x", (unsigned char)c);
+                out += buf;
+            } else {
+                out += c;
+            }
+        }
+    }
+    return out;
+}
```
jdemel: Since we work with a very defined set here: Why would we need this?

jdemel: Since we have the benefit of C++ here, fmtlib would be the tool of choice for string manipulation.

lib/qa_utils.cc
```diff
@@ -788,16 +804,30 @@ bool run_volk_tests(volk_func_desc_t desc,
                     std::string puppet_master_name,
                     bool absolute_mode,
                     bool benchmark_mode,
+                    unsigned int trials,
                     const std::vector<float>& float_edge_cases,
                     const std::vector<lv_32fc_t>& complex_edge_cases)
 {
+    // Ensure at least 1 trial to avoid empty-vector crashes in statistics
+    if (trials == 0)
+        trials = 1;
```
jdemel: always curly braces please.

jdemel (commented):
Thank you for your PR.

I can see that multiple different changes are done in this PR. It would be great if you could separate those for a better discussion.

For the feature in the title: How is the trial parameter different from the iter parameter?

mtibbits:
> Thank you for your PR.
> 
> I can see that multiple different changes are done in this PR. It would be great if you could separate those for a better discussion.
> 
> For the feature in the title: How is the trial parameter different from the iter parameter?

The iter parameter changes the input vector length.  The parameter that I've introduced:

 -T < N: number of trials >

Reruns the same kernel N times, where N is the value provided with the '-T' flag.  Each trial is run with the same vector length.

I specifically wanted a variance estimate to make a statistical determination that one implementation is statistically, significantly faster than another (for a given hardware profile / compiler).

jdemel:
> > Thank you for your PR.
> > I can see that multiple different changes are done in this PR. It would be great if you could separate those for a better discussion.
> > For the feature in the title: How is the trial parameter different from the iter parameter?
> 
> The iter parameter changes the input vector length. The parameter that I've introduced:
> 
> -T < N: number of trials >

Please check all `volk_profile` flags/options.

jdemel:
In order to be able to merge this PR, all the tests would need to pass.

mtibbits:
> > > Thank you for your PR.
> > > I can see that multiple different changes are done in this PR. It would be great if you could separate those for a better discussion.
> > > For the feature in the title: How is the trial parameter different from the iter parameter?
> > 
> > 
> > The iter parameter changes the input vector length. The parameter that I've introduced:
> > -T < N: number of trials >
> 
> Please check all `volk_profile` flags/options.

-i / --iter sets the number of iterations within a single timing measurement. Each "iteration" calls the kernel function once, and the total elapsed time is divided by the iteration count to get a per-call time. More iterations → more stable single measurement (amortizes overhead), but you only get one time estimate. Default is 1987 (set in test_params on line 29).

-T / --trials (from PR #818) sets the number of independent timing trials. Each trial runs the full iteration loop separately, producing its own time measurement. With multiple trials you get statistical information: mean, stddev, min, and the full array of per-trial times in JSON output. Default is 1 trial (old behavior), or 10 if -T is given without a value.

In short:

iter = how many kernel calls per measurement (controls precision of a single data point)
trials = how many independent measurements to take (controls ability to assess variance/noise)
With -i 2000 -T 10, each trial runs the kernel 2000 times to get one timing value, and that's repeated 10 times to produce 10 independent timing values from which mean/stddev/min are computed.

apps/volk_profile.cc
mtibbits: Fair point.  I removed this.

lib/qa_utils.cc
mtibbits: Done.

PR #819, opened 2026-03-02
feat: add AVX-512BW/VL/VBMI/VBMI2/VNNI and F16C ISA support

mtibbits:
## Motivation

Several AVX-512 sub-ISAs (`avx512bw`, `avx512vl`, `avx512vbmi`,
`avx512vbmi2`, `avx512vnni`) and `f16c` were absent from `archs.xml`
and `machines.xml`, making it impossible to guard kernel implementations
with `LV_HAVE_AVX512BW` etc. or select them at runtime. Additionally,
the arch bitmask was stored as a plain `int`, which overflows with more
than ~32 simultaneous flags.

### `gen/archs.xml`
- Add `f16c` arch with `-mf16c` / `/arch:AVX` compiler flags
- Add `avx512bw`, `avx512vl`, `avx512vnni`, `avx512vbmi`, `avx512vbmi2`
  with appropriate flags and 64-byte alignment

### `gen/machines.xml`
- Add machine entries for `avx512bw`, `avx512vl`, `avx512vbmi`,
  `avx512vnni`, `avx512vbmi2` with correct cumulative arch chains
- Fix `avx512dq` machine to include `avx512bw` as a prerequisite

### `lib/volk_rank_archs.h` / `lib/volk_rank_archs.c`
- Widen `impl_deps` from `int*` to `uint64_t*` to support >32 ISA flags
- Use `int64_t` for best-value accumulators
- Add guard when `n_impls == 0`
- Improve fallback: search for `"generic"` by name before returning
  index 0; emit more informative stderr messages
- Replace `strncmp(..., 20)` with `strcmp` for exact name matching

### `tmpl/`
- Propagate `uint64_t impl_deps` through all code-generation templates

### `lib/CMakeLists.txt`
- Add `overrule_arch(avx512bw ...)` guard for non-x86 targets

## Test plan

- [ ] Confirm `LV_HAVE_AVX512BW`, `LV_HAVE_AVX512VL`, `LV_HAVE_AVX512VBMI` etc. appear in generated headers on a capable machine
- [ ] `volk_profile` selects an AVX-512BW/VBMI implementation when running on a machine that supports it
- [ ] Build succeeds on ARM (non-x86) with the new `overrule_arch` guard
- [ ] Existing `volk_profile` and correctness tests pass

### DIFF
--- a/gen/machines.xml
+++ b/gen/machines.xml
@@ -73,13 +73,38 @@
 </machine>
 
 <!-- trailing | bar means generate without either for MSVC -->
+<machine name="avx512bw">
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 avx512f avx512bw orc|</archs>
+</machine>
+
+<!-- trailing | bar means generate without either for MSVC -->
 <machine name="avx512cd">
 <archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 avx512f avx512cd orc|</archs>
 </machine>
 
 <!-- trailing | bar means generate without either for MSVC -->
 <machine name="avx512dq">
-<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 avx512f avx512dq orc|</archs>
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 avx512f avx512bw avx512dq orc|</archs>
+</machine>
+
+<!-- trailing | bar means generate without either for MSVC -->
+<machine name="avx512vl">
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 f16c avx512f avx512bw avx512vl orc|</archs>
+</machine>
+
+<!-- trailing | bar means generate without either for MSVC -->
+<machine name="avx512vbmi">
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 f16c avx512f avx512bw avx512vl avx512vbmi orc|</archs>
+</machine>
+
+<!-- trailing | bar means generate without either for MSVC -->
+<machine name="avx512vnni">
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 f16c avx512f avx512bw avx512vl avx512vnni orc|</archs>
+</machine>
+
+<!-- trailing | bar means generate without either for MSVC -->
+<machine name="avx512vbmi2">
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 f16c avx512f avx512bw avx512vl avx512vbmi avx512vbmi2 orc|</archs>
 </machine>
 
 </grammar>
--- a/lib/volk_rank_archs.c
+++ b/lib/volk_rank_archs.c
@@ -8,6 +8,7 @@
  */
 
 
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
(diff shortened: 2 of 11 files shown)


gen/machines.xml
```diff
@@ -72,14 +72,39 @@
...
 <!-- trailing | bar means generate without either for MSVC -->
 <machine name="avx512dq">
-<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 avx512f avx512dq orc|</archs>
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 avx512f avx512bw avx512dq orc|</archs>
+</machine>
+
+<!-- trailing | bar means generate without either for MSVC -->
+<machine name="avx512vl">
+<archs>generic 32|64| mmx| sse sse2 sse3 ssse3 sse4_1 sse4_2 popcount avx fma avx2 f16c avx512f avx512bw avx512vl orc|</archs>
+</machine>
```
jdemel, line 93: Looking at the following table: https://en.wikipedia.org/wiki/AVX-512#CPUs_with_AVX-512

VL, DQ, BW seem to be part of the same release. Thus, it'd make sense to expect either all 3 or none of them. Could you please check?

lib/volk_rank_archs.c
```diff
@@ -20,26 +21,44 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
...
 )
 {
+    if (n_impls == 0) {
+        fprintf(stderr, "Volk error: no implementations available\n");
+        return -1;
+    }
     unsigned int i;
     for (i = 0; i < n_impls; i++) {
-        if (!strncmp(impl_names[i], impl_name, 20)) {
+        if (!strcmp(impl_names[i], impl_name)) {
```
jdemel, line 30: The 20 was on purpose. Please check if this is necessary and how to best update or not update the code.

lib/volk_rank_archs.c
```diff
@@ -20,26 +21,44 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
...
+                impl_name);
+        for (i = 0; i < n_impls; i++) {
+            if (!strcmp(impl_names[i], "generic")) {
+                return i;
+            }
+        }
+    }
+    // neither requested impl nor "generic" found — return first available
+    fprintf(stderr, "Volk warning: no generic impl found, returning index 0\n");
+    return 0;
```
jdemel, line 47: This changes the behavior of the whole function. Unfortunately, we learned that we need to keep the behavior as is. Please make sure your changes do not change the general behavior.

lib/volk_rank_archs.c
```diff
@@ -20,26 +21,44 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
...
-                    const int* impl_deps,     // requirement mask per implementation
-                    const bool* alignment,    // alignment status of each implementation
-                    size_t n_impls,           // number of implementations available
-                    const bool align          // if false, filter aligned implementations
+int volk_rank_archs(const char* kern_name,       // name of the kernel to rank
+                    const char* impl_names[],    // list of implementations by name
+                    const uint64_t* impl_deps,   // requirement mask per implementation
+                    const bool* alignment,       // alignment status of each implementation
+                    size_t n_impls,              // number of implementations available
+                    const bool align             // if false, filter aligned implementations
```
jdemel, line 55: Unfortunately, this might break the API. Could you investigate if and why this is safe or not?
It will definitely break the ABI. This would be a problem.

tmpl/volk.tmpl.c
```diff
@@ -41,6 +41,10 @@ struct volk_machine *get_machine(void)
       }
     }
     machine = max_machine;
+    if (!machine) {
+      fprintf(stderr, "VOLK: no machine found matching CPU capabilities\n");
+      abort();
+    }
```
jdemel, line 47: Why is this needed now?

tmpl/volk.tmpl.c
```diff
@@ -95,6 +103,8 @@ size_t volk_get_alignment(void)
 
 bool volk_is_aligned(const void *ptr)
 {
+    if (__alignment_mask == 0)
+        return false;
```
jdemel, line 107: Please, always use curly braces.
Why do we need this check?

tmpl/volk.tmpl.c
```diff
@@ -176,6 +186,10 @@ static inline void __${kern.name}(${kern.arglist_full})
         get_machine()->${kern.name}_n_impls,
         impl_name
     );
+    if (index < 0) {
+        fprintf(stderr, "VOLK: no implementation found for ${kern.name}\n");
+        return NULL;
+    }
```
jdemel, line 192: What changed that makes this check necessary?

tmpl/volk.tmpl.h
```diff
@@ -63,14 +63,7 @@ VOLK_API bool volk_is_aligned(const void *ptr);
...
 <%
-deprecated_kernels = ('volk_16i_x5_add_quad_16i_x4', 'volk_16i_branch_4_state_8',
-                      'volk_16i_max_star_16i', 'volk_16i_max_star_horizontal_16i',
-                      'volk_16i_permute_and_scalar_add', 'volk_16i_x4_quad_max_star_16i',
-                      'volk_32fc_s32fc_multiply_32fc', 'volk_32fc_s32fc_x2_rotator_32fc',
-                      'volk_32fc_x2_s32fc_multiply_conjugate_add_32fc')
-from platform import system
-if system() == 'Windows':
-    deprecated_kernels = ()
+deprecated_kernels = ()
```
jdemel, line 66: Why is this change necessary?

jdemel (changes requested):
I have a couple of questions.

Generally, could you outline why these changes where made and how these are currently used?

Without an example kernel, it is hard to validate if we need these defines and if they work correctly.

mtibbits:
> I have a couple of questions.
> 
> Generally, could you outline why these changes where made and how these are currently used?
> 
> Without an example kernel, it is hard to validate if we need these defines and if they work correctly.

Understood.  I'll submit an exemplar kernel shortly.  Bottom line, I've found several bugs, and rewritten several kernels.  I did not want to upload a ton of PRs _or_ a ton of changes in one PR.  So I was trying to provide meaningful chunks.

gen/machines.xml
mtibbits: Fair.  So if you want we could collapse vl, bw, and dq.

PR #820, opened 2026-03-02
fix: buffer and integer safety fixes in volk_malloc and volk_prefs

mtibbits:
## Summary

Two small safety fixes addressing potential buffer overruns and integer
overflow in core library files.

### `lib/volk_malloc.c`
- When rounding the requested allocation size up to the next alignment
  boundary, the addition of the padding value could silently wrap a
  `size_t` on 32-bit targets if the requested size was near `SIZE_MAX`.
  Added an explicit overflow check; returns `NULL` if overflow would occur.

### `lib/volk_prefs.c`
- Config file path construction used `strncpy` + `strcat` pairs, which
  can produce unterminated strings if the first call fills the buffer
  exactly, and can overflow on the `strcat`. Replaced all four
  occurrences with a single `snprintf` call per path.

## Test plan

- [ ] `volk_profile` correctly reads/writes volk config (existing behaviour unchanged)
- [ ] Existing unit tests pass

### DIFF
--- a/lib/volk_malloc.c
+++ b/lib/volk_malloc.c
@@ -7,6 +7,7 @@
  * SPDX-License-Identifier: LGPL-3.0-or-later
  */
 
+#include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -45,7 +46,11 @@ void* volk_malloc(size_t size, size_t alignment)
     // requested for correct alignment. Any allocation size change here will in general
     // not impact the end result since initial size alignment is required either way.
     if (size % alignment) {
-        size += alignment - (size % alignment);
+        size_t padding = alignment - (size % alignment);
+        if (padding > SIZE_MAX - size) {
+            return NULL; // overflow guard
+        }
+        size += padding;
     }
 #if HAVE_POSIX_MEMALIGN
     // quoting posix_memalign() man page:
--- a/lib/volk_prefs.c
+++ b/lib/volk_prefs.c
@@ -20,19 +20,20 @@
 #endif
 #include <volk/volk_prefs.h>
 
+static const size_t CONFIG_PATH_MAX = 4096;
+
 void volk_get_config_path(char* path, bool read)
 {
     if (!path)
         return;
     const char* suffix = "/.volk/volk_config";
     const char* suffix2 = "/volk/volk_config"; // non-hidden
-    const size_t path_max = 512;
     char* home = NULL;
 
     // allows config redirection via env variable
     home = getenv("VOLK_CONFIGPATH");
     if (home != NULL) {
-        snprintf(path, path_max, "%s%s", home, suffix2);
+        snprintf(path, CONFIG_PATH_MAX, "%s%s", home, suffix2);
         if (!read || access(path, F_OK) != -1) {
             return;
         }
@@ -41,7 +42,7 @@ void volk_get_config_path(char* path, bool read)
     // check for user-local config file
     home = getenv("HOME");
     if (home != NULL) {
-        snprintf(path, path_max, "%s%s", home, suffix);
+        snprintf(path, CONFIG_PATH_MAX, "%s%s", home, suffix);
         if (!read || (access(path, F_OK) != -1)) {
             return;
         }
@@ -50,7 +51,7 @@ void volk_get_config_path(char* path, bool read)
     // check for config file in APPDATA (Windows)
     home = getenv("APPDATA");
     if (home != NULL) {
-        snprintf(path, path_max, "%s%s", home, suffix);
+        snprintf(path, CONFIG_PATH_MAX, "%s%s", home, suffix);
         if (!read || (access(path, F_OK) != -1)) {
             return;
         }
@@ -58,7 +59,7 @@ void volk_get_config_path(char* path, bool read)
 
     // check for system-wide config file
     if (access("/etc/volk/volk_config", F_OK) != -1) {
-        snprintf(path, path_max, "%s%s", "/etc", suffix2);
+        snprintf(path, CONFIG_PATH_MAX, "%s%s", "/etc", suffix2);
         if (!read || (access(path, F_OK) != -1)) {
             return;
         }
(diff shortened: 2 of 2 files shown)


lib/volk_malloc.c
```diff
@@ -45,7 +46,11 @@ void* volk_malloc(size_t size, size_t alignment)
     // requested for correct alignment. Any allocation size change here will in general
     // not impact the end result since initial size alignment is required either way.
     if (size % alignment) {
-        size += alignment - (size % alignment);
+        size_t padding = alignment - (size % alignment);
+        if (padding > SIZE_MAX - size) {
+            return NULL; // overflow guard
+        }
+        size += padding;
```
marcusmueller, line 53: generally, right thing to do, but a high power of two will always inherently be a multiple of `alignment`, wouldn't it?

marcusmueller:
hm if this was done with claude code completely, do you have the legal authority over the code to submit it to this project?

lib/volk_prefs.c
```diff
@@ -26,13 +26,13 @@ void volk_get_config_path(char* path, bool read)
         return;
     const char* suffix = "/.volk/volk_config";
     const char* suffix2 = "/volk/volk_config"; // non-hidden
+    const size_t path_max = 512;
```
marcusmueller: constants in CAPITAL_LETTERS, PLS.

marcusmueller (changes requested):

lib/volk_prefs.c
marcusmueller: This function should probably be changed in signature to accept the length of the char buffer passed, and then calculate the `path_max` from that.

However, not possible in patch release, and this deserves to be patched.

Also, 512 is far too small for Linux, MacOS and ×BSD. The maximum length of a path on any given system is notoriously hard to guess, so for next major release, we could consider dropping this C function altogether, have it be implemented anew in C++ (by the way, fundamentally different way – the paths we look into are all wrong), and give it a new `extern C` wrapper that,

- wraps a preferred C++ API that just returns a `std::filesystem::path` object,
- has not only a buffer, but also a length parameter,
- instead of `void`, returns an error code that, if positive indicates how many bytes where written (copied from a a C++ `std::fs::path`) into that buffer, so that the consuming application can check itself that all it wanted, it got, and if it didn't, can decide to try again with a larger buffer.

Gah, I hate handling strings in C.

mtibbits:
Please note up front that I am not a lawyer.  But my understanding is that,
based on the terms of use, the software generated belongs to me.  However,
it does not remove the risk that Claude might have reproduced copyrighted
material.  But I do not believe that to be the case in these submissions.
For reference, I have a paid subscription to Anthropic, Claude Max, the
terms of use are here:

https://www.anthropic.com/legal/consumer-terms

The important paragraph is:

Rights and Responsibilities. You are responsible for all Inputs you submit
to our Services and all Actions. By submitting Inputs to our Services, you
represent and warrant that you have all rights, licenses, and permissions
that are necessary for us to process the Inputs under our Terms and to
provide the Services to you, including for example, to integrate with
third-party services, to share Materials with others at your direction, and
to take Actions. You also represent and warrant that your submitting Inputs
to us or directing Claude to take Actions will not violate our Terms,
our Acceptable
Use Policy <https://anthropic.com/aup>, or any laws or regulations
applicable to those Inputs or Actions. As between you and Anthropic, and to
the extent permitted by applicable law, you retain any right, title, and
interest that you have in the Inputs you submit. Subject to your compliance
with our Terms, we assign to you all of our right, title, and interest—if
any—in Outputs.

lib/volk_prefs.c
```diff
@@ -51,17 +51,15 @@ void volk_get_config_path(char* path, bool read)
...
+        snprintf(path, CONFIG_PATH_MAX, "%s%s", home, suffix);
         if (!read || (access(path, F_OK) != -1)) {
             return;
         }
     }
 
     // check for system-wide config file
     if (access("/etc/volk/volk_config", F_OK) != -1) {
-        strncpy(path, "/etc", 512);
-        strcat(path, suffix2);
```
jdemel, line 64: This could be simplified for brevity.

lib/volk_prefs.c
```diff
@@ -96,7 +94,7 @@ size_t volk_load_preferences(volk_arch_pref_t** prefs_res)
         }
         prefs = (volk_arch_pref_t*)new_prefs;
         volk_arch_pref_t* p = prefs + n_arch_prefs;
-        if (sscanf(line, "%s %s %s", p->name, p->impl_a, p->impl_u) == 3 &&
+        if (sscanf(line, "%127s %127s %127s", p->name, p->impl_a, p->impl_u) == 3 &&
```
jdemel, line 97: Why is this change necessary?

jdemel (commented):
Thank you for your PR. There are a couple of comments and questions that would need to be discussed. Thanks.

lib/volk_malloc.c
mtibbits: This is a check for overflow.

lib/volk_prefs.c
mtibbits: This was a first pass to make it cleaner and avoid a non-null terminated string.

lib/volk_prefs.c
mtibbits: The change from %s to %127s in sscanf prevents a buffer overflow.

volk_arch_pref_t has fixed-size fields — looking at the struct, each of p->name, p->impl_a, and p->impl_u is a char[128]. With plain %s, sscanf writes until it hits whitespace with no length limit. If a config file line contains a token longer than 127 characters, sscanf would write past the end of the field, corrupting adjacent memory (classic stack/heap buffer overflow).

The %127s width specifier tells sscanf to read at most 127 characters (leaving room for the null terminator), so the write is bounded to exactly the size of each field.

This matters here because volk_prefs.c reads from a user-controlled file (~/.volk/volk_config), so the input isn't trusted. Without the width limit, a malformed config file could trigger undefined behavior.

Issue #821, opened 2026-03-03
volk_profile do not filter out failed kernels


viktor3308:
In the output of volk_profile, you may sometimes notice tolerance test failure messages for certain architectures, but these architectures can still end up in volk_config if they are fast enough.

In the [source code](https://github.com/gnuradio/volk/blob/511fff053e190316919733fb148b70946f139a9e/lib/qa_utils.cc#L1386), there is a collection of test failure flags (arch_results), which are later used when filtering architectures ([unaligned](https://github.com/gnuradio/volk/blob/511fff053e190316919733fb148b70946f139a9e/lib/qa_utils.cc#L1398) and [aligned](https://github.com/gnuradio/volk/blob/511fff053e190316919733fb148b70946f139a9e/lib/qa_utils.cc#L1402)). However, the flag is taken only from the last iteration of the argument comparison loop. It seems that this flag should account for failures in comparing any of the arguments, not just the last one.

This looks like a bug in the code.

PR #822, opened 2026-03-03
Filter failed kernels

viktor3308:
A fix for issue #821, if it is indeed a bug (which it appears to be).

### DIFF
--- a/lib/qa_utils.cc
+++ b/lib/qa_utils.cc
@@ -1383,7 +1383,8 @@ bool run_volk_tests(volk_func_desc_t desc,
                 }
             }
         }
-        arch_results.push_back(!fail);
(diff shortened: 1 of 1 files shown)


lib/qa_utils.cc
```diff
@@ -1383,7 +1383,8 @@ bool run_volk_tests(volk_func_desc_t desc,
                 }
             }
         }
-        arch_results.push_back(!fail);
+        volk_test_time_t* result = &results->back().results[arch_list[i]];
+        arch_results.push_back(result->pass);
```
jdemel, line 1387: We're still inside the loop for one architecture. I don't see how this would fail.
But: I'd argue the actual problem in this loop is the complexity. It would make more sense to refactor it and create a simpler problem.

PR #823, opened 2026-03-03
fix: remove unused variables and dead assignments in 6 kernels

mtibbits:
## Summary

Remove variables that are written but never subsequently read, and dead pointer increments at the end of loop bodies. These were identified by static analysis and manual review. No behavioral change — the generated code is semantically identical.

| Kernel | Implementation | Change |
|---|---|---|
| `volk_16i_max_star_horizontal_16i` | `a_ssse3` | Remove dead `p_src0` and `p_target` increments at end of SIMD loop body |
| `volk_32f_accumulator_s32f` | `a/u_avx512f`, `a/u_avx`, `a/u_sse` | Replace `aVal = _mm*_setzero_ps()` initializer with bare declaration; always overwritten by load before use |
| `volk_32f_exp_32f` | `a/u_sse2` | Remove `tmp = _mm_setzero_ps()` inside loop body; `tmp` is fully overwritten before it is read |
| `volk_32fc_accumulator_s32fc` | `a/u_avx512f`, `a/u_avx`, `a/u_sse` | Scope `aVal` as a loop-local variable and remove redundant `setzero` initializer |
| `volk_32fc_index_min_16u` | `a_sse3` | Remove final `sq_dist` update after last element comparison; value not read again |
| `volk_32fc_index_min_32u` | `a_sse3` | Same as above |

## Test plan

- [ ] `volk_profile -R volk_16i_max_star_horizontal_16i` — correctness check passes
- [ ] `volk_profile -R volk_32f_accumulator_s32f` — correctness check passes
- [ ] `volk_profile -R volk_32f_exp_32f` — correctness check passes
- [ ] `volk_profile -R volk_32fc_accumulator_s32fc` — correctness check passes
- [ ] `volk_profile -R volk_32fc_index_min_16u` — correctness check passes
- [ ] `volk_profile -R volk_32fc_index_min_32u` — correctness check passes

### DIFF
--- a/kernels/volk/volk_16i_max_star_horizontal_16i.h
+++ b/kernels/volk/volk_16i_max_star_horizontal_16i.h
@@ -133,7 +133,6 @@ static inline void volk_16i_max_star_horizontal_16i_a_ssse3(int16_t* target,
         xmm0 = _mm_load_si128(p_src0);
 
         xmm2 = _mm_xor_si128(xmm2, xmm2);
-        p_src0 += 1;
 
         xmm3 = _mm_hsub_epi16(xmm0, xmm1);
         xmm2 = _mm_cmpgt_epi16(xmm2, xmm3);
@@ -145,8 +144,6 @@ static inline void volk_16i_max_star_horizontal_16i_a_ssse3(int16_t* target,
         xmm0 = _mm_shuffle_epi8(xmm0, xmm3);
 
         _mm_storel_pd((double*)p_target, bit128_p(&xmm0)->double_vec);
-
-        p_target = (__m128i*)((int8_t*)p_target + 8);
     }
 
     for (i = (bound << 4) + (intermediate << 3);
--- a/kernels/volk/volk_32f_accumulator_s32f.h
+++ b/kernels/volk/volk_32f_accumulator_s32f.h
@@ -67,7 +67,7 @@ static inline void volk_32f_accumulator_s32f_a_avx512f(float* result,
     const float* aPtr = inputBuffer;
 
     __m512 accumulator = _mm512_setzero_ps();
-    __m512 aVal = _mm512_setzero_ps();
+    __m512 aVal;
 
     for (; number < sixteenthPoints; number++) {
         aVal = _mm512_load_ps(aPtr);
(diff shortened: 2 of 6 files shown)


kernels/volk/volk_32f_accumulator_s32f.h
```diff
@@ -67,7 +67,7 @@ static inline void volk_32f_accumulator_s32f_a_avx512f(float* result,
     const float* aPtr = inputBuffer;
 
     __m512 accumulator = _mm512_setzero_ps();
-    __m512 aVal = _mm512_setzero_ps();
+    __m512 aVal;
```
jdemel, line 71: While correct, I suggest to clean up the function for brevity, e.g. make variable declarations more local. Have local loop counter variables.

kernels/volk/volk_16i_max_star_horizontal_16i.h
```diff
@@ -145,8 +144,6 @@ static inline void volk_16i_max_star_horizontal_16i_a_ssse3(int16_t* target,
         xmm0 = _mm_shuffle_epi8(xmm0, xmm3);
 
         _mm_storel_pd((double*)p_target, bit128_p(&xmm0)->double_vec);
-
-        p_target = (__m128i*)((int8_t*)p_target + 8);
```
jdemel, line 149: This kernel is deprecated because we don't know about any consumers. Also, we don't have tests. I'd like to keep it as is for the time being.

jdemel (changes requested):
Thank you for your PR. I have a few comments that could improve this PR.

kernels/volk/volk_32f_accumulator_s32f.h
mtibbits: I agree.  As noted in the other PR, #819, I simply wanted to make incremental changes that were easier to review.  On one of my next passes, I was going to submit the numerous scope reductions that I found across all of the kernels.

kernels/volk/volk_16i_max_star_horizontal_16i.h
mtibbits: I can submit the test I wrote as a part of this check in or a later PR.  Which would you prefer?

jdemel: One PR should include the change, the test and the clean up. For reviews, it makes most sense to focus on one code section.

PR #856, opened 2026-04-03
Ranking: Fail in a sensible manner, print useful info

### LINKED ISSUE #516: volk_get_index broken, stuck in infinite loop

marcusmueller:
Closes #516

### DIFF
--- a/lib/volk_rank_archs.c
+++ b/lib/volk_rank_archs.c
@@ -21,14 +21,36 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
 )
 {
     unsigned int i;
+
+    if (n_impls == 0) {
+        // uuuh so we're picking an implementation out of 0 available ones?
+        fprintf(stderr,
+                "Asked to return implementation for \"%.20s\", but 0 implementations "
+                "given: Bailing.",
+                impl_name);
+        return -1;
+    }
+
+    // Find the implementation whose name matches (in the first 20 characters)
     for (i = 0; i < n_impls; i++) {
         if (!strncmp(impl_names[i], impl_name, 20)) {
             return i;
         }
     }
+
     // TODO return -1;
     // something terrible should happen here
-    fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
+    if (!strcmp("generic", impl_name)) {
+        // we're in the error case, and we've already tried to look for the generic
+        // implementation, and failed: There's no good way to proceed here.
+        fprintf(stderr,
+                "Volk warning: tried to find \"generic\" impl, but failed. "
+                "Returning index 0, just to give any working implementation.\n");
+        return 0;
+    }
+    fprintf(stderr,
+            "Volk warning: no arch \"%.20s\" found, returning \"generic\" impl\n",
+            impl_name);
     return volk_get_index(impl_names, n_impls, "generic"); // but we'll fake it for now
 }
 
@@ -40,6 +62,7 @@ int volk_rank_archs(const char* kern_name,    // name of the kernel to rank
                     const bool align          // if false, filter aligned implementations
 )
 {
+    fprintf(stderr, "Ranking %s\n", kern_name);
     size_t i;
     static volk_arch_pref_t* volk_arch_prefs;
     static size_t n_arch_prefs = 0;
--- a/tmpl/volk.tmpl.c
+++ b/tmpl/volk.tmpl.c
@@ -136,8 +136,17 @@ static inline void __init_${kern.name}(void)
     const int *impl_deps = get_machine()->${kern.name}_impl_deps;
     const bool *alignment = get_machine()->${kern.name}_impl_alignment;
     const size_t n_impls = get_machine()->${kern.name}_n_impls;
-    const size_t index_a = volk_rank_archs(name, impl_names, impl_deps, alignment, n_impls, true/*aligned*/);
-    const size_t index_u = volk_rank_archs(name, impl_names, impl_deps, alignment, n_impls, false/*unaligned*/);
+
+    // find the highest-ranked implementation for a given kernel. Will
+    // (warningly) fall back to "generic" if the specified name couldn't be found,
+    // and will properly fail with -1 if there's no implementations at all.
+    const int index_a = volk_rank_archs(
+        name, impl_names, impl_deps, alignment, n_impls, true /*aligned*/);
+    const int index_u = volk_rank_archs(
+        name, impl_names, impl_deps, alignment, n_impls, false /*unaligned*/);
+    assert(index_a >= 0);
+    assert(index_u >= 0);
+
     ${kern.name}_a = get_machine()->${kern.name}_impls[index_a];
     ${kern.name}_u = get_machine()->${kern.name}_impls[index_u];
 


lib/volk_rank_archs.c
```diff
@@ -21,14 +21,36 @@ int volk_get_index(const char* impl_names[], // list of implementations by name
...
     // TODO return -1;
     // something terrible should happen here
-    fprintf(stderr, "Volk warning: no arch found, returning generic impl\n");
+    if (!strcmp("generic", impl_name)) {
+        // we're in the error case, and we've already tried to look for the generic
+        // implementation, and failed: There's no good way to proceed here.
+        fprintf(stderr,
+                "Volk warning: tried to find \"generic\" impl, but failed. "
+                "Returning index 0, just to give any working implementation.\n");
+        return 0;
```
jdemel, line 49: Looking at the previous `return -1` for errors, wouldn't you want the same value here too? Also, since `-1` is an error condition, a quick `#define ERROR_IMPL_NOT_FOUND -1` would probably help long term =D

jdemel (commented):
Thanks! One little comment but besides: LGTM.

lib/volk_rank_archs.c
marcusmueller: no, I don't want that. (We've never returned -1 before, the case where n_impls == 0 is something where there's absolutely no way to continue. So far we've just been happily falling back to things. If we break that behaviour, and I agree we absolutely should, then we need to mark this as a relevant API break including versioning and everything. So far, this can all end up in a patch release.)

marcusmueller:
@jdemel  I think we should more broadly have a design roundtable on how we'd like our dispatching API to look like. I'm tempted to think we should probably should think about our endless kernel .h files, our repetitive and historically very bugprone "and now we also deal with the final n_points%(SIMD multiple) values", and make use of the fact that we could, in fact, let kernels return the number of items processed.

This might call for a mightier interface than our current "the symbol you're using is bent towards the generally optimal implementation at ldload time".

lib/volk_rank_archs.c
jdemel: The previous behavior would be: Find the correct impl and return the index, that's preserved. But now, the error handling changes:
1. If there's no impl -> return -1, which is a good improvement.
2. return 0, for whatever impl if we couldn't find one and we weren't looking for the generic impl.
3. recursive call to `volk_get_index` with `"generic"` is never reached anymore. Either, we returned the correct index for generic already, or we returned 0, or we are back to the infinite loop case that the PR is trying to fix.

unless I just misunderstood the code, it looks like this would need some more attention.

tmpl/volk.tmpl.c
```diff
@@ -136,8 +136,17 @@ static inline void __init_${kern.name}(void)
...
+
+    // find the highest-ranked implementation for a given kernel. Will
+    // (warningly) fall back to "generic" if the specified name couldn't be found,
+    // and will properly fail with -1 if there's no implementations at all.
+    const int index_a = volk_rank_archs(
+        name, impl_names, impl_deps, alignment, n_impls, true /*aligned*/);
+    const int index_u = volk_rank_archs(
+        name, impl_names, impl_deps, alignment, n_impls, false /*unaligned*/);
+    assert(index_a >= 0);
+    assert(index_u >= 0);
```
jdemel, line 148: Assert should not be present in a release compile anymore. Is this check still intentional?

jdemel:
These are all very valid points. I'd really like to improve in every area. My understanding of your suggestions is that we'd design VOLK v4 though because this is a major API breaking change. We can discuss this but I argue that the comments to a PR might not be the right place.

> @jdemel I think we should more broadly have a design roundtable on how we'd like our dispatching API to look like. I'm tempted to think we should probably should think about our endless kernel .h files, our repetitive and historically very bugprone "and now we also deal with the final n_points%(SIMD multiple) values", and make use of the fact that we could, in fact, let kernels return the number of items processed.

So these are multiple bullets:
- "endless kernel header files", would you like to have headers for every implementation? The current library design wouldn't allow for compilation units here.
- tail handling: yes improvements would be great. If we start to discuss this, we might want to rephrase this to prologue/epilogue handling though.
- Return number of processed samples: This would be a very different API, at least potentially.

> This might call for a mightier interface than our current "the symbol you're using is bent towards the generally optimal implementation at ldload time".

I would assume you don't want to re-check the implementation again and again during runtime. I would expect that VOLK continues to run on the same CPU (or at the very least the same kind of CPU if you really migrate between servers) but migrating to an entirely different CPU would also imply a re-launch of the application and then, we can re-load the optimal implementations.

lib/volk_rank_archs.c
marcusmueller: no, it's

1. if no impl => -1
2. if desired impl not found, return index of "generic"
3. if "generic" isn't found, but was asked for, return 0


The 1-depth recursion to find "generic" in case 2. still exists

tmpl/volk.tmpl.c
marcusmueller: it's an intentional check in conformance with the two asserts four lines down from there ;) 

But I'd agree, we *should* be doing better. Any recommendations?

jdemel: I see, it was just past the review window =D

jdemel (approved):
LGTM. The discussion raised some general concerns that require further considerations. They're out of scope for this PR. Thanks :)

PR #857, opened 2026-04-13
cmake: Add `ENABLE_UTILITY_APPS` to avoid building volk_profile and avoid fmt dependency

xerpi:
When cross-compiling/targeting bare-metal, we don't want to build the `volk_profiler` host app, which also allows us to avoid the `fmt` dependency.
Add an `ENABLE_APPS` CMake option (defaults to `ON`) so that users can disable building `volk_profiler` and pulling `fmt`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,6 +50,7 @@ endif()
 list(INSERT CMAKE_MODULE_PATH 0 ${CMAKE_CURRENT_SOURCE_DIR}/cmake/Modules
 ) #location for custom "Modules"
 
+include(CMakeDependentOption)
 include(VolkBuildTypes)
 #select the release build type by default to get optimization flags
 if(NOT CMAKE_BUILD_TYPE)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -347,10 +324,49 @@ else()
...
+            message(
+                STATUS "Static build: using FetchContent to build fmt as static library ...")
+        else()
+            message(STATUS "fmt package not found. Using FetchContent to download ...")
+        endif()
+        include(FetchContent)
+        FetchContent_Declare(
+            fmt
+            GIT_REPOSITORY https://github.com/fmtlib/fmt.git
+            GIT_TAG 10.2.1
```
jdemel: could you update to the latest tag? that should be 12.1.0

CMakeLists.txt
```diff
@@ -347,10 +324,49 @@ else()
 endif()
 message(STATUS "  Modify using: -DENABLE_TESTING=ON/OFF")
 
+########################################################################
+# Host-side utility apps
```
jdemel: I guess the challenge is to express the "host-side" concept. Would it make sense to call them just "utility apps". I assume you want to describe the fact that these apps rely on an OS being present instead of a bare-metal situation.

jdemel (commented):
Thanks! Only a very few minor comments from my side.

CMakeLists.txt
xerpi: I've changed it to just "Utility apps" and the CMake option to `ENABLE_UTILITY_APPS`

jdemel (approved):
Thanks. That looks good to me.

PR #858, opened 2026-04-14
cmake: Add `VOLK_STATIC_DISPATCH` for compile time machine selection

xerpi:
When `VOLK_STATIC_DISPATCH` is set to a machine name (e.g. `neonv8`, `avx2_64_mmx_orc`), the build generates a header only static dispatch layer that maps generic kernel names directly to the best implementation via `#define` and `static inline` wrappers. No runtime CPU detection, no function pointer indirection, no filesystem access.

This is useful for baremetal/embedded targets where the CPU is known at compile time and the runtime dispatch infrastructure (`cpu_features`, `volk_prefs`, `volk_rank_archs`) is not available or desirable.

The generated `volk_dispatch.h` either contains:
- Static dispatch:  `LV_HAVE_*` defines, kernel header includes, `#define` aliases and `static inline` dispatchers.
- Dynamic dispatch: extern function pointer declarations (unchanged).

The common parts (`#include`s, `VOLK_OR_PTR`) live in a static `volk.h` that includes the generated `volk_dispatch.h`.

When static dispatch is active, `ENABLE_APPS`, `ENABLE_TESTING`, `ENABLE_PROFILING` and `ENABLE_MODTOOL` are automatically disabled. `cpu_features`, `fmt`, `ORC` and `dlfcn` dependencies are all skipped.

For example, compiling with `-DVOLK_STATIC_DISPATCH="avx2_64_mmx_orc"` produces a `libvolk` with just `volk_free` and `volk_malloc` VOLK-exported symbols, and a `volk_dispatch.h` header (included by `volk.h`) that looks like:
```c
//! Returns the name of the machine this instance will use
static inline const char *volk_get_machine(void)
{
    return "avx2_64_mmx_orc";
}

//! Get the machine alignment in bytes
static inline size_t volk_get_alignment(void)
{
    return 32;
}

//! Is the pointer on a machine alignment boundary?
static inline bool volk_is_aligned(const void *ptr)
{
    return ((intptr_t)(ptr) & (intptr_t)31) == 0;
}

#define LV_HAVE_GENERIC 1
#define LV_HAVE_64 1
#define LV_HAVE_MMX 1
#define LV_HAVE_SSE 1
#define LV_HAVE_SSE2 1
#define LV_HAVE_SSE3 1
#define LV_HAVE_SSSE3 1
#define LV_HAVE_SSE4_1 1
#define LV_HAVE_SSE4_2 1
#define LV_HAVE_POPCOUNT 1
#define LV_HAVE_AVX 1
#define LV_HAVE_FMA 1
#define LV_HAVE_AVX2 1
#define LV_HAVE_ORC 1

#include <volk/volk_16i_32fc_dot_prod_32fc.h>
#include <volk/volk_16i_branch_4_state_8.h>
#include <volk/volk_16i_convert_8i.h>
....

#define volk_16i_32fc_dot_prod_32fc_a volk_16i_32fc_dot_prod_32fc_a_avx2_fma
#define volk_16i_32fc_dot_prod_32fc_u volk_16i_32fc_dot_prod_32fc_u_avx2_fma

static inline void volk_16i_32fc_dot_prod_32fc(lv_32fc_t*  result, const short*  input, const lv_32fc_t*  taps, unsigned int  num_points)
{
    if (volk_is_aligned(VOLK_OR_PTR(result,VOLK_OR_PTR(input,VOLK_OR_PTR(taps,0)))))
        volk_16i_32fc_dot_prod_32fc_a(result, input, taps, num_points);
    else
        volk_16i_32fc_dot_prod_32fc_u(result, input, taps, num_points);
}

#define volk_16i_branch_4_state_8_a volk_16i_branch_4_state_8_a_ssse3
#define volk_16i_branch_4_state_8_u volk_16i_branch_4_state_8_generic

static inline void volk_16i_branch_4_state_8(short*  target, short*  src0, char**  permuters, short*  cntl2, short*  cntl3, short*  scalars)
{
    if (volk_is_aligned(VOLK_OR_PTR(target,VOLK_OR_PTR(src0,VOLK_OR_PTR(permuters,VOLK_OR_PTR(cntl2,VOLK_OR_PTR(cntl3,VOLK_OR_PTR(scalars,0))))))))
        volk_16i_branch_4_state_8_a(target, src0, permuters, cntl2, cntl3, scalars);
    else
        volk_16i_branch_4_state_8_u(target, src0, permuters, cntl2, cntl3, scalars);
}

#define volk_16i_convert_8i_a volk_16i_convert_8i_a_avx2
#define volk_16i_convert_8i_u volk_16i_convert_8i_u_avx2

static inline void volk_16i_convert_8i(int8_t*  outputVector, const int16_t*  inputVector, unsigned int  num_points)
{
    if (volk_is_aligned(VOLK_OR_PTR(outputVector,VOLK_OR_PTR(inputVector,0))))
        volk_16i_convert_8i_a(outputVector, inputVector, num_points);
    else
        volk_16i_convert_8i_u(outputVector, inputVector, num_points);
}
....
```

With `VOLK_STATIC_DISPATCH="neonv8"` :
```c
#define volk_16i_32fc_dot_prod_32fc_a volk_16i_32fc_dot_prod_32fc_neonv8
#define volk_16i_32fc_dot_prod_32fc_u volk_16i_32fc_dot_prod_32fc_neonv8

static inline void volk_16i_32fc_dot_prod_32fc(lv_32fc_t*  result, const short*  input, const lv_32fc_t*  taps, unsigned int  num_points)
{
    if (volk_is_aligned(VOLK_OR_PTR(result,VOLK_OR_PTR(input,VOLK_OR_PTR(taps,0)))))
        volk_16i_32fc_dot_prod_32fc_a(result, input, taps, num_points);
    else
        volk_16i_32fc_dot_prod_32fc_u(result, input, taps, num_points);
}
```


xerpi:
Only [Build on ubuntu22.04 armv7 g++](https://github.com/gnuradio/volk/actions/runs/24378968661/job/71198225163?pr=858#logs) failed:
```
  99% tests passed, 1 tests failed out of 196
  
  Total Test time (real) = 229.54 sec
  
  The following tests FAILED:
  	 43 - qa_volk_32f_log2_32f (Failed)
```
Could be a flaky test.

EDIT: After running again now it's passing.

xerpi:
Again, `qa_volk_32f_log2_32f` failed...
Also, `kernels/volk/volk_32f_expfast_32f.h` was polluting the global `#define`-space, so I added some `#undef`s.

kernels/volk/volk_32f_expfast_32f.h
```diff
@@ -397,4 +397,9 @@ volk_32f_expfast_32f_rvv(float* bVector, const float* aVector, unsigned int num_
 }
 #endif /*LV_HAVE_RVV*/
 
+#undef Mln2
+#undef A
+#undef B
+#undef C
```
jdemel: Where is this coming from? Could that change have ended up here by accident?

CMakeLists.txt
```diff
@@ -154,12 +176,11 @@ else()
     message(STATUS "Building Volk without cpu_features")
 endif()
 
-# fmt - required for qa_utils table printing
-# For static builds, always use FetchContent to ensure we get a static library
-if(NOT ENABLE_STATIC_LIBS)
+# fmt - required for qa_utils table printing (apps and tests only)
+if(NOT VOLK_STATIC_DISPATCH AND NOT ENABLE_STATIC_LIBS)
     find_package(fmt QUIET)
 endif()
```
jdemel: I have feeling that it'd make most sense to have all the fmtlib related changes in your other PR #857 . I suggest to move the whole `find_package` code to one place where we also decide if we want to do fetch content or rely on system packages.

CMakeLists.txt
```diff
@@ -247,27 +268,35 @@ install(
...
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/volk/volk_malloc.h
+    ${CMAKE_CURRENT_BINARY_DIR}/include/volk/volk_dispatch.h
+    ${CMAKE_CURRENT_BINARY_DIR}/include/volk/volk_config_fixed.h
+    ${CMAKE_CURRENT_BINARY_DIR}/include/volk/volk_typedefs.h
+    ${CMAKE_CURRENT_BINARY_DIR}/include/volk/volk_version.h)
+
+if(NOT VOLK_STATIC_DISPATCH)
+    list(APPEND VOLK_INSTALL_HEADERS
+        ${CMAKE_CURRENT_SOURCE_DIR}/include/volk/volk_prefs.h
+        ${CMAKE_CURRENT_SOURCE_DIR}/include/volk/constants.h
```
jdemel: Why do the constants need to be excluded?

jdemel (commented):
So far looks good. I couldn't review it completely yet, though.

kernels/volk/volk_32f_expfast_32f.h
xerpi: VOLK is `#define`ing `Mln2`, `A`, `B`, `C`, which is polluting the global define-space, and my code had variables named A and B so I was getting compiler errors.
Unless they are part of the public API, I think it's better to `#undef` them.
I left a note here: https://github.com/gnuradio/volk/pull/858#issuecomment-4242727843

CMakeLists.txt
xerpi: `constants.h` contains many functions that seem targeted for "dynamic dispatch" / running with a proper OS beneath:
```c
VOLK_API const char* volk_prefix();
VOLK_API const char* volk_version();
VOLK_API const char* volk_c_compiler();
VOLK_API const char* volk_compiler_flags();
VOLK_API const char* volk_available_machines();
```
I think some of them would still make sense in baremetal/embedded systems (all except `volk_prefix` and `volk_available_machines`).

jdemel: I checked what `constants.c` expands to. Only `volk_prefix` does anything dynamically. Thinking of it: Would it make sense to set this to something static for your case? The other cases don't hurt. A static string might not yield information in your case but it doesn't hurt either, does it?

CMakeLists.txt
jdemel: I just merged the other PR. Resolving this thread.

jdemel:
> Only [Build on ubuntu22.04 armv7 g++](https://github.com/gnuradio/volk/actions/runs/24378968661/job/71198225163?pr=858#logs) failed:
> 
> ```
>   99% tests passed, 1 tests failed out of 196
>   
>   Total Test time (real) = 229.54 sec
>   
>   The following tests FAILED:
>   	 43 - qa_volk_32f_log2_32f (Failed)
> ```
> 
> Could be a flaky test.
> 
> EDIT: After running again now it's passing.

yes, unfortunately it is.

kernels/volk/volk_32f_expfast_32f.h
jdemel: Found it. There's quite a few problems:
- the constant names do not tell any information.
- the variables pollute global namespace

Making a `#define` local to a specific header, would be difficult to achieve. Having those constants with more descriptive names would at least help.
Unfortunately, I can't find the description of the underlying algorithm right away.

Would it help your case to rename these variables with prefixes? `VOLK_EXPFAST_...`? The `#undef` part would be a work around but would add even more confusion. Also, I fear this makes the system more prone to errors.

jdemel:
I like those additions. Though I'm worried this will break the API in some way. If we break the API, we can only add this feature in a new major release "v4". 

`volk.h` is the canonical entrypoint for everything VOLK, that's why this needs to be carefully changed. Can you go into further detail how your changes do not affect the current API? 

Can you add a CI test for your case? From prior experience: everything that does not receive a CI test will break soon.

kernels/volk/volk_32f_expfast_32f.h
xerpi: Actually, I find it quite standard practice in open-source projects to `#undef` macros after use, precisely to avoid polluting the preprocessor namespace. That said, I'll also rename the constants with the `VOLK_EXPFAST_` prefix as you suggested.

CMakeLists.txt
xerpi: I've used `#cmakedefine VOLK_STATIC_DISPATCH` in `constants.c.in` so that `volk_prefix()` calls `getenv("VOLK_PREFIX")` on dynamic dispatch builds, and just returns a static string (the configured install prefix, or empty if not set) on static dispatch builds. This way the API is the same in both cases.

xerpi:
> I like those additions. Though I'm worried this will break the API in some way. If we break the API, we can only add this feature in a new major release "v4".
> 
> `volk.h` is the canonical entrypoint for everything VOLK, that's why this needs to be carefully changed. Can you go into further detail how your changes do not affect the current API?
> 
> Can you add a CI test for your case? From prior experience: everything that does not receive a CI test will break soon.

Updates:
- Renamed `Mln2`, `A`, `B`, `C` macros in `volk_32f_expfast_32f.h` to `VOLK_EXPFAST_*` (keeping the `#undef`s at the end).
- `constants.c.in` is now compiled in both static and dynamic dispatch builds, using `#cmakedefine VOLK_STATIC_DISPATCH` to conditionally include the `getenv` call in `volk_prefix()` only for dynamic dispatch. `constants.h` is now always installed.

Regarding API compatibility:
- The `constants.h` API is now identical in both static and dynamic dispatch builds.
- However, `volk_prefs.h` and `volk_cpu.h` are still excluded from static dispatch installs, as those APIs depend on runtime machine detection which is not applicable in a static dispatch context.

I've added a `build-ubuntu-static-dispatch` CI job that builds with `-DVOLK_STATIC_DISPATCH=generic`.

xerpi:
Actually, I realized that some of the options such as disabling apps and testing are not related to static dispatch but to cross-compiling. We can still generate a static dispatch build for the host architecture and run tests without any problem. So I switched some options to be gated on `CMAKE_CROSSCOMPILING` instead of `VOLK_STATIC_DISPATCH`.

tests/CMakeLists.txt
```diff
@@ -6,10 +6,6 @@
 # SPDX-License-Identifier: LGPL-3.0-or-later
 #
 
-if(NOT ENABLE_TESTING)
-    return()
-endif(NOT ENABLE_TESTING)
-
```
jdemel, line 12: The intention here is that we can gate tests behind that common switch. The common place for these decisions is at the beginning of the corresponding CMakeLists.txt file. I assume this was part of some debug change and might need to be reverted?

CMakeLists.txt
```diff
@@ -396,7 +417,9 @@ add_subdirectory(lib)
 ########################################################################
 # Add tests
 ########################################################################
-add_subdirectory(tests)
+if(ENABLE_TESTING)
+    add_subdirectory(tests)
+endif()
```
jdemel, line 422: Ah. That relates to my earlier comment. I suggest to keep the checks in the folder where they are needed. Thanks =D

jdemel (changes requested):
Thanks for your PR. It looks good but there's still something broken with the static compile test. I suggest to fix this. Also, that'd help a lot to reproduce what you implemented. Thanks.

tests/CMakeLists.txt
xerpi: That's a good point. However, in most of the CMake projects I've seen so far, the top-level `CMakeLists.txt`, which reads the `option`s, is the one that gates adding subdirectories, like:
>
```cmake
if(ENABLE_TESTING)
    add_subdirectory(tests)
endif()
```
>
and currently, the other options follow the same pattern:
>
```cmake
if(ENABLE_UTILITY_APPS)
    add_subdirectory(apps)
endif()

if(ENABLE_MODTOOL)
    add_subdirectory(python/volk_modtool)
endif()
```
>
That said, I'm happy to switch all of them to always be included and do the check + `return()` in the sub-`CMakeLists.txt` instead, if you prefer!

jdemel: Thanks. It's basically part of a very old discussion on CMake structure preferences.

I'm fine with it.

jdemel (commented):
Thanks for your update. One last comment.

Unfortunately, I didn't see a notification that there was any activity in this issue. So, I totally missed your update. Since GH threw errors while I tried to add my latest comment, I assume GH also has problems with notifications. 
Anyways, I'm sorry that my reply is late.

jdemel:
> Thanks for your update. One last comment.
> 
> Unfortunately, I didn't see a notification that there was any activity in this issue. So, I totally missed your update. Since GH threw errors while I tried to add my latest comment, I assume GH also has problems with notifications. Anyways, I'm sorry that my reply is late.

And the last comment is missing... Since commenting on lines of code does not work at the moment in [tmpl/volk_dynamic_dispatch.tmpl.h](https://github.com/gnuradio/volk/pull/858/changes#diff-eb935e247e7978a5a2ae302f5b9882e7d747c7e584a7326469754edefe08b481)
```c
/*!
 * Is the pointer on a machine alignment boundary?
 *
 * Note: for performance reasons, this function
 * is not usable until another volk API call is made
 * which will perform certain initialization tasks.
 *
 * \param ptr the pointer to some memory buffer
 * \return 1 for alignment boundary, else 0
 */
VOLK_API bool volk_is_aligned(const void *ptr);
```
The comment/documentation provides some important and non-trivial information on the behavior of this function that should be restored.

xerpi:
> > Thanks for your update. One last comment.
> > Unfortunately, I didn't see a notification that there was any activity in this issue. So, I totally missed your update. Since GH threw errors while I tried to add my latest comment, I assume GH also has problems with notifications. Anyways, I'm sorry that my reply is late.
> 
> And the last comment is missing... Since commenting on lines of code does not work at the moment in [tmpl/volk_dynamic_dispatch.tmpl.h](https://github.com/gnuradio/volk/pull/858/changes#diff-eb935e247e7978a5a2ae302f5b9882e7d747c7e584a7326469754edefe08b481)
> 
> ```c
> /*!
>  * Is the pointer on a machine alignment boundary?
>  *
>  * Note: for performance reasons, this function
>  * is not usable until another volk API call is made
>  * which will perform certain initialization tasks.
>  *
>  * \param ptr the pointer to some memory buffer
>  * \return 1 for alignment boundary, else 0
>  */
> VOLK_API bool volk_is_aligned(const void *ptr);
> ```
> 
> The comment/documentation provides some important and non-trivial information on the behavior of this function that should be restored.

I've updated the static dispatch docs for `volk_is_aligned()` to make the behavior explicit:

> Unlike dynamic dispatch builds, static dispatch embeds the machine alignment in this inline function, so no runtime initialization is required before calling it.

Also, I've rebased from `main`.

jdemel (approved):

jdemel:
Thank you very much!

Issue #859, opened 2026-04-14
32f_log2_32f_neon QA flaky in Ubuntu 25.10 arm g++14 CI


marcusmueller:
[qa_volk_32f_log2_32f](https://github.com/gnuradio/volk/actions/runs/24390320633/job/71234356978?pr=858#step:7:664)

```
43: Test command: /usr/bin/sh "/home/runner/work/volk/volk/build/lib/volk_32f_log2_32f_test.sh" "/home/runner/work/volk/volk/build/lib"
43: Working Directory: /home/runner/work/volk/volk/build/lib
43: Test timeout computed to be: 10000000
43: 
43: RUN_VOLK_TESTS: volk_32f_log2_32f(vlen=131071, iter=1, tol=5e-06)
43: arch                       |           time |     throughput |  speedup |    max_abs |
43: ---------------------------+----------------+----------------+----------+------------+
43: generic                    |        0.86 ms |    1225.2 MB/s |        - |          - |
43: neon                       |        0.17 ms |    6085.6 MB/s |    4.97x |    5.0e-06 |
43: neonv8                     |        0.16 ms |    6358.1 MB/s |    5.19x |    4.1e-06 | *
43: Best aligned arch          | neonv8 (5.19x)
43: Best unaligned arch        | neonv8 (5.19x)
43: 
43: volk_32f_log2_32f: fail on arch neon
43:   index |        in0 |   expected |     actual |   rel_err |       tol
43: --------+------------+------------+------------+-----------+----------
43:  125098 |     0.2499 |    -2.0004 |    -2.0004 |   2.5e-06 |   5.0e-06
43: ----------------------------------------------------------------------------------------
 43/196 Test  #43: qa_volk_32f_log2_32f ....................................................................................................................***Failed    1.98 se
```

marcusmueller:
(note that this issue is **not** that the test is *wrong*. I just think it should always be failing this kernel, if at all.)

marcusmueller:
(also note: volk_neon_intrinsics.h claims of _vlog2_poly_neonv8 that accuracy is no worse than `~1.55e6`, which would indicate this failure should not happen, the rest of the kernel is integer math. So, either the comment is wrong, the polynomial coefficients are wrong, or something else is wrong here.)

marcusmueller:
random thoughts:

[...]

PR #860, opened 2026-04-14
qa_utils: print errors with enough precision to represent tolerance

marcusmueller:
Issues like #859 highlight that we're not seeing the numerical values to a precision good enough to debug. Change that.

### DIFF
--- a/lib/qa_utils.cc
+++ b/lib/qa_utils.cc
@@ -9,6 +9,7 @@
  */
 
 #include "qa_utils.h"
+#include <fmt/ranges.h>
 #include <volk/volk.h>
 
 #include <volk/volk.h>        // for volk_func_desc_t
@@ -18,6 +19,8 @@
 #include <stdint.h>    // for uint16_t, uint64_t
 #include <sys/time.h>  // for CLOCKS_PER_SEC
 #include <sys/types.h> // for int16_t, int32_t
+#include <algorithm>
+#include <array>
 #include <chrono>
 #include <cmath>    // for sqrt, fabs, abs
 #include <cstring>  // for memcpy, memset
@@ -26,9 +29,11 @@
 #include <limits>   // for numeric_limits
 #include <map>      // for map, map<>::mappe...
 #include <random>
+#include <tuple>
 #include <vector> // for vector, _Bit_refe...
 
-#include <fmt/core.h>
+#include <fmt/format.h>
+#include <fmt/ranges.h>
 
 // Warmup time for CPU frequency scaling (ms)
 static double g_warmup_ms = 2000.0;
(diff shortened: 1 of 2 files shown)


lib/qa_utils.cc
```diff
@@ -26,9 +29,11 @@
 #include <limits>   // for numeric_limits
 #include <map>      // for map, map<>::mappe...
 #include <random>
+#include <tuple>
 #include <vector> // for vector, _Bit_refe...
 
-#include <fmt/core.h>
+#include <fmt/format.h>
+#include <fmt/ranges.h>
```
jdemel: This is a double include. I suggest to remove the empty lines between includes, let clang-format do the sorting and automatically remove duplicates =D

lib/qa_utils.cc
```diff
@@ -617,23 +650,34 @@ void print_error_table(const std::vector<unsigned int>& fail_indices,
                        float tol,
                        int max_errors = 10)
 {
+    constexpr std::array<const char*, 5> columns{
```
jdemel, line 676: Are we able to use `std::string_view` in place of `char*` here?

lib/qa_utils.cc
```diff
@@ -606,6 +611,34 @@ bool icompare(t* expected,
...
+    if (tol >= 1.0f) {
+        /* |tolerance| > 1 : you get no courtesy digit after the dot */
+        return { tol_length, 0 };
+    }
+    const float log10 = std::log10(tol);
+    return { tol_length,
+             std::min<unsigned int>(tol_length - dot_length, -std::floor(log10)) };
+}
+template <class T>
+unsigned int max_col_length(const T& columns)
```
jdemel: could you add a new line between functions? =D

could you rename the function to `max_column_width`? That's what is essentially computed, correct?

lib/qa_utils.cc
```diff
@@ -617,23 +650,34 @@ void print_error_table(const std::vector<unsigned int>& fail_indices,
                        float tol,
                        int max_errors = 10)
 {
+    constexpr std::array<const char*, 5> columns{
+        "index", "expected", "actual", "rel_err", "tol"
+    };
+    unsigned int index_len = fmt::formatted_size("{}", columns[0]) + 1;
+    unsigned int col_len = max_col_length(columns);
+    /* choose enough digits to represent tolerance */
+    const auto prec_tup = tol_precision(tol, col_len);
```
jdemel: I suggest to use structured binding here. Something like:
```suggestion
    const auto [tol_length, value_prec] = tol_precision(tol, col_len);
```
This would improve readability here a lot. I was looking for the exact meaning of these variables for a while.

lib/qa_utils.cc
```diff
@@ -617,23 +650,34 @@ void print_error_table(const std::vector<unsigned int>& fail_indices,
...
+    fmt::print("{}::{}\n", std::get<0>(prec_tup), std::get<1>(prec_tup));
+    col_len = std::max(col_len + 1, std::get<0>(prec_tup));
+    unsigned int val_prec = std::get<1>(prec_tup);
+
     if (fail_indices.empty())
         return;
```
jdemel: Could you do me a favor and add curly braces here?

jdemel (commented):
Thanks a lot. Only very minor comments.

lib/qa_utils.cc
marcusmueller: No; `std::string_view` never holds the data itself, and we need something to actually store these `"index"` and so on. We could have a `std::array<std::string,5>` instead, but that would not improve the code – these are compile-time constant const, zero-terminated strings, so the compiler always knows how long they are, and they can never be modified.

`constexpr std::string(…)` only starts to exist with C++20 std lib, so converting to `std::string would mean `columns`  can no longer be `constexpr`, and then compile format checks and -compilation no longer works.

lib/qa_utils.cc
marcusmueller: looking at the fact I returned a tuple there, I'm almost convinced that was my original intent. Huh! Done.

lib/qa_utils.cc
marcusmueller: sure, but only if I get to do that consistently (OK, I `clang-tidy-20 -p ./build/ --fix --format-style=file --checks readability-braces\* lib/qa_utils.cc` for that and didn't do it myself.)

lib/qa_utils.cc
marcusmueller: good call.

lib/qa_utils.cc
jdemel: Interestingly, `std::string_view` would work in a constexpr context: https://stackoverflow.com/a/77358740

Anyways, resolving this comment. Can be deferred to a different PR.

jdemel (approved):
LGTM.

Thanks for working on this PR!

PR #863, opened 2026-05-31
mod_range: Add proper include guard for every implementation

### LINKED ISSUE #862: The libvolk library is missing aligned versions of `volk_32f_s32f_s32f_mod_range_32f`
But in kernels/volk/volk_32f_s32f_s32f_mod_range_32f.h, both aligned and unaligned versions of this function are present.
The aligned versions are located directly after the unaligned ones, without #ifdef and #endif between them.
For example, after copying the following lines in between and recompiling the library, all versions were present:

```
#endif /* LV_HAVE_AVX */

#ifdef LV_HAVE_AVX
#include <xmmintrin.h>
```
This makes no difference in C, but it seems to be relevant for some script files.

Thank you.

jdemel:
Previously, multiple implementations, specifically the aligned and unaligned versions, were smashed into one block. While not a bug per-se, it lead some scripts to believe that there's only one implementation instead of an aligned and unaligned version.

In the build folder, previously one could find:
```bash
$ grep -A 1 mod_range lib/volk_machine_avx512f_64_mmx_orc.c
    "volk_32f_s32f_s32f_mod_range_32f",
    {"generic", "u_avx", "u_sse2", "u_sse"},
```
but now:
```bash
$ grep -A 1 mod_range lib/volk_machine_avx512f_64_mmx_orc.c
    "volk_32f_s32f_mod_rangepuppet_32f",
    {"generic", "u_sse", "a_sse", "u_sse2", "a_sse2", "u_avx", "a_avx"},
```

Fixes: #862

### DIFF
--- a/kernels/volk/volk_32f_s32f_s32f_mod_range_32f.h
+++ b/kernels/volk/volk_32f_s32f_s32f_mod_range_32f.h
@@ -113,6 +113,10 @@ static inline void volk_32f_s32f_s32f_mod_range_32f_u_avx(float* outputVector,
     volk_32f_s32f_s32f_mod_range_32f_generic(
         outPtr, inPtr, lower_bound, upper_bound, num_points - eight_points * 8);
 }
+#endif /* LV_HAVE_AVX */
+
+#ifdef LV_HAVE_AVX
+#include <xmmintrin.h>
 static inline void volk_32f_s32f_s32f_mod_range_32f_a_avx(float* outputVector,
                                                           const float* inputVector,
                                                           const float lower_bound,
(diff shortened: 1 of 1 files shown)


marcusmueller:
haha, that takes me back!

marcusmueller (approved):
This seems correct, but I forget *why* we need to surround each protokernel in its own `#ifdev LV_HAVE_*`, even if two consecutive ones are about the same machine.

Is this about consistency (good reason!) or does our source code handling fall over at some point if we don't? 

In both cases, do we document that all somewhere?

jdemel:
The reason for the separate blocks can be found in how we build all our different kernel implementations.

Improving/Documenting this would be another PR.

marcusmueller:
> The reason for the separate blocks can be found in how we build all our different kernel implementations.

haha, ok, if that's how we can explain that:

> Improving/Documenting this would be another PR.

seems important, because the question here is *how* and *where* that interacts.

PR #867, opened 2026-05-31
CI: cache builds

marcusmueller:
This should significantly speed up CI.

For now, uses the wrong fallback readonly repo; will fill in valid read-only credentials in a follow-up commit once this is proven to work correctly.

Then, after that, we add R/W credentials to the repo action secrets and start populating the cache.

### DIFF
--- a/.github/workflows/android_build.yml
+++ b/.github/workflows/android_build.yml
@@ -31,11 +31,11 @@ jobs:
           submodules: "recursive"
 
       - name: Update repositories
-        run: sudo apt update
+        run: sudo DEBIAN_FRONTEND=noninteractive apt-get update > /dev/null
 
       # All dependencies
       - name: Install dependencies
-        run: sudo apt install -y cmake python3-mako curl
+        run: sudo DEBIAN_FRONTEND=noninteractive apt-get install -y cmake python3-mako curl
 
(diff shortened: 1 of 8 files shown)


jdemel (commented):
Thanks for the PR!

I really like the idea to use sccache.

There's one more strategic thing, that we should discuss. Adding scripts that add cmake options would be one way.

Another way would be to go with CMakePresets. These 2 approaches tend to clash. Also, presets tend to integrate better with other tooling.

What's your take on this?

marcusmueller:
In general, you specify the compiler launcher with as `CMAKE_{LANGUAGE}_COMPILER_LAUNCHER`, regardless of how you set that CMake variable.

Since using sccache is pretty specific to CI here, I would strongly suggest we keep it minimally-invasive and just, as this PR does, pass it as `cmake -DCMAKE_C_COMPILER_LAUNCHER…` and not put it in any scripts or presets:

> Adding scripts that add cmake options would be one way.

As you can see, all that happens is that when sccache is available, it's getting used by including `-DCMAKE_{C,CXX}_LAUNCHER` in the, anyways CI workflow-specific, `cmake` command line. And I'd keep it that way :)

jdemel (approved):

PR #870, opened 2026-08-11
tests: use correct fmt include

marcusmueller:
On current fmt (12.2 fmt, probably starting earlier), fmt/core.h does no
longer include fmt::format.

Crofton found this through his OE builds. Kudos.

### DIFF
--- a/lib/qa_utils.cc
+++ b/lib/qa_utils.cc
@@ -28,7 +28,7 @@
 #include <random>
 #include <vector> // for vector, _Bit_refe...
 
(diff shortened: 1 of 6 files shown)


balister:
I applied this in the recipe and the build succeeded.

jdemel (approved):
LGTM.

FYI: traditionally, we included `fmt/core.h` because it is the smaller include and sufficient for most of the fmt features. Though, one of the very recent releases tried to align the header structure with the STL... which breaks our lean approach.

PR #871, opened 2026-08-13
CMake: don't disable QA, and tools, on cross builds generally

marcusmueller:
This must have unnoticedly slipped in, but also means our CI was not running QA since the the static fix has been merged :grimace:


balister:
OK, this solves my problem. Continuing my OE build post holiday updates.

jdemel:
I'd like to add a screenshot of the presumably broken CI job view. But GH seems to fail at accepting PNGs via drag and drop and the file selector. Anyways, the raw logs are there but the webview doesn't show any results anymore.

As far as I can tell, the issue seem to be spurious/flaky tests.

PR #873, opened 2026-08-25
Convert tests

### LINKED ISSUE #872: Float to int32 conversion may cause overflow
There's an issue in `volk_32f_s32f_convert_32i`.

The intended logic is:
1. scale the input float value by scalar
2. clamp float value to 32bit integer range [-2147483648, 2147483647], or [-2147483648, 2147483648)
3. Convert to int.

Unfortunately, the operation:
```C
float value = 2147483648.0f - 1.0f;
```
does not result in `2147483647.0f`, but `2147483648.0f` since the next lower float representable number is `2147483520.0f`. Thus,
```C
int32_t result = (int32_t) 2147483648.0f - 1.0f;
```
results in `-2147483648`, aka overflow.

It may be possible to clamp the float value to `2147483520.0f` but a number such as `6167483520.0f` would then result in `2147483520` instead of `2147483647` which would've been the expected value.

jdemel:
This PR started as an effort to implement tests for the converter with gtest. It uncovered a bug and it continued with a lot of code clean-up and additional implementations for AVX512.

Fixes #872


CMakeLists.txt
```diff
@@ -23,6 +23,8 @@ enable_language(C)
 
 enable_testing()
 
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON CACHE BOOL "Export compile commands" FORCE)
```
marcusmueller: an explanation *why* we enforce that as a comment or in the commit msg would certainly help!

jdemel: Thanks for your early review. :) This is something I intend to remove.

Once I'm done with all the changes and the CI actually passes again, I'd be very happy if you could review this PR. I'd ask for a review at this point. Thanks!

marcusmueller: Sure, Started reviewing since this isn't marked "draft" :)

kernels/volk/volk_32f_s32f_convert_16i.h
jdemel, line 1: @Ka-zam Can you have a look at the NEON changes? The new and more extensive tests revealed some bugs that are fixed now. However, some NEON code looks way more verbose now. Maybe there's a better way to implement all of this.

CMakeLists.txt
jdemel: I removed this change. The PR is ready now. Happy to receive your feedback.

PR #878, opened 2026-09-10
volk_32f_exp_32f: Add AVX2 implementation

harshitthek:
Add aligned (_a_avx2) and unaligned (_u_avx2) AVX2 implementations for
volk_32f_exp_32f. Mechanically widened from the existing SSE2 Cephes
minimax range reduction, processing 8 floats per iteration instead of 4.

Tested with volk_test_all (vlen=131071):
- a_avx2: 7.31x over generic, max_rel 1.2e-07
- u_avx2: 6.56x over generic, max_rel 1.2e-07

### DIFF
--- a/kernels/volk/volk_32f_exp_32f.h
+++ b/kernels/volk/volk_32f_exp_32f.h
@@ -164,6 +164,74 @@ volk_32f_exp_32f_a_sse2(float* bVector, const float* aVector, unsigned int num_p
 
 #endif /* LV_HAVE_SSE2 for aligned */
 
+#ifdef LV_HAVE_AVX2
+#include <immintrin.h>
+
+static inline void
+volk_32f_exp_32f_a_avx2(float* bVector, const float* aVector, unsigned int num_points)
+{
+    const unsigned int eighthPoints = num_points / 8;
+
+    const __m256 one = _mm256_set1_ps(1.0);
+    const __m256 exp_hi = _mm256_set1_ps(88.3762626647949);
+    const __m256 exp_lo = _mm256_set1_ps(-88.3762626647949);
+    const __m256 log2EF = _mm256_set1_ps(1.44269504088896341);
+    const __m256 half = _mm256_set1_ps(0.5);
+    const __m256 exp_C1 = _mm256_set1_ps(0.693359375);
+    const __m256 exp_C2 = _mm256_set1_ps(-2.12194440e-4);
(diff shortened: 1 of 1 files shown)


kernels/volk/volk_32f_exp_32f.h
```diff
@@ -249,6 +327,84 @@ volk_32f_exp_32f_u_sse2(float* bVector, const float* aVector, unsigned int num_p
 
 #endif /* LV_HAVE_SSE2 for unaligned */
 
+#ifdef LV_HAVE_AVX2
+#include <immintrin.h>
+
+static inline void
+volk_32f_exp_32f_u_avx2(float* bVector, const float* aVector, unsigned int num_points)
```
jdemel, line 324: I know that this file is written in an "what one might call ancient C-style" now. However, the new kernels don't need to mimic this style. If you want, it'd be great to modernize it in a few ways:
- keep the loop index in the loop
- make variables const where possible
- make constants, e.g., `one`, constant
- remove the pointer redefinitions `float *bPtr = bVector;`
- keep variables like `__m256 aVal` ... in the loop.
- use the generic kernel for the tail call.

Let me know if you want to work on that. Otherwise, this is fine.

jdemel (commented):
Thanks and welcome :)

kernels/volk/volk_32f_exp_32f.h
harshitthek: Thanks! Will push an updated commit with all six changes.

harshitthek:
Thanks! Pushed an update with all the style changes.

For the generic tail call: volk_32f_exp_32f_generic is defined later in the file (in the unaligned section), so calling it from the AVX2 implementations causes an implicit declaration error. Kept a manual expf loop instead. Happy to move the generic definition earlier in a follow-up if you'd prefer that.

kernels/volk/volk_32f_exp_32f.h
jdemel: Thanks :)

jdemel (approved):

Regarding the generic kernel question: When I update or work on a kernel I move the generic kernel to the top. It solves your issue but also, it provides often the simplest way to understand what that kernel does without the need to understand the SIMD code.

harshitthek:
Thanks for the review and the tip about moving generic to the top. Will do that in future kernel work.

PR #879, opened 2026-09-12
Restrict fmtlib fetch content

### LINKED ISSUE #876: VOLK 3.3 _may_ generate libfmt.a stump
If `libfmt` is not already installed, then VOLK 3.3 will generate the static library `libfmt.a` & install it into `$CMAKE_INSTALL_PREFIX/lib`. This stump library will then conflict with the system installed `libfmt` when building gnuradio. Suggest we add a cmake check that `fmt` is already available to use, to avoid this issue.

jdemel:
Restrict the CMake FetchContent option to static builds explicitly. This should not interfere with other systems where fmtlib should've been installed.

A few more lines where touched because I auto-formatted the CMake file as well.

Fixes #876


jdemel:
It turns out, quite a few configurations depend on the FetchContent approach to provide fmtlib. Also, switching to a system fmtlib install turn out to be challenging.

PR #882, opened 2026-09-13
docs: add specific overflow warnings for 16i kernels

### LINKED ISSUE #707: Various kernels with 16i inputs do not handle integer overflow consistently
In #77, the range of integers used when testing 16i inputs was reduced to -7 .. +7, to accommodate the `volk_16ic_x2_dot_prod_16ic` kernel which does not handle integer overflows consistently between its various protokernels. In fact, other kernels suffer from the same problem. If the testing range is increased to `std::numeric_limits<int16_t>::min()` .. `std::numeric_limits<int16_t>::max()`, then the following tests fail:

* `qa_volk_16i_32fc_dot_prod_32fc`
* `qa_volk_16ic_magnitude_16i`
* `qa_volk_16ic_x2_dot_prod_16ic`
* `qa_volk_16ic_x2_multiply_16ic`

Overflows in `volk_16ic_x2_dot_prod_16ic` cause sporadic test failures even with the reduced range, as documented in #676; #706 should improve the situation, but in the longer term, either these kernels should be improved to handle overflows consistently, or the limitations of these kernels should be documented and the test framework should ensure that inputs do not stray outside the documented limits.

ItsMadEyeMoody:
fixes #707 

## Summary
Add specific overflow warnings to the documentation of four 16-bit integer kernels

## Changes
- **volk_16ic_x2_multiply_16ic**: Replaced vague "Saturation is not checked" with specific overflow conditions and guidance.
- **volk_16ic_x2_dot_prod_16ic**: Fixed misleading claim that "results are saturated" — clarified that only the accumulation uses saturation, while the intermediate multiplication can still overflow.
- **volk_16ic_magnitude_16i**: Added warning about output overflow when the magnitude of the input exceeds `SHRT_MAX`.
- **volk_16i_32fc_dot_prod_32fc**: Added warning about precision inconsistencies across protokernels with full-range inputs.

## Context
Issue #707 identified that these kernels produce inconsistent results across different protokernels (generic, SSE, AVX, NEON) when inputs span the full `int16_t` range. The test framework currently works around this by limiting test inputs to the range -6 to +6.
As discussed by @argilo and @jdemel, this PR documents the specific conditions under which overflow can occur so users can make informed decisions about their input ranges.


kernels/volk/volk_16ic_magnitude_16i.h
```diff
@@ -15,6 +15,11 @@
  * Computes the magnitude of the complexVector and stores the results
  * in the magnitudeVector.
  *
+ * \b WARNING: When the magnitude of an input complex value exceeds
+ * SHRT_MAX (e.g., both real and imaginary parts are near their maximum
+ * values), the result may overflow the int16_t output range. Different
+ * protokernels may produce inconsistent results in this case.
+ *
```
jdemel, line 26: So the interesting thing here:
Under which specific conditions can overflow happen? e.g., real is SHRT_MAX and imag > 0?
What would be the expected behavior and what is the actual behavior? Which kernels follow the expected behavior? Which kernels do not follow the expected behavior?
Those would be valuable additions to the documentation.

kernels/volk/volk_16i_32fc_dot_prod_32fc.h
```diff
@@ -18,6 +18,10 @@
  * vectors. The result is a single value stored in the \p result
  * address and will be complex.
  *
+ * \b WARNING: When input values span the full int16_t range, different
+ * protokernels may produce slightly inconsistent results due to
+ * differences in intermediate precision and accumulation order.
```
jdemel: Can you quantify the error? And can you quantify the precision tolerance?

kernels/volk/volk_16ic_x2_dot_prod_16ic.h
```diff
@@ -12,9 +12,15 @@
...
- * data type.
+ * Multiplies two input complex vectors (16-bit integer each component)
+ * element-wise and accumulates them, storing the result. The accumulation
+ * is performed using saturating addition.
+ *
+ * \b WARNING: The intermediate element-wise multiplication does not use
+ * saturation and may overflow if input values span the full int16_t range.
+ * Different protokernels may produce inconsistent results in this case.
+ * To avoid overflow, ensure that input values are small enough that their
+ * products fit within the int16_t range.
```
jdemel: Could you add that this is a performance trade-off?

jdemel (commented):
Thanks for these additions!

gitty-prakhar:
Thanks for the feedback I've updated the documentation to include the exact overflow conditions and expected vs actual behavior across protokernels, quantified the error tolerance and noted the performance trade offs for non saturating arithmetic

kernels/volk/volk_16ic_magnitude_16i.h
```diff
@@ -15,6 +15,15 @@
...
  * in the magnitudeVector.
  *
+ * \b WARNING: Overflow occurs when the computed magnitude exceeds
+ * SHRT_MAX, i.e., when re*re + im*im > SHRT_MAX*SHRT_MAX. For
+ * example, re = SHRT_MAX and im > 0, or both |re| and |im| > 23170.
+ * The expected behavior is to saturate the output to SHRT_MAX. The
+ * AVX2 protokernel does this correctly via _mm256_packs_epi32. However,
+ * the SSE and generic protokernels cast the float result to int16_t,
+ * which is undefined for out-of-range values. The NEON protokernels
+ * use a narrowing conversion that wraps on overflow.
```
jdemel, line 25: Thanks for looking at this in more detail. Since this is a documentation PR, let's leave it as is. Could you open an issue that describes these differences though? This should be fixed at some point.

